PR #14, opened 2015-07-28
Code review comments

wlevine:
My comments on code are in the patch, sorry I didn't look at that much of it.

Some general comments:

I couldn't find an easy way to test daru as I edited the source. What I had to do was run `pry -I./lib` and then `require 'daru'`. It would be nice if there were a more convenient way to do this. For example, nmatrix has a `rake pry` task, and some projects have a `bin/console` script, that just opens irb/pry with the appropriate libraries loaded. It also would be nice to have a rake task for installing the gem.

I think using `and` and `or` instead of `&&` and `||` is pretty frowned upon in Ruby. [One explanation of why](http://stackoverflow.com/questions/2083112/difference-between-or-and-in-ruby).

Is there a way to save plots as images programatically? Without iPython? I may have missed it, but I couldn't find it. I realize that this is a limitation of your plotting library, and not daru, so I'm definitely not asking you to implement it, but this is really an essential feature. I had a difficult time installing iruby/ipython, so it would also be nice to be able to do it without them.

I took a quick look at your query code. I would really, really appreciate it if you developed something that was more powerful and that was just powered by plain ruby. Let me try to give you an example. In particle physics, we use a library called [ROOT](https://root.cern.ch/drupal/), where the essential data storage abstraction is a class called [TTree](https://root.cern.ch/root/html/TTree.html), which is roughly equivalent to a dataframe. Say we have a TTree with three "columns": position_x, position_y, momentum. There is a really handy method `TTree::Draw()`, so that if we want to plot a histogram of momentum, selecting only events (rows) with certain values of position we can just say `tree->Draw("momentum", "position_x*position_x + position_y*position_y < 1")`. `TTree::Draw()` is great and we use it all the time, but for any type of refined analysis it becomes useless because the syntax used in the selection string is not powerful enough. For more complex selection criteria (and we always want more complex selection criteria), we have to use a totally different method which is pretty unwieldy. The reason for this is that we're working in C++ and we can't just pass around arbitrary C++ code, instead we just pass a small string that approximates a very tiny subset of C, which is very limiting. But we don't have that limitation in Ruby. I think it would be really great to have a selection function that took a block which could evaluate an arbitrarily complex function for each row and then return true or false. That way we wouldn't have to learn any new syntax, and you wouldn't have to worry about implementing query code for every possible use case. I haven't really deeply thought this through, so maybe this wouldn't work as well I think. Or maybe it's already possible to do this.

### DIFF
--- a/lib/daru/accessors/nmatrix_wrapper.rb
+++ b/lib/daru/accessors/nmatrix_wrapper.rb
@@ -10,6 +10,7 @@ module Daru
       end
 
       def map!(&block)
+        #Not sure why you can't just use map! on a NMatrix slice?
         @data = NMatrix.new [@size*2], map(&block).to_a, dtype: nm_dtype
         self
       end
@@ -21,19 +22,31 @@ module Daru
       attr_reader :size, :data, :nm_dtype
       
       def initialize vector, context, nm_dtype=:int32
+        #Check that vector is correct shape and type.
+        #
+        #It would be really nice if this worked with nx1 matrices, since
+        #that's what NVector produces. I think this would just require a call
+        #to reshape. Right now it only works with 1xn
+        #matrices and flat 1D NMatrix's.
         @size = vector.size
+        #It's too bad NMatrix doesn't have resize method.
         @data = NMatrix.new [@size*2], vector.to_a, dtype: nm_dtype
         @context = context
         @nm_dtype = @data.dtype
         # init with twice the storage for reducing the need to resize
       end
 
+      #Shouldn't this take just a single integer argument?
       def [] *index
         return @data[*index] if index[0] < @size
         nil
       end
  
       def []= index, value
+        #Do you want to allow the case where index > @size? It doesn't seem like
+        #this is handled properly?
+
+        #what if index > @size*2?
         resize     if index >= @data.size
         @size += 1 if index == @size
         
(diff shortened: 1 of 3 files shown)


v0dro:
I'm not sure if I correctly understand your concern about the query syntax. If I get you right, you basically want a more powerful querying syntax that will depend on the sort of string querying that you demonstrated right?

You can query and plot with the current syntax in this manner:

``` ruby
momentum = df.where((df[:position_x]*df[:position_x] + df[:postion_y]*df[:position_y]).lt(1))[:momentum]
momentum.plot type: :histogram

```

IMO this suffices for simple cases.

In case you want block-based querying, you can use the `DataFrame#filter` method:

``` ruby

new_df = df.filter(:row) do |row|
  row[:position_x]*row[:position_x] +  row[:position_y]*row[:position_y] < 1
end

new_df[:momentum].plot type: :histogram
```

v0dro:
About the plotting library, I'm planning to add support for Ivan's GNU plot gem soon. That should allow exporting plots directly.

wlevine:
`DataFrame#filter` looks like what I was asking for, thanks for pointing it out.

v0dro:
I'm closing this PR, but if you have any additional points to raise please feel free to write them here.

PR #23, opened 2015-09-21
faster algorithm for inner join

peterktung:
Code used for benchmarking: https://gist.github.com/peterktung/63c3b6d87fe22cb85621

Results:
```
Before:
Rehearsal --------------------------------------------------------
full_inner:          413.590000   0.850000 414.440000 (415.140870)
sparse_inner:        386.930000   0.380000 387.310000 (387.571845)
--------------------------------------------- total: 801.750000sec

full_inner:          388.470000   0.370000 388.840000 (389.073046)
sparse_inner:        400.040000   0.460000 400.500000 (400.807328)

After:
Rehearsal --------------------------------------------------------
full_inner:           85.980000   0.130000  86.110000 ( 86.224135)
sparse_inner:         83.670000   0.070000  83.740000 ( 83.777885)
--------------------------------------------- total: 169.850000sec

full_inner:           87.270000   0.100000  87.370000 ( 87.423039)
sparse_inner:         85.260000   0.110000  85.370000 ( 85.442580)
```

### DIFF
--- a/daru.gemspec
+++ b/daru.gemspec
@@ -51,7 +51,6 @@ EOF
 
   spec.add_runtime_dependency 'reportbuilder', '~> 1.4'
   spec.add_runtime_dependency 'spreadsheet', '~> 1.0.3'
-  spec.add_runtime_dependency 'bloomfilter-rb', '~> 2.1'
 
   spec.add_development_dependency 'bundler', '~> 1.10'
   spec.add_development_dependency 'rake'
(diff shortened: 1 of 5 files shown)


v0dro:
Agreed. You have any algorithms in mind?

v0dro:
daru needs to work with both MRI and JRuby. Since the bloomfilters-rb gem depends on MRI, lets keep the previous implementation in case the interpreter being used by the user is JRuby.

Could you please change the code such that the bloomfilters code is only used in case the `bloomfilters-rb` gem is installed? Also, don't add it as a runtime dependency. Add it as a development so that it's tested with rspec.

You can do something like `Daru.has_gsl?`, see [this code](https://github.com/v0dro/daru/blob/master/lib/daru.rb#L56) for reference.

Also please don't change version numbers in Pull Requests. That should be a separate PR/commit.

peterktung:
Updated based on your feedback.

peterktung:
I just found a bug when it's a one/many to many join, I will fix and add a test case to it.

v0dro:
Looks good thanks! I'll add a note in the README later.

PR #25, opened 2015-10-06
faster filter_row implementation

peterktung:
Benchmark code here: https://gist.github.com/peterktung/f60f83d7ec9d6e84fe71
```
Before:
Rehearsal --------------------------------------------------------
filter_row:           98.750000   1.090000  99.840000 ( 99.890700)
---------------------------------------------- total: 99.840000sec

filter_row:           98.130000   1.120000  99.250000 ( 99.311869)

After:
Rehearsal --------------------------------------------------------
filter_row:            2.180000   0.010000   2.190000 (  2.192253)
----------------------------------------------- total: 2.190000sec

filter_row:            2.180000   0.010000   2.190000 (  2.187111)
```

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -906,19 +906,19 @@ module Daru
     def filter_rows &block
       return to_enum(:filter_rows) unless block_given?
 
-      df = Daru::DataFrame.new({}, order: @vectors.to_a)
-      marked = []
+      values = to_a[0].map{|h| h.values}
+      index = @index.to_a
 
-      @index.each do |index|
-        keep_row = yield access_row(index)
-        marked << index if keep_row
+      arr_index = index.length - 1
+      index.reverse.each do |i|
+        keep_row = yield access_row(i)
+        unless keep_row
+          index.delete_at(i)
+          values.delete_at(i)
+        end
       end
 
-      marked.each do |idx|
-        df.row[idx] = self[idx, :row]
-      end
-
-      df
+      Daru::DataFrame.rows(values, {order: @vectors.to_a, index: index})
     end
 
     # Iterates over each vector and retains it in a new DataFrame if the block returns
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -909,7 +909,6 @@ module Daru
       values = to_a[0].map{|h| h.values}
       index = @index.to_a
 
-      arr_index = index.length - 1
       index.reverse.each do |i|
         keep_row = yield access_row(i)
         unless keep_row


lib/daru/dataframe.rb
```diff
@@ -906,19 +906,19 @@ def filter_vector vec
...
-        keep_row = yield access_row(index)
-        marked << index if keep_row
-      end
-
-      marked.each do |idx|
-        df.row[idx] = self[idx, :row]
+      values = to_a[0].map{|h| h.values}
+      index = @index.to_a
+
+      arr_index = index.length - 1
```
v0dro: This variable looks like it's not being used anywhere. Remove it please.

lib/daru/dataframe.rb
```diff
@@ -906,19 +906,19 @@ def filter_vector vec
...
+      index.reverse.each do |i|
```
v0dro: Why `reverse`? Wont this access rows in reverse order?

v0dro:
Comments in code.

lib/daru/dataframe.rb
```diff
@@ -906,19 +906,19 @@ def filter_vector vec
...
-      marked = []
-
-      @index.each do |index|
-        keep_row = yield access_row(index)
-        marked << index if keep_row
-      end
-
-      marked.each do |idx|
-        df.row[idx] = self[idx, :row]
+      values = to_a[0].map{|h| h.values}
```
v0dro: Could you try `self.to_matrix.to_a` to get rows in an Array of Arrays and see if it's faster? I have a hunch it might be.

lib/daru/dataframe.rb
peterktung: filter will miss deletes or delete the wrong rows if reverse is not used at we are deleting while traversing through the index.

lib/daru/dataframe.rb
peterktung: <code>to_matrix</code> only turns columns of numeric types to a matrix format.  I'm not sure if I can use this the way because <code>values</code> is used to build the dataframe at the end.

v0dro: Point. Leave it.

lib/daru/dataframe.rb
```diff
@@ -906,19 +906,18 @@ def filter_vector vec
...
+      index = @index.to_a
+
+      index.reverse.each do |i|
+        keep_row = yield access_row(i)
+        unless keep_row
+          index.delete_at(i)
```
v0dro: Over here you're assuming that the index is strictly numerical (consists only of integers from 0..n-1). If it turns out to be something else like a string, this will fail.

lib/daru/dataframe.rb
```diff
@@ -906,19 +906,18 @@ def filter_vector vec
...
-        keep_row = yield access_row(index)
-        marked << index if keep_row
-      end
-
-      marked.each do |idx|
-        df.row[idx] = self[idx, :row]
+      values = to_a[0].map{|h| h.values}
+      index = @index.to_a
+
+      index.reverse.each do |i|
```
v0dro: Any specific reason for reversing the array before iteration?

v0dro:
I'm sorry I was quite busy with work over the past month. Please find comments in the code.

v0dro:
Also, in the future, could you please rebase one feature into a single commit? It would help keep the commit tree clean.

PR #28, opened 2015-10-21
Make reportbuilder an optional dependency

bradleypriest:
The `reportbuilder` gem brings in amongst other things the entirety of prawn.
Technically this is a breaking change, but if that's a problem I can add a warning when the previous method is called

### DIFF
--- a/daru.gemspec
+++ b/daru.gemspec
@@ -49,7 +49,6 @@ Cheers!
 *************************************************************************
 EOF
 
-  spec.add_runtime_dependency 'reportbuilder', '~> 1.4'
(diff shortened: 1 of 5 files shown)


v0dro:
Why is the prawn dependency a problem? In my knowledge it uses up to date versions of all dependencies too.

Report building is a necessary feature for quick interactive statistical analysis.

bradleypriest:
I just like to keep the number of dependencies down in our app regardless of whether they're up to date.

We're doing the analysis without needing to output it into a CSV/PDF, if that's a minority usecase then I can understand

v0dro:
Yeah it's a minor use case :P 

Most people want to 'see' their data interactively, hence the report building. Closing this.

PR #30, opened 2015-11-09
Fix IO.from_sql

dansbits:
The IO.from_sql method was using the dbi column_info to create the vectors of the data frame. This is resulting in an error:

```RuntimeError: can't add a new key into hash during iteration```

I've updated it to use the column_names method instead which does not result in updating the object you're iterating through.

### DIFF
--- a/daru.gemspec
+++ b/daru.gemspec
@@ -60,8 +60,10 @@ EOF
   spec.add_development_dependency 'rspec'
   spec.add_development_dependency 'awesome_print'
   spec.add_development_dependency 'nyaplot', '~> 0.1.5'
-  spec.add_development_dependency 'nmatrix', '~> 0.1.0'
+  spec.add_development_dependency 'nmatrix', '~> 0.2.0'
   spec.add_development_dependency 'distribution', '~> 0.7'
   spec.add_development_dependency 'rb-gsl', '~>1.16'
   spec.add_development_dependency 'bloomfilter-rb', '~> 2.1'
+  spec.add_development_dependency 'dbd-sqlite3'
+  spec.add_development_dependency 'dbi'
 end
--- a/lib/daru/io/io.rb
+++ b/lib/daru/io/io.rb
@@ -120,24 +120,33 @@ module Daru
         writer.close
       end
 
-      # Loading/writing from SQL databases
+      # Execute a query and create a data frame from the result
+      #
+      # @param dbh [DBI::DatabaseHandle] A DBI connection to be used to run the query
+      # @param query [String] The query to be executed
+      #
+      # @return A dataframe containing the data resulting from the query
 
       def from_sql dbh, query
-        require 'dbi'
         sth     = dbh.execute(query)
         vectors = {}
-        fields  = []
-        sth.column_info.each do |c|
-          vectors[c[:name]] = Daru::Vector.new([])
-          vectors[c[:name]].rename c[:name]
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: daru.gemspec [hunk 1]
```diff
@@ -51,6 +51,7 @@ EOF
 
   spec.add_runtime_dependency 'reportbuilder', '~> 1.4'
   spec.add_runtime_dependency 'spreadsheet', '~> 1.0.3'
+  spec.add_runtime_dependency 'dbi'
```
v0dro: OK the rest seems fine but this runtime dependency on dbi will involve additional dependencies on daru itself, which I'd like to keep to a minimal.

What you can do is add it as a development dependency, and use something like the `Daru.has_gsl?` method before doing a `require` on dbi. You can throw an error asking the user to do `gem install dbi` if he wants to use the `from_sql` method.

See this code to see how you can add a `Daru.has_dbi?` method: https://github.com/v0dro/daru/blob/master/lib/daru.rb#L57

lib/daru/io/io.rb
```diff
@@ -124,20 +124,24 @@ def dataframe_write_csv dataframe, path, opts={}
...
+        df = Daru::DataFrame.new(vectors, order: fields)
+
         sth.fetch do |row|
-          ds.add_row(row.to_a)
+          df.add_row(row.to_a)
         end
-        ds.update
-        ds
+
+        df
```
v0dro: `df.update` will ensure that missing values are duly updated in the resulting dataframe. Removing it might lead to wrong results.

You can keep an option in the `from_sql` method, something like `:update_missing_values` and set the default to `true`. This can serve as a switch for updating the track of missing values that Daru::DataFrame keeps. This way people who dont care about the missing values being updated can disable the update from running.

An `if opts[:update_missing_values]` statement around the `df.update` will serve as the switch.

Also, don't forget to add documentation describing the new option. We follow YARD for docs. You can refer to other methods to see how it can be done.

spec/io/io_spec.rb
```diff
@@ -85,8 +85,26 @@
...
+      after { FileUtils.rm("daru_test") }
+
       it "loads data from an SQL database" do
-        # TODO: write these tests
+        accounts = subject
+        expect(accounts.class).to eq Daru::DataFrame
+        expect(accounts.nrows).to eq 2
+        expect(accounts.row[0][:id]).to eq 1
+        expect(accounts.row[0][:name]).to eq "Homer"
       end
```
v0dro: Tests look great :)

v0dro:
Comments in code.

Also, could you rebase this feature into a single commit?

daru.gemspec
dansbits: Moved this to a development dependency. I also removed the require 'dbi' from within the method. Because the user has to pass in a DBI::DatabaseHandle, it should be installed before calling from_sql.

lib/daru/io/io.rb
dansbits: I added a df.update. I didn't have any strong opinions about calling update, just wasn't sure if it was necessary.

dansbits:
@v0dro I believe this is all taken care of. I squashed my commits into one and updated the yard docs for both DataFrame.from_sql and IO.from_sql. Just let me know if you have any other comments.

v0dro:
Thanks! Merging.

Issue #34, opened 2015-11-23
group_by fails with nils


gnilrets:
Think it would be possible to support group-by when the value is nil?

Thanks!

````ruby
df = Daru::DataFrame.new(a: ["a",nil,"b","b","c"])
df.group_by([:a]).size
````
````
ArgumentError: comparison of Array with Array failed
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/daru-0.1.1/lib/daru/core/group_by.rb:21:in `sort'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/daru-0.1.1/lib/daru/core/group_by.rb:21:in `initialize'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/daru-0.1.1/lib/daru/dataframe.rb:1213:in `new'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/daru-0.1.1/lib/daru/dataframe.rb:1213:in `group_by'
<main>:1:in `<main>'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/backend.rb:44:in `eval'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/backend.rb:44:in `eval'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/backend.rb:12:in `eval'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/kernel.rb:87:in `execute_request'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/kernel.rb:47:in `dispatch'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/kernel.rb:37:in `run'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/command.rb:70:in `run_kernel'
/Users/gnilrets/git/itk/noodling/brdd/vendor/bundle/gems/iruby-0.2.7/lib/iruby/command.rb:34:in `run'
[...]

PR #37, opened 2015-12-12
FIX - group_by no longer fails when nils are present

gnilrets:
nils get sorted to the top.


gnilrets:
This should resolve #34

PR #38, opened 2015-12-12
FIX - DataFrame dup now duplicates the underlying vector arrays.

### LINKED ISSUE #36: Is dup working properly?
Given
````ruby
v1 = Daru::Vector.new(1.upto(5).to_a)
df1 = Daru::DataFrame.new(v1: v1)
````

I would expect that `dup` would create a new dataframe object with new vectors.  However, if I recode the vector in `df1`, the vector in `df2` is changed too.
````ruby
df2 = df1.dup

df1[:v1].recode! { |v| v * 2 }
df2[:v1] == df1[:v1]
````
Returns `true` when I would expect it to return `false`.

I can get the result I want if I `dup` each vector individually.  But this is cumbersome when I've got a larger dataframe.
````ruby
df2 = df1.dup
df2[:v1] = df1[:v1].dup

df1[:v1].recode! { |v| v * 2 }
df2[:v1] == df1[:v1]
````
Returns `false` as expected.

### PR DESCRIPTION
**Pull request by gnilrets**
Prior to this fix, if a dataframe was duplicated and a vector of the duplicated
dataframe was destructively recoded, the original dataframe would
also have a modified vector.

Resolves #36

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -14,30 +14,30 @@ module Daru
     include Daru::Plotting::DataFrame if Daru.has_nyaplot?
 
     class << self
-      # Load data from a CSV file. Specify an optional block to grab the CSV 
-      # object and pre-condition it (for example use the `convert` or 
+      # Load data from a CSV file. Specify an optional block to grab the CSV
+      # object and pre-condition it (for example use the `convert` or
       # `header_convert` methods).
-      # 
+      #
       # == Arguments
-      # 
+      #
       # * path - Path of the file to load specified as a String.
-      # 
+      #
       # == Options
-      # 
+      #
       # Accepts the same options as the Daru::DataFrame constructor and CSV.open()
       # and uses those to eventually construct the resulting DataFrame.
       #
       # == Verbose Description
       #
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/daru/dataframe.rb [hunk 1]
```diff
@@ -423,17 +423,17 @@ def row
...
+    #
+    # * +vectors_to_dup+ - An Array specifying the names of Vectors to
     # be duplicated. Will duplicate the entire DataFrame if not specified.
     def dup vectors_to_dup=nil
       vectors_to_dup = @vectors.to_a unless vectors_to_dup
 
       src = []
       vectors_to_dup.each do |vec|
-        src << @data[@vectors[vec]].to_a
+        src << @data[@vectors[vec]].dup.to_a
```
v0dro: If you do `vector.dup.to_a` you will first duplicate a `Daru::Vector` and convert it to an Array. A faster and smarter  way to do this would be `vector.to_a.dup` so that only a Ruby Array is being duplicated.

Please make this change before I can merge.

v0dro:
Also dont forget to rebase your commits.

gnilrets:
Fixed and rebased.

v0dro:
Great :)

PR #46, opened 2016-02-08
Exception handling for invalid index

### LINKED ISSUE #43: Vector#[]= with an unknown index
```ruby
v = Daru::Vector.new([1,2,3], index: [:a, :b, :c])
v[:a] = 999 # from docs, works 
v[:x] = 999 # fails with cryptic undefined method `each' for :x:Symbol
```

The error message displayed should precisely convey what went wrong.

lokeshh:
Fixes #43 

The problem is also with `v[:x]`:

```ruby
[2] pry(main)> v[:x]
NoMethodError: undefined method `map' for :x:Symbol
from /home/ubuntu/workspace/daru/lib/daru/vector.rb:242:in `[]'
```

Should I fix this?

### DIFF
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -220,6 +220,10 @@ module Daru
 
         return result
       else
+      raise TypeError, "Invalid index type #{location.inspect}.\
+        \nUsage: v[:a, :b] gives elements with keys :a and :b for vector v." \
+        if location.is_a? Array
+
         unless indexes[1]
           case location
           when Range
@@ -238,9 +242,13 @@ module Daru
           indexes = indexes.map { |e| named_index_for(e) }
         end
 
-        Daru::Vector.new(
-          indexes.map { |loc| @data[@index[loc]] }, 
-          name: @name, index: indexes, dtype: @dtype)
+        begin
+          Daru::Vector.new(
+            indexes.map { |loc| @data[@index[loc]] },
+            name: @name, index: indexes, dtype: @dtype)
+        rescue NoMethodError
+          raise IndexError, "Specified index #{pos.inspect} does not exist."
+        end
       end
     end
 
@@ -271,7 +279,11 @@ module Daru
       if pos.is_a?(Numeric)
         @data[pos] = value
       else
-        pos.each { |tuple| self[tuple] = value }
+        begin
+          pos.each { |tuple| self[tuple] = value }
+        rescue NoMethodError
+          raise IndexError, "Specified index #{pos.inspect} does not exist."
+        end
       end
 
       set_size


v0dro:
Yes go ahead with your fix. Looks good.

lib/daru/vector.rb
```diff
@@ -238,9 +238,15 @@ def [](*indexes)
...
-          indexes.map { |loc| @data[@index[loc]] }, 
-          name: @name, index: indexes, dtype: @dtype)
+        begin
+          Daru::Vector.new(
+            indexes.map { |loc| @data[@index[loc]] },
+            name: @name, index: indexes, dtype: @dtype)
+        rescue NoMethodError
+          raise IndexError, "Specified index #{pos.inspect} does not exist."
+        rescue TypeError
+          raise IndexError, "One or more of specified indices #{pos.inspect} do not exist."
```
lokeshh: NoMethodError results with `v[:x]` and TypeError with `v[[:x, :y]]`, where `:x` is not a valid index.

v0dro: The correct way to get both `:x` and `:y` is `v[:x, :y]`.

So `v[[:x,:y]]` should return a TypeError. Maybe you can handle the type error differently and provide a more verbose output for the same.

lokeshh: @v0dro I am confused here. `v[[:x, :y]]` does not return a TypeError and nor does `v[[[:x, :y]]]` if `:x` and `:y` are valid indexes. So, should I treat them as wrong input types?

v0dro: Yes, an Array like `[:x, :y]` passed in place of plain old `:x, :y` will be wrong input. A check for this which raises a TypeError should fix it.

lokeshh: @v0dro The documentation says:

```ruby
    # == Usage
    #   # For vectors employing single layer Index
    # 
    #   v[:one, :two] # => Daru::Vector with indexes :one and :two
    #   v[:one]       # => Single element
    #   v[:one..:three] # => Daru::Vector with indexes :one, :two and :three
```
It would be counter intuitive if we let `[:x, :y]` fail and let `1..3` as right input as given in the documentation because both `[:x, :y]` and `1..3` are Enumerable. What do you say?

v0dro: I'm not saying we should let `[:x, :y]` to fail. I'm saying that say you have a Vector `vector` with indexes `:a, :b, :c and :y`:
``` ruby
vector = Daru::Vector.new([1,2,3,4], index: [:a, :b, :c, :y])
```

Now say I wrongly say I want indexes `:x` and `:y`, I should say it like this:
``` ruby
vector[:x, :y]
```

And NOT like this:
``` ruby
vector[[:x, :y]] # notice Array inside #[] method
```

Therefore, the second way (i.e. `[:x, :y]`) should raise a `TypeError` because the `#[]` method is not supposed to accept Arrays as inputs.

lib/daru/vector.rb
```diff
@@ -220,6 +220,10 @@ def [](*indexes)
 
         return result
       else
+      raise TypeError, "Invalid index type #{location.inspect}.\
+        \nUsage: v[:a, :b] gives elements with keys :a and :b for vector v." \
+        if location.is_a? Array
+
```
lokeshh: How about now? It checks for `v[[:a, :b]]` and return a TypeError like below:

```ruby
> v[[:a, :b]]
TypeError: Invalid index type [:a, :b].        
Usage: v[:a, :b] gives elements with keys :a and :b for vector v.
```

v0dro: Works. Can you rebase into a single commit?

Also I'm a bit tied down currently with the SciRuby GSOC application so please pardon me for late replies. I'll back to you whenever possible.

PR #47, opened 2016-02-10
Set name of first column

### LINKED ISSUE #41: DataFrame.from_excel name of first column nil
To replicate the problem:

```ruby
require 'daru'
# Create DataFrame
df = Daru::DataFrame.new({
  'col0' => [1,2,3,4,5,6],
  'col2' => ['a','b','c','d','e','f'],
  'col1' => [11,22,33,44,55,66]
  }, 
  index: ['one', 'two', 'three', 'four', 'five', 'six'], 
  order: ['col0', 'col1', 'col2']
)
# Check first column name
df['col0'].name
#"col0"

# Write an excel with DataFrame
df.write_excel('df.xls')
# Create DataFrame drom excel
df_xls = Daru::DataFrame.from_excel('df.xls')

# Problem check first column name
df_xls[:col0].name
# nil

# Other column name and vectors are ok
df_xls.vectors.to_a
# [:col0, :col1, :col2]
```

lokeshh:
Fixes #41 

Fixed the issue by passing name parameter while initializing the vector.

### DIFF
--- a/spec/dataframe_spec.rb
+++ b/spec/dataframe_spec.rb
@@ -521,6 +521,15 @@ describe Daru::DataFrame do
           }.to raise_error
       end
 
+    it "assigns correct name given empty dataframe" do
+      df_empty = Daru::DataFrame.new({})
+      df_empty[:a] = 1..5
+      df_empty[:b] = 1..5
+
+      expect(df_empty[:a].name).to equal(:a)
+      expect(df_empty[:b].name).to equal(:b)
+    end
+
       it "appends multiple vectors at a time" do
         # TODO
       end
@@ -560,6 +569,13 @@ describe Daru::DataFrame do
         @df_mi[:c,:one,:bar] = [100,200,300,400,100,200,300,400,100,200,300,400]
 
         expect(@df_mi).to eq(answer)
+      end
+
+      it "assigns correct name given empty dataframe" do
+        df_empty = Daru::DataFrame.new([], index: @multi_index, order: @order_mi)
+        df_empty[:c, :one, :bar] = 1..12
+
+        expect(df_empty[:c, :one, :bar].name).to eq "conebar"
       end
     end
   end
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -2204,7 +2204,8 @@ module Daru
       v = nil
 
       if @index.empty?
-        v = vector.is_a?(Daru::Vector) ? vector : Daru::Vector.new(vector.to_a)
+        v = vector.is_a?(Daru::Vector) ? vector :
+                            Daru::Vector.new(vector.to_a, name: set_name(name))
         @index = v.index
         assign_or_add_vector name, v
         set_size


v0dro:
Thanks! Could you also add a test with the example pointed out in the issue? That way we'll be sure that a future fix does not break anything.

lokeshh:
I didn't used the example pointed in the issue because the bug wasn't in reading or writing excel file but in row assigning when the DF was empty. So, I decided to add the test to `[]=`. Is that fine?

I was confused as to where to add test in `[]=` because there was `before do...end` block initializing a non-empty DF and the test I wanted could not have been appended into that. Therefore I created another `context do...end` block. What do you think?

v0dro:
Here's a better and cleaner way:

``` ruby

context "#[]=" do # this already exists
  it "assigns correct name given empty dataframe" do
    # put all tests here including that before block
    # ...
  end
end
```

lokeshh:
Where should I place `context Daru::Index` and `context Daru::MultiIndex`? I did exactly like you said, it gives error saying `context` block cannot be inside `it` error. This is what I did:

```ruby
  context "#[]=" do
    context "assigns correct name given empty dataframe" do
      @df_empty = Daru::DataFrame.new({})
      @df_empty[:a] = (1..5)
      @df_empty[:b] = (1..5)
      
      @df_empty.expect(@df_empty[:a].name).to equal(:a)
      @df_empty.expect(@df_empty[:b].name).to equal(:b)

      context Daru::Index do
        # All the tests under here
      end

      context Daru::MultiIndex do
        # All the tests under here
      end
    end
```

v0dro:
Write the test this way at [line 523](https://github.com/v0dro/daru/blob/master/spec/dataframe_spec.rb#L523):

``` ruby
    it "assigns correct name given empty dataframe" do
      df_empty = Daru::DataFrame.new({})
      df_empty[:a] = (1..5)
      df_empty[:b] = (1..5)

      df_empty.expect(df_empty[:a].name).to equal(:a)
      df_empty.expect(df_empty[:b].name).to equal(:b)
  end
```

Also put a test at [line 563](https://github.com/v0dro/daru/blob/master/spec/dataframe_spec.rb#L563) to handle the MultiIndex case. You can concatenate the names of a hierarchical index from left to right and assign that as the name here.

v0dro:
Don't forget to rebase!

spec/dataframe_spec.rb
```diff
@@ -561,6 +570,13 @@
...
         expect(@df_mi).to eq(answer)
       end
+
+      it "assigns correct name given empty dataframe" do
+        df_empty = Daru::DataFrame.new([], index: @multi_index, order: @order_mi)
+        df_empty[:c, :one, :bar] = 1..12
+
+        expect(df_empty[:c, :one, :bar].name).to eq "conebar"
+      end
     end
```
lokeshh: @v0dro I am wondering whether this test is any useful or not because the bug that was there had to deal with assigning correct name to column when DF didn't had any index built. In case of Multi Index, the DF cannot be created without first specifying the index, so the test which I had put up for Multi Index doesn't has anything to do with the bug. If I could built a Multi Index DF without assigning the index first, then I think the test would be useful.

v0dro: Ah makes sense. I had forgotten that. My bad.

But it's good to have a check anyway.

Issue #39, opened 2015-12-12
Sorting vectors and dataframes with nils


gnilrets:
I was starting to work on figuring out a solution for sorting vectors and data frames with nil values.  I noticed you've got a pending test - https://github.com/v0dro/daru/blob/master/spec/vector_spec.rb#L488.

I'd argue that nils should be placed at the beginning of the vector.  
1. SQL sorts nulls at the beginning.
2. Empty ('') and blank ('   ') strings are functionally similar to nil when dealing with datasets, and they sort before all other strings.  It would be strange to have empty/blank strings at the top and nils at the bottom.
3. nil data is usually "special" and if it's at the top of a sort, you're more likely to notice.

v0dro:
Hmmm that makes sense. I'm good with placing nils at the top.

Additionally, sorting currently uses a handmade Ruby variant of quick sort, which makes it slower than say `Array#sort`. If you have a solution to doing this faster (maybe use `Array#sort` somewhere so things happen at C level and are fast) while keeping the functionality same, I'm all ears!

gnilrets:
Why do you use a handmade variant of quicksort rather than Ruby's native sort?

v0dro:
The problem was that I couldn't figure out of a way to change the index along with the data, hence I went ahead with a hand made quick sort so that control could be maintained over the index objects as well.

Now that I think of it, I think zipping the indexes and data into tuples would do the trick, but I'm not sure if this can be extended to DataFrame.

lokeshh:
I think `[sort_by](ruby-doc.org/core-2.0.0/Enumerable.html#method-i-sort_by)` could be used to replace handmade variant. Isn't it?

v0dro:
Maybe only for `Daru::Vector`. DataFrame requires sorting only those specific columns that too in a nested manner.

lokeshh:
> Now that I think of it, I think zipping the indexes and data into tuples would do the trick, but I'm not sure if this can be extended to DataFrame.

I went ahead and took the liberty to implement your idea. There's a speed up.

> DataFrame requires sorting only those specific columns that too in a nested manner.

I am not sure what that means.

lokeshh:
> DataFrame requires sorting only those specific columns that too in a nested manner.

[...]

PR #52, opened 2016-02-15
FIX - Vector sort now sorts nil at the beginning

gnilrets:


gnilrets:
Resolves #39.  

So this fixes the issue but I'm not sure it's the best solution.  The problem is that if the user wanted to write their own sorter (by passing a block to sort), they'd still have to deal with the nil issue.  An alternative to the solution in this commit might be to add refinements to many of the core classes (NilClass, Fixnum, Float, etc) that allowed them to be sortable when nils are involved.  

````ruby
class NilClass
  include Comparable
  def <=>(other)
    return 0 if other.nil?
    -1
  end
end
````

That, of course, comes with its own set of issues.  Any thought on this?

gnilrets:
Also, it looks like my editor is removing trailing whitespace (which is good according to https://github.com/bbatsov/ruby-style-guide#no-trailing-whitespace).

v0dro:
Good fix. My only concern is speed. Will this slow things down significantly? The problem is that rewriting sorting in Ruby is slow as it is. The `if` conditions should not slow it down further.

gnilrets:
This commit should slightly reduce the number of evaluations that need to happen when your data doesn't have many nils.  Not sure if we can do much better.

v0dro:
This will work for now. Thanks. Merging.

PR #54, opened 2016-02-19
Refactor indexing

lokeshh:
Improve the current indexing infrastructure of Daru

### DIFF
--- a/lib/daru/core/group_by.rb
+++ b/lib/daru/core/group_by.rb
@@ -194,8 +194,15 @@ module Daru
         indexes.each do |idx|
           rows << transpose[idx]
         end
+
+        new_index =
+          begin
+            @context.index[indexes]
+          rescue IndexError
+            indexes
+          end
         Daru::DataFrame.rows(
-          rows, index: @context.index[indexes], order: @context.vectors)
+          rows, index: new_index, order: @context.vectors)
       end
 
      private
--- a/lib/daru/index.rb
+++ b/lib/daru/index.rb
@@ -90,7 +90,10 @@ module Daru
         end
       else
         v = @relation_hash[loc]
-        return loc if v.nil?
+        if !v
+          return loc if loc.is_a? Numeric and loc < size
+          raise IndexError, "Specified index #{loc.inspect} does not exist"
+        end
         v
       end
     end
@@ -228,7 +231,12 @@ module Daru
       case
       when key[0].is_a?(Range) then retrieve_from_range(key[0])
       when (key[0].is_a?(Integer) and key.size == 1) then try_retrieve_from_integer(key[0])
-      else retrieve_from_tuples key
+      else
+        begin
+          retrieve_from_tuples key
+        rescue NoMethodError
+          raise IndexError, "Specified index #{key.inspect} do not exist"
+        end
       end
     end
 
--- a/lib/daru/index.rb
+++ b/lib/daru/index.rb
@@ -81,7 +81,13 @@ module Daru
 
         slice first, last
       when key.size > 1
-        Daru::Index.new key.map { |k| self[k] }
+        if include? key[0]
+          Daru::Index.new key.map { |k| k }
+        else
+          # Assume the user is specifing values for index not keys
+          # Return index object having keys corresponding to values provided
+          Daru::Index.new key.map { |k| key k }
+        end
       else
         v = @relation_hash[loc]
         return loc if v.nil?
@@ -142,6 +148,14 @@ module Daru
       h = Marshal.load data
 
       Daru::Index.new(h[:relation_hash].keys)
+    end
+
+    # Provide an Index for sub vector produced
+    #
+    # @param input_indexes [Array] the input by user to index the vector
+    # @return [Object] the Index object for sub vector produced
+    def conform input_indexes
+      self
     end
   end # class Index
 
@@ -214,7 +228,7 @@ module Daru
       case
       when key[0].is_a?(Range) then retrieve_from_range(key[0])
       when (key[0].is_a?(Integer) and key.size == 1) then try_retrieve_from_integer(key[0])
-      else retrieve_from_tuples(key)
+      else retrieve_from_tuples key
       end
     end
 
@@ -236,7 +250,7 @@ module Daru
         chosen = find_all_indexes label, level_index, chosen
       end
 
-      return chosen[0] if chosen.size == 1
+      return chosen[0] if chosen.size == 1 and key.size == @levels.size
       return multi_index_from_multiple_selections(chosen)              
     end
 
@@ -329,6 +343,15 @@ module Daru
 
     def inspect
       "Daru::MultiIndex:#{self.object_id} (levels: #{levels}\nlabels: #{labels})"
+    end
+
+    # Provide a MultiIndex for sub vector produced
+    #
+    # @param input_indexes [Array] the input by user to index the vector
+    # @return [Object] the MultiIndex object for sub vector produced
+    def conform input_indexes
+      return self if input_indexes[0].is_a? Range
+      drop_left_level input_indexes.size
     end
   end
 end
--- a/spec/dataframe_spec.rb
+++ b/spec/dataframe_spec.rb
@@ -455,7 +455,7 @@ describe Daru::DataFrame do
       end
 
       it "returns a Vector if the last level of MultiIndex is tracked" do
-        expect(@df_mi[:a, :one]).to eq(
+        expect(@df_mi[:a, :one, :bar]).to eq(
           Daru::Vector.new(@vector_arry1, index: @multi_index))
       end
     end
--- a/spec/index_spec.rb
+++ b/spec/index_spec.rb
@@ -76,7 +76,7 @@ describe Daru::Index do
   context "#[]" do
     before do
       @id = Daru::Index.new [:one, :two, :three, :four, :five, :six, :seven]
-      @mixed_id = Daru::Index.new ['a','b','c',:d,:a,0,3,5]
+      @mixed_id = Daru::Index.new ['a','b','c',:d,:a,8,3,5]
     end
 
     it "works with ranges" do
@@ -88,12 +88,12 @@ describe Daru::Index do
       expect(@mixed_id[0..2]).to eq(Daru::Index.new(['a','b','c']))
 
       # If atleast one is a number then refer to actual indexing
-      expect(@mixed_id.slice('b',0)).to eq(Daru::Index.new(['b','c',:d,:a,0]))
+      expect(@mixed_id.slice('b',8)).to eq(Daru::Index.new(['b','c',:d,:a,8]))
     end
 
     it "returns multiple keys if specified multiple indices" do
-      expect(@id[0,1,3,4]).to eq(Daru::Index.new([0,1,3,4]))
-      expect(@mixed_id[0,5,3,2]).to eq(Daru::Index.new([5, 7, 6, 2]))
+      expect(@id[0,1,3,4]).to eq(Daru::Index.new([:one, :two, :four, :five]))
+      expect(@mixed_id[0,5,3,2]).to eq(Daru::Index.new(['a', 8, :d, 'c']))
     end
 
     it "returns correct index position for non-numeric index" do
@@ -102,7 +102,7 @@ describe Daru::Index do
     end
 
     it "returns correct index position for mixed index" do
-      expect(@mixed_id[0]).to eq(5)
+      expect(@mixed_id[8]).to eq(5)
       expect(@mixed_id['c']).to eq(2)
     end
   end
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -177,11 +177,12 @@ describe Daru::Vector do
               [:c,:one,:bar],
               [:c,:one,:baz],
               [:c,:two,:foo],
-              [:c,:two,:bar]
+              [:c,:two,:bar],
+              [:d,:one,:foo]
             ]
             @multi_index = Daru::MultiIndex.from_tuples(@tuples)
             @vector = Daru::Vector.new(
-              Array.new(12) { |i| i }, index: @multi_index,
+              Array.new(13) { |i| i }, index: @multi_index,
               dtype: dtype, name: :mi_vector)
           end
 
@@ -208,6 +209,12 @@ describe Daru::Vector do
               [:foo],
               [:bar]])
             expect(@vector[:c,:two]).to eq(Daru::Vector.new([10,11], index: mi,
+              dtype: dtype, name: :sub_sub_vector))
+          end
+
+          it "returns sub vector not a single element when passed the partial tuple" do
+            mi = Daru::MultiIndex.from_tuples([[:foo]])
+            expect(@vector[:d, :one]).to eq(Daru::Vector.new([12], index: mi,
               dtype: dtype, name: :sub_sub_vector))
           end
 
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -161,6 +161,11 @@ describe Daru::Vector do
             expect(v[1]).to eq(3)
             expect(v[0]).to eq(1)
           end
+
+          it "raises exception for invalid index" do
+            expect { @dv[:foo] }.to raise_error(IndexError)
+            expect { @dv[:obi, :foo] }.to raise_error(IndexError)
+          end
         end
 
         context Daru::MultiIndex do
@@ -230,6 +235,11 @@ describe Daru::Vector do
             ])
             expect(@vector[3..9]).to eq(Daru::Vector.new([3,4,5,6,7,8,9], index: mi,
               dtype: dtype, name: :slice))
+          end
+
+          it "raises exception for invalid index" do
+            expect { @vector[:foo] }.to raise_error(IndexError)
+            expect { @vector[:a, :two, :foo] }.to raise_error(IndexError)
           end
         end
       end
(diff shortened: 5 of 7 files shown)


lib/daru/index.rb
```diff
@@ -81,7 +81,13 @@ def [](*key)
...
       when key.size > 1
-        Daru::Index.new key.map { |k| self[k] }
+        if include? key[0]
+          Daru::Index.new key.map { |k| k }
+        else
+          # Assume the user is specifing values for index not keys
+          # Return index object having keys corresponding to values provided
+          Daru::Index.new key.map { |k| key k }
+        end
       else
```
lokeshh: @v0dro I am on my way to improve the indexing infrastructure. I have made a major change in Index behavior. I think an example will illustrate it best:

```ruby
[3] pry(main)> i = Daru::Index.new([:a, :b, :c])
=> #<Daru::Index:0x00000003f5ca50
 @keys=[:a, :b, :c],
 @relation_hash={:a=>0, :b=>1, :c=>2},
 @size=3>
[4] pry(main)> i[0, 2]
=> #<Daru::Index:0x00000003ed14f0
 @keys=[:a, :c],
 @relation_hash={:a=>0, :c=>1},
 @size=2>
```

Earlier it used to be:

```ruby
[1] pry(main)> i = Daru::Index.new([:a, :b, :c])
=> #<Daru::Index:0x000000031f89b8
 @keys=[:a, :b, :c],
 @relation_hash={:a=>0, :b=>1, :c=>2},
 @size=3>
[2] pry(main)> i[0, 1]
=> #<Daru::Index:0x000000031bccd8
 @keys=[0, 1],
 @relation_hash={0=>0, 1=>1},
 @size=2>
```

The motive behind this is to remove the frequent checking of type of index and moving the functionality of guessing that the user is perhaps giving the index values (not keys) from `Vector` class to `Index` class because it seems more inherent to index. 

Are you good with this?

If you are good with this then I plan to the same with `MultiIndex` and finally remove the conditionals.

spec/index_spec.rb
```diff
@@ -76,7 +76,7 @@
   context "#[]" do
     before do
       @id = Daru::Index.new [:one, :two, :three, :four, :five, :six, :seven]
-      @mixed_id = Daru::Index.new ['a','b','c',:d,:a,0,3,5]
+      @mixed_id = Daru::Index.new ['a','b','c',:d,:a,8,3,5]
```
v0dro: any particular reason why you replaced the 0 with 8?

v0dro:
Hey could you please pull the latest master and make your changes on top of that?

spec/index_spec.rb
lokeshh: The `0` was a key as well as index value, and since the first input that is `0` was a key the new code treated the whole input as key values rather than index values during [this test](https://github.com/v0dro/daru/pull/54/files#diff-4dd64e5576c9862a56edb0dbbab51831L96), that's why I had to change it to `8` so that the code treats input as index values.

v0dro:
I like your solution. But are you sure this won't break any other functionality?

Also, we need to make sure that data can be accessed by both index value and element position value. Ranges should be supported too.

If your solution fits into all this AND makes indexing more extensible and less fragile than it is now, do go ahead with the implementation.

spec/index_spec.rb
v0dro: Alright. But keep in mind that if `0` exists in the named Index, the Index must return the corresponding element position.

lokeshh:
Yes, this is more extensible and less fragile than earlier. Look at this:

```ruby
    def [](*indexes)
      # Get a proper index object
      indexes = @index[*indexes]

      # If one object is asked return it
      if indexes.is_a? Numeric
        return @data[indexes]
      end

      # Form a new Vector using indexes and return it
      Daru::Vector.new(
        indexes.map { |loc| @data[@index[loc]] },
        name: @name, index: indexes.factor_out, dtype: @dtype)
    end
```
This approach nicely encapsulates the general behavior. Any new index would simply need follow these rules and it's good to go. And this approach does no harm to access element by index value or element position value. The only thing is that now Index holds this responsibility whether user gave index value of element position value.

spec/index_spec.rb
lokeshh: Could you give an example?

v0dro: Example 1. Notice that since `0` is present in the index, daru should not return the _0th element_ but rather the element at index `0`. This applies for ranges too.
``` ruby
[18] pry(main)> a = Daru::Vector.new([1,2,3,4], index: [2,3,0,1])
=> 
#<Daru::Vector:88293040 @name = nil @size = 4 >
    nil
  2   1
  3   2
  0   3
  1   4

[19] pry(main)> a[0]
=> 3

```

Example 2. Here 0 isn't present in the index so it returns the _0th element_:
``` ruby

[24] pry(main)> b = Daru::Vector.new([4,3,5,9], index: [:a, :b, :c, :d])
=> 
#<Daru::Vector:87873110 @name = nil @size = 4 >
    nil
  a   4
  b   3
  c   5
  d   9

[25] pry(main)> b[0]
=> 4

```

v0dro:
So the `indexes` Array stores all indexes which the user wants.

The problem is that when you call `#map` on the `indexes` object, it again refers to `@index` to get the actual location in the data. So say for example `indexes` contains [0,1,5], you'll be passing that into `@index`, which would produce a wrong result.

lokeshh:
Since I am not sure what you mean, could you please provide an example which you think would produce a wrong result? That would be very helpful.

spec/index_spec.rb
lokeshh: My code is good with these tests!

lokeshh:
If I get what you meant then I think you meant that for a vector `v = Daru::Vector.new([1, 2, 3, 4, 5], index: [:a, :b, :c, :d, :e])`, `v[0, 1, 4]` would give wrong output. But this is not the case. Here's the output:
```ruby
[4] pry(main)> v[0, 1, 4]
=> 
#<Daru::Vector:22398300 @name = nil @size = 3 >
    nil
  a   1
  b   2
  e   5
```
The only place where this code falls short of I think is with differentiating whether `[0, 1, 4]` is position index or key values (and too when 0 is also an index). I think this is not hard to fix once I am clear about the desired functionality.

spec/index_spec.rb
v0dro: No problem then :)

lib/daru/index.rb
```diff
@@ -143,6 +149,11 @@ def self._load data
 
       Daru::Index.new(h[:relation_hash].keys)
     end
+
+    def factor_out
```
v0dro: I can't find any usage of this method in your code. What exactly do you plan to make it do?

lib/daru/index.rb
```diff
@@ -330,5 +341,11 @@ def values
     def inspect
       "Daru::MultiIndex:#{self.object_id} (levels: #{levels}\nlabels: #{labels})"
     end
+
+    def factor_out
```
v0dro: I think this method will modify the MultiIndex data structures for the purpose of displaying.

Moreover, better displaying of the MultiIndex is an issue in itself and not a part of revamping index infra.

v0dro:
My bad. I had thought that `Daru::Index` returns actual element positions with the `#[]` operator.

Your PR looks good the way it is apart from `factor_out`. Are you done with this feature?

lib/daru/index.rb
```diff
@@ -330,5 +341,11 @@ def values
...
+      # Remove levels not needed for display
```
v0dro: I don't quite understand this comment.

lib/daru/index.rb
```diff
@@ -330,5 +341,11 @@ def values
...
+      useless = levels.index { |l| l.length > 1 }
```
v0dro: Verbose variable naming please. Useless what?

lib/daru/index.rb
```diff
@@ -330,5 +341,11 @@ def values
...
+      drop_left_level useless
+    end
```
v0dro: I can see that you've used this method to remove the left most level on [this line](https://github.com/v0dro/daru/pull/54/files#diff-8129d70abcb202c11af77e9269873527R214) and have kept it here for a uniform interface between indexes.

But according to the previous functionality, the number of levels that were removed were equal to the [number of levels](https://github.com/v0dro/daru/pull/54/files#diff-8129d70abcb202c11af77e9269873527L215) that were specified by the user.

For example:
``` ruby
d =        Daru::MultiIndex.from_tuples([
  [:c,:one,:bar],          
  [:c,:one,:baz],          
  [:c,:two,:foo],          
  [:c,:two,:bar]          
])            
#=> Daru::MultiIndex:96634260 (levels: [[:c], [:one, :two], [:bar, :baz, :foo]]
#labels: [[0, 0, 0, 0], [0, 0, 1, 1], [0, 1, 2, 0]])
v = Daru::Vector.new([1,2,3,4], index: d)
#<Daru::Vector:101434370 @name = nil @size = 4 >
#                              nil
#[:c, :one, :bar]                1
#[:c, :one, :baz]                2
#[:c, :two, :foo]                3
#[:c, :two, :bar]                4

v[:c, :one]
#<Daru::Vector:101291480 @name = nil @size = 2 >
#          nil
#[:bar]      1
#[:baz]      2

v[:c]
#=> 
#<Daru::Vector:101176290 @name = nil @size = 4 >
#                      nil
#[:one, :bar]            1
#[:one, :baz]            2
#[:two, :foo]            3
#[:two, :bar]            4
```

I don't quite understand how the correct number of levels are being dropped in spite of you nowhere specifying the number of levels the user has asked for.

v0dro: Also `factor_out` is not a very verbose method name. Can you make it more specific? Also add documentation?

lokeshh: > I don't quite understand how the correct number of levels are being dropped in spite of you nowhere specifying the number of levels the user has asked for.

`factor_out` method drops all levels from left until a level with size greater than 1 is encountered.
For example
```ruby
[17] pry(main)> d = Daru::MultiIndex.from_tuples([[:a, :one], [:a, :two]])
=> Daru::MultiIndex:20963520 (levels: [[:a], [:one, :two]]
labels: [[0, 0], [0, 1]])
```
Here `[:a]` level got dropped because it's size was 1. But there's a little problem with this. It's a little inconsistent. For example:
```ruby
[21] pry(main)> d = Daru::MultiIndex.from_tuples([[:a, :one], [:b, :two]])
=> Daru::MultiIndex:22630440 (levels: [[:a, :b], [:one, :two]]
labels: [[0, 1], [0, 1]])
[22] pry(main)> v = Daru::Vector.new([1,2], index: d)
=> 
#<Daru::Vector:22281380 @name = nil @size = 2 >
                  nil
[:a, :one]          1
[:b, :two]          2

[23] pry(main)> v[:a]
=> 1
```
Notice, it never showed the `[:one]`.

I checkout out the previous functionality and it too didn't showed `[:one]`.

If I want to correct this inconsistency I think I would need to pass arguments to the `factor_out` how much levels to remove. There's another way to do this by not involving passing arguments to the `factor_out` method and that is to save inside the Index object how much levels user wants to remove at the time of (indexes = @index[*indexes]) and use it to then display the result. Thoughts?

v0dro: You just found a bug!

Yes you will absolutely have to pass the number of levels to be dropped because when a user specifies a given number of indexes for a MultiIndex vector/dataframe, only those many levels should be dropped.

Your previous approach worked mainly because the specs were so written. What if the levels were like this:
``` ruby
levels: [[:a], [:one], [:one, :two, :three, :four]]
```
And I have specified only `v[:a]` when calling the vector?
Don't rely only on the tests, they're there just as a reference.

v0dro: And about showing the `:one`; if I remember correctly, I had ported that functionality from Python's pandas. I think they follow the same convention if there's only one element with top level index `:a`.

lokeshh: @v0dro No, they don't follow the same convention. I think there's a bug in our implementation.

```python
>>> index = pd.MultiIndex.from_tuples([['a', 'one'], ['b', 'two']], names=['first', 'second'])
>>> s = pd.Series(np.random.randn(8), index=index)
>>> s
first  second
a      one       1.387915
b      two      -1.169805
dtype: float64
>>> s['a']
second
one    1.387915
dtype: float64
```

v0dro: Ah. Port the pandas functionality then.

lib/daru/index.rb
```diff
@@ -330,5 +342,12 @@ def values
     def inspect
       "Daru::MultiIndex:#{self.object_id} (levels: #{levels}\nlabels: #{labels})"
     end
+
+    def display input_indexes
+      # Remove levels not needed for display
+      return self if input_indexes[0].is_a? Range
+      remove_levels = input_indexes.length
+      drop_left_level remove_levels
```
v0dro: write this as `drop_left_level input_indexes.size`. No need for the extra variable.

lib/daru/index.rb
lokeshh: I just it fixed it. Now it's showing `:one`.

lib/daru/index.rb
```diff
@@ -330,5 +342,12 @@ def values
     def inspect
       "Daru::MultiIndex:#{self.object_id} (levels: #{levels}\nlabels: #{labels})"
     end
+
+    def display input_indexes
```
v0dro: The way this function is being called, it does not _display_ anything. It merely creates another index which conforms to a new requirement. The name you have given would have been legit if it were actually PRINTING something to STDOUT.

`conform` would be a better name.

v0dro: Don't forget to add docs. Looks good otherwise.

lib/daru/index.rb
v0dro: Yes absolutely.

lib/daru/index.rb
lokeshh: Where to add docs?

v0dro: Right above the method. Refer to elsewhere in code to see docs. If you have any questions, see the guides for YARD, that's what we use in daru: http://yardoc.org/

lokeshh:
Most of the work I think is done. Just a couple of things are left now like

- give a nice message when index specified is/are invalid. I think broke this functionality while improving `[]` method.
- refactor code for `#[]=` like `[]`.

spec/dataframe_spec.rb
```diff
@@ -455,7 +455,7 @@
       end
 
       it "returns a Vector if the last level of MultiIndex is tracked" do
-        expect(@df_mi[:a, :one]).to eq(
+        expect(@df_mi[:a, :one, :bar]).to eq(
           Daru::Vector.new(@vector_arry1, index: @multi_index))
```
lokeshh: It wasn't tracking the last level.

spec/vector_spec.rb
```diff
@@ -177,11 +177,12 @@
               [:c,:one,:bar],
               [:c,:one,:baz],
               [:c,:two,:foo],
-              [:c,:two,:bar]
+              [:c,:two,:bar],
+              [:d,:one,:foo]
             ]
```
lokeshh: Had to add `[:d, :one, :foo]` to test a special case i.e. single index exist under second layer.

v0dro:
You can revamp the whole index infra for `[]` and `[]=` in this PR itself. Just remember to keep one feature per commit so our working tree remains clean and readable.

lokeshh:
`[]=` falls in place every nicely. You would love this. Earlier it used to be:
```ruby
[1] pry(main)> v = Daru::Vector.new([1,2,3,4], index: [:a, :b, :c, :d])
=> 
#<Daru::Vector:20516460 @name = nil @size = 4 >
    nil
  a   1
  b   2
  c   3
  d   4

[2] pry(main)> v[:a, :b, :c] = 20
=> 20
[3] pry(main)> v
=> 
#<Daru::Vector:20516460 @name = nil @size = 4 >
    nil
  a  20
  b   2
  c   3
  d   4
```
But now it's
```ruby
[2] pry(main)> v[:a, :b, :c] = 20
=> 20
[3] pry(main)> v
=> 
#<Daru::Vector:11722400 @name = nil @size = 4 >
    nil
  a  20
  b  20
  c  20
  d   4
```

lib/daru/core/group_by.rb
```diff
@@ -194,8 +194,15 @@ def get_group group
...
+        new_index =
+          begin
+            @context.index[indexes]
+          rescue IndexError
+            indexes
+          end
         Daru::DataFrame.rows(
-          rows, index: @context.index[indexes], order: @context.vectors)
+          rows, index: new_index, order: @context.vectors)
       end
```
lokeshh: Had no option but to do this because I made Index object return an error whenever invalid indexes are supplied. This really helps in keeping things simple. Moreover I think Index should have this responsibility to raise Exception when invalid indexes are supplied.

v0dro: Yes it does. Good fix.

v0dro:
This looks great! Isn't it necessary to change anything in dataframe `[]` to reflect the changes?

Will merge once you confirm.

lokeshh:
I don't think it's necessary to change anything in dataframe but a lot of things can be certainly improved. I am currently understanding how the dataframe works inorder to see how it can improved given the changes in the index. Apart from that there's very little chance that something would go wrong with dataframe as all the tests are passing but I haven't taken a thorough look since I do not yet fully understand how it works.

v0dro:
Alright. I'm merging these changes for now. Send a PR with your improvements when you're ready.

PR #63, opened 2016-02-29
Tune up sort

lokeshh:
In reference to #39

### DIFF
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -535,24 +535,32 @@ module Daru
     def sort opts={}, &block
       opts = {
         ascending: true,
-        type: :quick_sort
       }.merge(opts)
-
+      
       block = lambda { |a,b|
-        return a <=> b if !(a.nil? || b.nil?)
+        av, ai = a
+        bv, bi = b
+        return av <=> bv if !(av.nil? || bv.nil?)
 
-        if a.nil? && b.nil?
-          0
-        elsif a.nil?
+        if av.nil? && bv.nil?
+          ai <=> bi
+        elsif av.nil?
           -1
         else
           1
         end
-      } unless block
+      } unless block      
 
-      order = opts[:ascending] ? :ascending : :descending
-      vector, index = send(opts[:type], @data.to_a.dup, @index.to_a, order, &block)
-      index = Daru::Index.new index
+      vector_index = @data.each_with_index
+      if block_given?
+        vector_index = vector_index.sort { |a,b| block.call(a[0], b[0]) }
+      else
+        vector_index = vector_index.sort { |a,b| block.call(a, b)}
+      end
+      vector_index.reverse! unless opts[:ascending]
+      vector, index = vector_index.transpose
+      old_index = @index.to_a
+      index = index.map { |i| old_index[i] }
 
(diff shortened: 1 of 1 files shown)


lokeshh:
Earlier:
```ruby
Sort a Vector without any args  0.190000   0.000000   0.190000 (  0.187377)
Sort vector in descending order with custom <=> operator  0.200000   0.000000   0.200000 (  0.204298)
```

Now:
```ruby
Sort a Vector without any args  0.070000   0.010000   0.080000 (  0.074862)
Sort vector in descending order with custom <=> operator  0.120000   0.000000   0.120000 (  0.123979)
```

v0dro:
why aren't tests passing?

lokeshh:
Funny fail
```ruby
       expected: 
       #<Daru::Vector:46315380 @name = nil @size = 6 >
           nil
         2 nil
         4 nil
         5   2
         1   4
         0  22
         3 111
     
            got: 
       #<Daru::Vector:46316220 @name = nil @size = 6 >
           nil
         4 nil
         2 nil
         5   2
         1   4
         0  22
         3 111
```

v0dro:
LOL it's not sorting the nils appropriately.

How about this fix: remove the nil elements, then sort the Array with `Array#sort`, and then prepend nils at the beginning of the vector. The `nil` indexes can be in any order. @gnilrets what do you think?

lokeshh:
@v0dro Are you talking about fixing this particular test or implementing this idea in `vector#sort`?

lib/daru/vector.rb
```diff
@@ -550,8 +549,14 @@ def sort opts={}, &block
...
       } unless block
 
-      order = opts[:ascending] ? :ascending : :descending
-      vector, index = send(opts[:type], @data.to_a.dup, @index.to_a, order, &block)
+      block2 = lambda { |a, b|
+        block.call(a[1], b[1])
+      }
+
+      index_vector = @index.to_a.zip(@data)
+      index_vector = index_vector.sort &block2
```
v0dro: There seems no need to declare `block2` as a separate variable. Simply pass `{ |a, b| block.call(a[1], b[1]) }` to `#sort`.

lib/daru/vector.rb
```diff
@@ -550,8 +549,14 @@ def sort opts={}, &block
...
+      index_vector = index_vector.reverse unless opts[:ascending]
```
v0dro: I think `#reverse!` is faster than `#reverse`.

v0dro:
Hmmm alright, just change the test.

gnilrets:
Another option might be to sort on the index as a secondary sort key.  That should make it consistent over different Ruby versions.  Not sure about any performance impacts.

v0dro:
@lokeshh could you try that out? Different indexes on different versions could prove to be problematic for certain use cases.

lokeshh:
Done.

lokeshh:
I am sorry, I just realized now that I forgot to place `nil`s at the top when order is descending. What to do now? Should I just push the new commit or what?

v0dro:
Create a new PR.

v0dro:
Make sure you mention in the commit message that it's an amendment for this one.

Issue #65, opened 2016-03-01
Vector mode function bug


sunshineyyy:
`require 'daru'`
`foo = Daru::Vector.new [1,2,3,2,4,4,4,4]`
foo.mode gives 2 instead of 4.
This was discovered in this [stackoverflow question](http://stackoverflow.com/questions/35710144/is-ruby-statsample-mode-function-correct)

PR #66, opened 2016-03-01
fix bug in Vector mode method

sunshineyyy:
This change will fix this issue. #65 
Also added Rspec test code to test the new Vector mode method.

### DIFF
--- a/spec/math/statistics/vector_spec.rb
+++ b/spec/math/statistics/vector_spec.rb
@@ -91,7 +91,8 @@ describe Daru::Vector do
 
       context "#mode" do
         it "returns the mode" do
-          @dv.mode
+          mode_test_example = Daru::Vector.new [1,2,3,2,4,4,4,4], dtype: dtype
+          expect(mode_test_example.mode).to eq(4)
         end
       end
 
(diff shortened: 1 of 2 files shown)


spec/math/statistics/vector_spec.rb
```diff
@@ -92,6 +92,9 @@
       context "#mode" do
         it "returns the mode" do
           @dv.mode
+          expect(@dv.mode).to eq(666)
+          mode_test_example = Daru::Vector.new [1,2,3,2,4,4,4,4], dtype: dtype
+          expect(mode_test_example.mode).to eq(4)
         end
```
v0dro: Thank you very much for this fix. Just one concern:

Could you please rewrite this whole test as:
``` ruby
mode_test_example = Daru::Vector.new [1,2,3,2,4,4,4,4], dtype: dtype
expect(mode_test_example.mode).to eq(4)
```

The above 2 lines aren't necessary anymore.

v0dro: Also don't forget to rebase your commits.

sunshineyyy:
Sure. I can totally rewrite this. I am not super familiar with the rebase. Should I make another commit, then squash the two commits and submit another pull request?

v0dro:
Just make another commit, squash them both into a single commit, and then push to your upstream with the `--force` flag. That should reflect it here.

sunshineyyy:
I think it worked. Thanks!

PR #67, opened 2016-03-01
Dataframe#sort speed up

lokeshh:
In reference to #39

### DIFF
--- a/benchmarks/sorting.rb
+++ b/benchmarks/sorting.rb
@@ -28,8 +28,8 @@ Benchmark.bm do |x|
 
   x.report("Sort two columns with custom operators in different orders of DataFrame") do
     df.sort([:c,:a], ascending: [true, false], 
-      by: { c: lambda { |a,b| a.to_s <=> b.to_s },
-            a: lambda { |a,b| (a+1) <=> (b+1) } })
+      by: { c: lambda { |a| a.to_s },
+            a: lambda { |a| a+1 } })
   end
 end
 
@@ -42,3 +42,10 @@ end
 # Sort single column of DataFrame                                          2502.450000 0.000000 2502.450000 (2503.808073)
 # Sort two columns of DataFrame                                            0.540000    0.000000 0.540000    (  0.537670)
 # Sort two columns with custom operators in different orders of DataFrame  2084.160000 7.260000 2091.420000 (2092.716603)
+
+#                                         ===== Current Benchamarks =====
+# Sort a Vector without any args                                           0.070000   0.000000   0.070000 (  0.070323)
+# Sort vector in descending order with custom <=> operator                 0.120000   0.000000   0.120000 (  0.119462)
+# Sort single column of DataFrame                                          0.940000   0.010000   0.950000 (  0.950349)
+# Sort two columns of DataFrame                                            1.490000   0.010000   1.500000 (  1.505680)
+# Sort two columns with custom operators in different orders of DataFrame  1.480000   0.000000   1.480000 (  1.495839)
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1406,8 +1406,8 @@ module Daru
       end
     end
 
-    # Sorts a dataframe (ascending/descending)according to the given sequence of
-    # vectors, using the attributes provided in the blocks.
+    # Sorts a dataframe (ascending/descending) in the given pripority sequence of
+    # vectors, with or without a block.
     #
     # @param order [Array] The order of vector names in which the DataFrame
     #   should be sorted.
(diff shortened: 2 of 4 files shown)


lokeshh:
This uses `Array#sort_by` to replace handmade sort. This has been made possible because `Array#sort_by` can take multiple attributes to sort the array by.

There's one pitfall though. Whereas earlier it used to be like:

```ruby
df.sort [:a, :b], by: {a: lambda { |a,b| a.abs <=> b.abs }, b: lambda { |a,b| a.to_i <=> b.to_i }}
```
Now one is allowed to do it this way only
```ruby
df.sort [:a, :b], by: {a: lambda { |a| a.abs }, b: lambda { |a| a.to_i }}
```

This is because `Array#sort_by` can only accept these sort of blocks.

lokeshh:
Is [this](https://github.com/v0dro/daru/blob/e587a5823a3dc859541dbd377c1b95d1e7bc5117/benchmarks/sorting.rb#L43) some sort of error. How can sorting two columns take so much less time than sorting single column?

v0dro:
Don't think it's a mistake. Did you try running that script on your computer?

v0dro:
This is a good fix that can have far reaching implications in speed and usability, but I would like to confirm if it's alright to change the way the blocks are accessed.

@gnilrets @mohawkjohn @dansbits could use your advice here.

Also, @lokeshh, will it be possible maintain backwards compatibility with the original interface to `:by`? Do research that option too. I think it can be achieved by assuming that the methods users will call on `a` and `b` in the block (`{|a,b| a.method <=> b.method }`) would be the same, so maybe we just call `method` inside your block on the single block, and thereby not break the interface. An error will be raised if the two methods are different.

Also, will this be extensible enough to allow sorting of missing data too?

v0dro:
I might have gone wrong in the benchmarks while copy-pasting and editing or something.

benchmarks/sorting.rb
```diff
@@ -3,7 +3,7 @@
 require 'benchmark'
 require 'daru'
 
-vector = Daru::Vector.new(10000.times.map.to_a.shuffle)
+vector = Daru::Vector.new(1000.times.map.to_a.shuffle)
 df = Daru::DataFrame.new({
```
v0dro: Let the benchmarking code be the same in the final commit. Also, just append your results at the end of the file without deleting anything so we can keep track of speed over time.

lib/daru/dataframe.rb
```diff
@@ -1438,12 +1438,26 @@ def sort! vector_order, opts={}
...
       idx = @index.to_a
-      send(opts[:type], vector_order, idx, opts[:by], opts[:ascending])
+      # send(opts[:type], vector_order, idx, opts[:by], opts[:ascending])
+      idx = @index.to_a.sort_by { |r| lol(r, vector_order, opts) }
       self.index = Daru::Index.new(idx)
+      # Following three lines are very slow 
+      # Master -> 54, 53, 51
+      # This branch (with following lines) -> 12, 13, 5
+      # This branch (without following lines) -> .03, .17, .06
+      for v in self.vectors
```
v0dro: Am I correct in saying that these lines reconstruct the dataframe after the index has been sorted?

How about somehow keeping track of the index of the elements and passing that as a tuple to `sort_by` so they're all sorted together?

lokeshh:
Wow... there's been some very massive increase. I ran the `benchmarks/sorting.rb` for original 10,000 vectors and here's the result:

```ruby
Sort a Vector without any args  0.080000   0.010000   0.090000 (  0.087908)
Sort vector in descending order with custom <=> operator  0.130000   0.000000   0.130000 (  0.182502)
Sort single column of DataFrame  2.190000   0.120000   2.310000 (  3.241833)
Sort two columns of DataFrame  2.390000   0.060000   2.450000 (  3.500302)
Sort two columns with custom operators in different orders of DataFrame  2.400000   0.060000   2.460000 (  3.473831)
```

v0dro:
Beautiful!

gnilrets:
I've avoided sorting dataframes and vectors due to the very poor performance, so anything that makes a dramatic improvement is a great step in the right direction.

v0dro:
@lokeshh you just need to make the tests pass now :)

lokeshh:
I've added support for `nil`'s and it can sort dataframes consisting 10^5 rows in 3 seconds.

lib/daru/dataframe.rb
```diff
@@ -2001,93 +2017,13 @@ def possibly_multi_index? index
...
-    end
-
-    def create_logic_blocks vector_order, by={}
-      universal_block = lambda { |a,b| a <=> b }
-      vector_order.each do |vector|
-        by[vector] ||= universal_block
+    def create_logic_blocks vector_order, by={}, ascending
+      # Display nils at top
+      universal_block_ascending = lambda { |a| a or -Float::INFINITY }
+      universal_block_decending = lambda { |a| a or Float::INFINITY }
```
v0dro: clever :)

gnilrets: Will this work when the vector contains strings and nils or other objects and nils?

v0dro: Good catch. @lokeshh  in the examples, the vectors being sorted are strictly numerical. Could you try with another example that contains objects other than just numbers and nils?

lokeshh: It's failing with non-numeric data. I have a way to solve this but it's not very clean. It would involve creating a new class (see [here](http://stackoverflow.com/a/73496/2623749)).

v0dro: @gnilrets suggested something like this in https://github.com/v0dro/daru/issues/62

I have a concern that creating so many objects would create GC bottlenecks.

lokeshh: No problem. I have got another solution then.

lokeshh: Done. Implemented!

lib/daru/dataframe.rb
```diff
@@ -2001,93 +2028,13 @@ def possibly_multi_index? index
...
-    def create_logic_blocks vector_order, by={}
-      universal_block = lambda { |a,b| a <=> b }
-      vector_order.each do |vector|
-        by[vector] ||= universal_block
+    def create_logic_blocks vector_order, by={}, ascending
+      # Display nils at top
+      universal_block_ascending = lambda { |a| [(a.nil?)? 0:1, a] }
+      universal_block_decending = lambda { |a| [(a.nil?)? 1:0, a] }
+      vector_order.each_with_index do |vector, i|
+        by[vector] ||= ascending[i] ? universal_block_ascending
```
v0dro: Here's a use case which might break this:

Since you're using `||=` to assign blocks to `opts[:by]`, what if a user specifies a block `{a: a.to_s }` for the vector `:a`, and the vector `:a` contains nils? I think that will fail since your algorithm inside the block in the `sort` function relies on the fact that each call of the block produces a tuple of the form `[0 or 1, element ]`.

v0dro:
Tests failures are similar to that in #63

lokeshh:
Should I solve them the same way by sorting through index?

v0dro:
Yep. Consistency is important.

Also check my comment on the code, I have a feeling it's an edge case.

lib/daru/dataframe.rb
lokeshh: It's working fine. I tried.

lokeshh:
Done. I now need to arrange the code better so it looks good and understandable. @gnilrets @v0dro If you find some discrepancies meanwhile please let me know.

lib/daru/vector.rb
```diff
@@ -59,6 +59,8 @@ def map!(&block)
     # Store a hash of labels for values. Supplementary only. Recommend using index
     # for proper usage.
     attr_accessor :labels
+    # Store vector data in an array
+    attr_accessor :data
```
v0dro: Sorry for seeing this so late but this should be `attr_reader`. You don't want to allow users to manipulate the `@data` variable directly.

v0dro:
That edge case I mentioned earlier is not working. I pulled from your branch and tested. Here's the code:

``` ruby
require 'daru'
df = Daru::DataFrame.new({a: ['a', 'aa', nil, 'aaaa'], b: [2,4,5,nil]})
# => 
#<Daru::DataFrame:95207900 @name = 2bf19e45-231d-4468-abd0-4344500e8761 @size = 4>
#                    a          b 
#         0          a          2 
#         1         aa          4 
#         2        nil          5 
#         3       aaaa        nil 
df.sort([:a, :b], by: { b: lambda {|x| x + 4}})
# NoMethodError: undefined method `+' for nil:NilClass
# from (pry):5:in `block in __pry__'
```

lokeshh:
Oh! I see what you meant earlier now. This is not an edge case I think. I
simply haven't introduced the feature of handling nils automatically when
user provided a block. I assumed user would do that when he will provide
the block.

v0dro:
Then please document that in the method docs. This will be a great daru specific feature, I could't see the other dataframe libraries use it.

lokeshh:
What will be a great feature? Are you in favor of keeping the
responsibility of handling nils to the user when he provides the block?

v0dro:
Sorting with the block function even when nils are present.

If handling nils in daru is going to slow down the sorting significantly, then let the user take care of it. But we need to make sure that it's well documented.

lokeshh:
> If handling nils in daru is going to slow down the sorting significantly, then let the user take care of it. But we need to make sure that it's well documented.

Ok, I will document it. I was thinking that we can also have an option like `handle_nils: [true, false]` that user can provide to tell whether he wants the nils to be automatically managed when he passes a block or not. Sometimes the user might want to handle the nils in his own way. For example, he might want them to be displayed in the end. In that case the user can handle the nils himself. What do you say?

lib/daru/dataframe.rb
lokeshh: I am sorry I just saw it now. 

Yes, in the latest implementation only positional index get sorted and then all vectors are reconstructed.

benchmarks/sorting.rb
lokeshh: Ok, will do that once everything is done.

gnilrets:
Can we confirm how much of a slowdown handling nils is going to cost?  Nils are so common in the data I've had to deal with on a daily basis for last 8 years and 3 jobs.  In order to use Daru consistently, I'd probably have to monkey patch Daru or add some other Daru wrapper objects in my projects to handle nils.

I'm not convinced the GC involved with creating Sortable objects every time one has to sort is really going to be an issue.  I'm not sure the best way to test this either.

lokeshh:
For [this](https://github.com/v0dro/daru/blob/master/benchmarks/sorting.rb) benchmark, handling nils take `1.72` seconds and sorting without handling nils take `1.24` seconds.

lokeshh:
> I'd probably have to monkey patch Daru or add some other Daru wrapper objects in my projects to handle nils.

@gnilrets As far as this PR goes, Daru is going to handle nils given no block. @v0dro and me were discussing about whether nils should be handled automatically when user provides a block or not. 

It's not a big deal to handle nils for the user. For example, if user want to handle nils and have a block like `lambda { |a| a.abs }`, he just needs to change this block to `lambda { |a| [(a.nil?)?0:1, a.abs] }`, this will sort with nils at top. On the other hand one can do `lambda { |a| [(a.nil?)?1:0, a.abs] }` if he wants to have nils at the bottom. 

So, I was proposing a solution that user can pass an option to specify whether he wants the nils to be automatically handled **given a block** or not. Nils will be automatically handled **given no block**. What do you say?

gnilrets:
@lokeshh - I misunderstood.  Thanks for the clarification!

v0dro:
That works. It satisfies needs of both novice and advanced users. Having an option to choose if nils should be handled automatically or not would be great. Make sure you document your effort well. Once you're done, you can update the daru notebooks at [sciruby-notebooks](https://github.com/sciruby/sciruby-notebooks) to provide further clarification where necessary.

lokeshh:
Done. Do I need iRuby to run the sciruby-notebooks?

v0dro:
Yes.

* Clone the repo and modify according to what you need.
* Run the notebook with a fresh iRuby kernel and make sure the whole notebook works without any errors.
* Push changes.

lokeshh:
I encountered a problem while modifying one of the notebooks. It's [here](https://github.com/SciRuby/iruby/issues/73)

PR #70, opened 2016-03-04
Tune up sort [amendment]

lokeshh:
Amendment to #63 

- Places nils at top when descending
- Add relevant tests to check nils ordering with numeric and non-numeric data

### DIFF
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -502,11 +502,32 @@ describe Daru::Vector do
             expect(sorted).to eq(Daru::Vector.new(["My", "Jazz", "Guitar", "My Jazz Guitar"], index: [2,1,3,0]))
           end
 
-          it "places nils near the beginning of the vector" do
+          it "places nils near the beginning of the vector when sorting ascendingly" do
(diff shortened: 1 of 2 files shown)


spec/vector_spec.rb
```diff
@@ -502,11 +502,32 @@
             expect(sorted).to eq(Daru::Vector.new(["My", "Jazz", "Guitar", "My Jazz Guitar"], index: [2,1,3,0]))
           end
 
-          it "places nils near the beginning of the vector" do
+          it "places nils near the beginning of the vector when sorting ascendingly" do
             with_nils = Daru::Vector.new [22,4,nil,111,nil,2]
 
             expect(with_nils.sort).to eq(Daru::Vector.new([nil,nil,2,4,22,111], index: [2,4,5,1,0,3]))
           end if dtype == :array
+
```
lokeshh: Should I use `if dtype == :array` in below tests like used here?

v0dro: I had placed that there because NMatrix 0.1.0 would segfault with nils sometimes. I guess that does not happen anymore with NMatrix 0.2.1. So no need.

PR #76, opened 2016-03-12
Massive performance enhancement for joins.

gnilrets:
This commit replaces the previous cartesian and bloomfilter
algorithms for joining with a sorted merge method.  The
result is that joins are now O(n) instead of O(n^2).

Check benchmarks/joining.rb for details on performance enhancement.


gnilrets:
I had to change some of the tests to be ordered by the join keys (a consequence of using a sorted merge algorithm), but they have not been changed in any substantial way.

v0dro:
Great!

PR #82, opened 2016-03-27
added new statistics

shahsaurabh0605:
I have added variance, range, median and mode statistical methods for dataframes and completed the remaining TODO specs.

### DIFF
--- a/lib/daru/maths/statistics/dataframe.rb
+++ b/lib/daru/maths/statistics/dataframe.rb
@@ -2,24 +2,30 @@ module Daru
   module Maths
     module Statistics
       module DataFrame
-        # Calculate mean of numeric vectors.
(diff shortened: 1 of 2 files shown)


lib/daru/maths/statistics/dataframe.rb
```diff
@@ -6,6 +6,27 @@ module DataFrame
         def mean
           compute_stats :mean
         end
+        
+        # Calculate variance of numeric vectors
+        def variance_sample
+          compute_stats :variance_sample
```
v0dro: All these methods can be replaced with `define_method`. You can add documentation with the YARD `@method` syntax.

shahsaurabh0605:
All the methods are now in `define_method`.

Issue #49, opened 2016-02-11
Sorting a DataFrame and then creating a pivot_table for it produces wrong aggregate values


sebastianiorga:
``` ruby
$> df = Daru::DataFrame.new({
    a: [1,2,3,4,5,6]*100,  
    b: ['a','b','c','d','e','f']*100,  
    c: [11,22,33,44,55,66]*100,  
    d: ['r']*600  
})    
=> 
#<Daru::DataFrame:54383340 @name = d132569b-0e1e-4516-ae37-fc41c07bde58 @size = 600>
                    a          b          c          d 
         0          1          a         11          r 
         1          2          b         22          r 
         2          3          c         33          r 
         3          4          d         44          r 
         4          5          e         55          r 
         5          6          f         66          r 
         6          1          a         11          r 
         7          2          b         22          r 
         8          3          c         33          r 
         9          4          d         44          r 
        10          5          e         55          r 
        11          6          f         66          r 
        12          1          a         11          r 
        13          2          b         22          r 
        14          3          c         33          r 
       ...        ...        ...        ...        ... 

$> a = df.dup.sort([:b]).dup.pivot_table(index: [:b, :d], values: [:a, :c], agg: :sum)
=> 
#<Daru::DataFrame:56265240 @name = f7fd6275-b6a4-4ad9-b456-7f2aad0fc3ac @size = 6>
                    a          c 
["a", "r"]        346       3806 
["b", "r"]        350       3850 
["c", "r"]        354       3894 
["d", "r"]        346       3806 
["e", "r"]        350       3850 
["f", "r"]        354       3894 

$> b = df.dup.pivot_table(index: [:b, :d], values: [:a, :c], agg: :sum)
=> 
#<Daru::DataFrame:57262440 @name = 49eb6d38-ed2f-42f0-a764-d026d94eb18c @size = 6>
                    a          c 
["a", "r"]        100       1100 
["b", "r"]        200       2200 
["c", "r"]        300       3300 
["d", "r"]        400       4400 
["e", "r"]        500       5500 
["f", "r"]        600       6600 

```

I'm still chewing on what the cause is. I think the problem is related to the difference between how GroupBy#get_group and GroupBy#select_groups_from work. GroupBy#select_groups_from produces incorrect dataframes that look like what pivot_tale is making use of. #get_group produces correct results.

The code below is while inside a pry session triggered by a binding.pry inside GroupBy#apply_method.

``` ruby
$> self
[...]

PR #85, opened 2016-03-31
bug in pivot_table solved

shahsaurabh0605:
This PR solves the issue [#49](https://github.com/v0dro/daru/issues/49). Describing briefly : 

Our `pivot_table` requires the index of dataframe to be in the form 0 -> numRows.  If you sort the dataframe then the index keys won't have values 0 -> numRows and things break down. Any dataframe with numerical index other than of the form 0 -> numRows will create error in `pivot_table`.

While `pivot_table` does not depend upon the dataframe index,  it would be best to first assign any dataframe index to 0 -> numRows and then carry out the regular routine for `pivot_table`.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1331,7 +1331,7 @@ module Daru
     #
     # @param [Daru::Index] idx New index object on which the rows of the dataframe
     #   are to be indexed.
-    # @example Reassgining index of a DataFrame
+    # @example Reassigning index of a DataFrame
     #   df = Daru::DataFrame.new({a: [1,2,3,4], b: [11,22,33,44]})
     #   df.index.to_a #=> [0,1,2,3]
     #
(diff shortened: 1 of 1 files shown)


lib/daru/dataframe.rb
```diff
@@ -1578,6 +1578,10 @@ def pivot_table opts={}
       raise ArgumentError,
         "Specify grouping index" if !opts[:index] or opts[:index].empty?
 
+      arr = []
+      self.size.times {|i| arr[i] = i}
+      new_dataframe_index = Daru::Index.new(arr)
+      self.index = new_dataframe_index
```
v0dro: You're changing the DataFrame interally by re-assigning `self.index`. This is unexpected behaviour for `pivot_table`.

shahsaurabh0605: This seems to be kind of unexpected behaviour, but I don't think this would change any of the results. This will infact solve the bug.

v0dro: Of course it will change the results. It will alter the dataframe. That's a change of result.

v0dro:
Closed for not solving the problem without breaking something else. In this case modifying the dataframe on which the method would be called.

PR #88, opened 2016-04-01
Fix doc recommending against using []

gnilrets:
The comments on the #vector method say that it has been deprecated, but the comments
on #[] say to use #vector.  I think #[] is actually preferred.


v0dro:
you're right. thanks.

PR #89, opened 2016-04-02
added percent_change function

shahsaurabh0605:
I have added percent_change function in vector.rb. Have a look and suggest me of any changes to be done.

### DIFF
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -377,6 +377,40 @@ module Daru
           end
         end
 
+        # The percent_change method computes the percent change over
+        # the given number of periods.
+        #
+        # @param [Integer] periods (1) number of nils to insert at the beginning.
+        #
+        # @example
+        # 
+        #   vector = Daru::Vector.new([4,6,6,8,10],index: ['a','f','t','i','k'])
+        #   vector.percent_change
+        #   #=>
+        #   #   <Daru::Vector:28713060 @name = nil @size: 5 >
+        #   #              nil
+        #   #   a	
+        #   #   f	   0.5
+        #   #   t	   0.0
+        #   #   i	   0.3333333333333333
+        #   #   k          0.25
+        def percent_change periods = 1
+          type == :numeric or raise TypeError, "Vector must be numeric"
+          value = self.only_valid
+          arr = []
+          i = 1 
+          ind = @data.find_index{|x|!x.nil?}
+          (periods...size).each do |j|
+            if j==ind || @missing_values.has_key?(@data[j])
+              arr[j] = nil
+            else
+              arr[j] = (value.data[i] - value.data[i - 1]) / value.data[i - 1].to_f
+              i+=1
+            end 
+          end
+          Daru::Vector.new(arr, index: @index, name: @name)
+        end
+
         # Performs the difference of the series.
         # Note: The first difference of series is X(t) - X(t-1)
         # But, second difference of series is NOT X(t) - X(t-2)
--- a/spec/math/statistics/vector_spec.rb
+++ b/spec/math/statistics/vector_spec.rb
@@ -383,6 +383,20 @@ describe Daru::Vector do
       expect(acf[3]).to be_within(0.001) .of(0.486)
     end
   end
+  
+  context "#percent_change" do
+    it "calculates percent change" do
+      vector = Daru::Vector.new([4,6,6,8,10],index: ['a','f','t','i','k'])
+      expect(vector.percent_change).to eq(
+      Daru::Vector.new([nil, 0.5, 0.0, 0.3333333333333333, 0.25], index: ['a','f','t','i','k']))
+    end
+
+    it "tests for numerical vectors with nils" do
+      vector2 = Daru::Vector.new([nil,6,nil,8,10],index: ['a','f','t','i','k'])
+      expect(vector2.percent_change).to eq(
+      Daru::Vector.new([nil, nil, nil, 0.3333333333333333, 0.25], index: ['a','f','t','i','k']))
+    end
+  end
 
   context "#diff" do
     it "performs the difference of the series" do


lib/daru/vector.rb
```diff
@@ -576,6 +576,38 @@ def sorted_data &block
       @data.to_a.sort(&block)
     end
 
+    # Create a new vector by specifying the periods of nils.
+    #
+    # == Description
```
v0dro: Follow YARD documentation syntax.

v0dro:
Could you add a test case?

shahsaurabh0605:
Added the specs and rewritten documentation in YARD syntax.

v0dro:
I didn't see this before, but this is a stats function and belongs in this file: https://github.com/v0dro/daru/blob/master/lib/daru/maths/statistics/vector.rb

shahsaurabh0605:
Sorry, I missed that. Added the function in appropriate place.

lib/daru/maths/statistics/vector.rb
```diff
@@ -377,6 +377,36 @@ def sample_without_replacement(sample=1)
...
+        #   #   a	
+        #   #   b	   -0.3333333333333333
+        #   #   c	   -0.5
+        #   #   f	   -0.5
+        def percent_change periods = 1
+          value = []
+          arr = []
+          new_index = [] 
+          self.index.each do |e|
+            if self[e].is_a?(Numeric)
```
v0dro: You can replace this loop with `self.type == :numeric`. When the type is numeric it indicates that all the members are numerics or missing. You can then call only_valid to get only the numerical values.

v0dro: And btw, percent change for any data apart from numeric should not be supported.

v0dro: So if a string is encountered, it's supposed to raise an error.

shahsaurabh0605:
Now, the percent change is applicable to only numeric vectors.

lib/daru/maths/statistics/vector.rb
```diff
@@ -377,6 +377,35 @@ def sample_without_replacement(sample=1)
...
+        #   #=>
+        #   #   <Daru::Vector:28713060 @name = nil @size: 5 >
+        #   #              nil
+        #   #   a	
+        #   #   f	   0.5
+        #   #   t	   0.0
+        #   #   i	   0.3333333333333333
+        #   #   k          0.25
+        def percent_change periods = 1
+          value = []
```
v0dro: remove this line

lib/daru/maths/statistics/vector.rb
```diff
@@ -377,6 +377,35 @@ def sample_without_replacement(sample=1)
...
+          arr = []
```
v0dro: put this inside the `if`.

spec/math/statistics/vector_spec.rb
```diff
@@ -383,6 +383,14 @@
       expect(acf[3]).to be_within(0.001) .of(0.486)
     end
   end
+  
+  context "#percent_change" do
+    it "calculates percent change" do
+      vector = Daru::Vector.new([4,6,6,8,10],index: ['a','f','t','i','k'])
```
v0dro: What if there's missing data in the vector? Vectors with all numeric data and missing data still qualify as numeric vectors.

shahsaurabh0605:
I have made the changes such that `percent_change` now handles only numeric vectors alongwith missing data.

lib/daru/maths/statistics/vector.rb
```diff
@@ -377,6 +377,39 @@ def sample_without_replacement(sample=1)
...
+        #   #   i	   0.3333333333333333
+        #   #   k          0.25
+        def percent_change periods = 1
+          type == :numeric or raise TypeError, "Vector must be numeric"
+          arr = []
+          i = periods
+          (periods...size).each do |j|
+            if @missing_values.has_key?(@data[i])
+              arr[j] = nil
+              @data.delete_at(i)
```
v0dro: Didn't you notice that `delete_at` changes the data of the vector on which percent_change is being called?

v0dro: You know you really should think about all the side effects your code can have before you commit.

v0dro: Why not simply call `Vector#only_valid` and work on the vector that is returned by it?

shahsaurabh0605:
I have made the changes such that `Vector#only_valid` is used. I will try to be more careful of the side effects from now on.

spec/math/statistics/vector_spec.rb
v0dro: Could you add a test case for missing data?

v0dro: Here's a case:
``` ruby

 v=Daru::Vector.new([nil,6,nil,8,10],index: ['a','f','t','i','k'])
```

lib/daru/maths/statistics/vector.rb
```diff
@@ -377,6 +377,40 @@ def sample_without_replacement(sample=1)
...
+        #   #   <Daru::Vector:28713060 @name = nil @size: 5 >
+        #   #              nil
+        #   #   a	
+        #   #   f	   0.5
+        #   #   t	   0.0
+        #   #   i	   0.3333333333333333
+        #   #   k          0.25
+        def percent_change periods = 1
+          type == :numeric or raise TypeError, "Vector must be numeric"
+          value = self.only_valid
```
v0dro: Now that I think of it, you can get away with working with `@data` only.

Here's a sample:
``` ruby

def percent_change periods = 1
   type == :numeric or raise TypeError, "Vector must be numeric"
   arr = []
   i = 1 
   ind = @data.find_index{|x|!x.nil?}
   (periods...size).each do |j|
     if j==ind || @missing_values.has_key?(@data[j])
       arr[j] = nil
     else
       arr[j] = (@data[i] - @data[i - 1]) / @data[i - 1].to_f
       i+=1
     end 
   end
   Daru::Vector.new(arr, index: @index, name: @name)
 end
```

shahsaurabh0605: This code won't work for test cases like this : 
`vector = Daru::Vector.new([2,4,5,nil,6,7,nil])`
Actually, counter `i` must run through only numbers excluding the nils. So working with `only_valid` will be the best in this case. I initially tried to do this with deleting nils from `@data` but that is obviously not the way out.

v0dro: Well then can you please make it work? Refer the pandas `pct_change` function.

shahsaurabh0605: Is there any problem with using `only_valid`? I think this should be the best to serve our purpose here.

v0dro: I don't have a problem. But will it work with missing data?

shahsaurabh0605: I tested it for all the test cases including missing data. According to me it works!

v0dro: OK then can you add those tests to the spec?

shahsaurabh0605:
Added the tests.

PR #92, opened 2016-04-07
[WIP] Order columns as given in CSV 

### LINKED ISSUE #91: Daru::DataFrame.from_csv change original order of the columns
```ruby
# Create csv with headers unordered
# Tipical csv example from other source
$> aux = Daru::DataFrame.new(
  {
    'XBeer' => ['Kingfisher', 'Snow', 'Bud Light', 'Tiger Beer', 'Budweiser'],
    'Gallons sold' => [500, 400, 450, 200, 250]
  },
  index: ['India', 'China', 'USA', 'Malaysia', 'Canada']
)
$> aux.vectors = Daru::Index.new(["Gallons sold", "Beer"])
# Now headers are unordered
$> aux
=>
#<Daru::DataFrame:40144608 @name = f6ec2382-11d8-4446-8325-802d1cdca032 @size = 5>
           Gallons so      Beer
     India        500 Kingfisher
     China        400       Snow
       USA        450  Bud Light
  Malaysia        200 Tiger Beer
    Canada        250  Budweiser

$> aux.write_csv('df_original_order.csv')
##############################################################
# Now with this file we can reproduce the problem
# Open the csv from other source
$> df = Daru::DataFrame.from_csv('df_original_order.csv')
# Doing some computation or concat new data
# ...
# Now saving the results (headers are ordered)
$> df.write_csv('df_other_order.csv')
```
why i can't keep the original order?
How can i read csv without change the order of the columns? or at least write the csv in arbitrary order?

lokeshh:
Fixes #91

### DIFF
--- a/lib/daru/io/io.rb
+++ b/lib/daru/io/io.rb
@@ -86,6 +86,9 @@ module Daru
           headers.each_with_index do |h, i|
             hsh[h] = csv_as_arrays[i]
           end
+
+          # Order columns as given in CSV
+          daru_options[:order] = headers.to_a
         else
           opts[:header_converters] ||= :symbol
 
--- a/spec/io/io_spec.rb
+++ b/spec/io/io_spec.rb
@@ -15,7 +15,7 @@ describe Daru::IO do
 
       it "works properly for repeated headers" do
         df = Daru::DataFrame.from_csv('spec/fixtures/repeated_fields.csv',header_converters: :symbol)
-        expect(df.vectors.to_a).to eq(['a1', 'age_1', 'age_2', 'city', 'id', 'name_1', 'name_2'])
+        expect(df.vectors.to_a).to eq(["id", "name_1", "age_1", "city", "a1", "name_2", "age_2"])
 
         age = Daru::Vector.new([3, 4, 5, 6, nil, 8])
         expect(df['age_2']).to eq(age)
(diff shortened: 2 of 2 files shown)


lokeshh:
@v0dro What do `:headers` option and `:header_converters` do?

v0dro:
They are Ruby CSV options. Check the Ruby docs.

lokeshh:
This doesn't work when `headers: true`.

v0dro:
`headers` is supposed to be an Array that sets the CSV reader to treat the first line of the CSV file as data and use the Array supplied in the option as the header for the DataFrame.

``` ruby

2.2.1 :013 > q=CSV.read('a.csv', headers: ['a', 'b','c','d','e'])
 => #<CSV::Table mode:col_or_row row_count:6> 
2.2.1 :014 > q.headers
 => ["a", "b", "c", "d", "e"] 
2.2.1 :015 > q[0]
 => #<CSV::Row "a":"Year" "b":"Make" "c":"Model" "d":"Description" "e":"Price"> 
```

The file looks like this:

```
Year,Make,Model,Description,Price
1997,Ford,E350,"ac, abs, moon",3000.00
1999,Chevy,"Venture ""Extended Edition""","",4900.00
1999,Chevy,"Venture ""Extended Edition, Very Large""",,5000.00
1996,Jeep,Grand Cherokee,"MUST SELL!
air, moon roof, loaded",4799.00

```

lokeshh:
I saw `headers: true ` in the first spec in `io_spec.rb`.

v0dro:
That's used if you want to use the first line as the header and return a `CSV::Table` object.

See the docs here: http://ruby-doc.org/stdlib-2.0.0/libdoc/csv/rdoc/CSV.html#method-c-new

PR #95, opened 2016-04-13
added exponential moving variance function

shahsaurabh0605:
I have added the exponential moving variance function `emv` for the vectors. Also, why is the example in the documentation of `ema` not in the form of vector? Does that need to be changed?


v0dro:
Yes it can be changed. Good code.

PR #96, opened 2016-04-14
Modify apply_method to handle nils

### LINKED ISSUE #44: Daru::Core::GroupBy#apply_method sometimes chokes on nils
Using `daru` from the master branch on github.
Example:

```ruby
$> d = Daru::DataFrame.new({ key: [:a, :b, :c, :b], val: [1, nil, nil, 2] })
=> 
#<Daru::DataFrame:42220140 @name = 2ff8be51-8af2-4a10-8ea9-248409523960 @size = 4>
                  key        val 
         0          a          1 
         1          b        nil 
         2          c        nil 
         3          b          2 

$> d.pivot_table(index: [:key])
NoMethodError: undefined method `mean' for nil:NilClass
from /home/seb/.rvm/gems/ruby-2.2.1/bundler/gems/daru-99db9f85ab6e/lib/daru/core/group_by.rb:228:in `block (2 levels) in apply_method'
[6] pry(#<#<Class:0x007fd28039b110>>)> 
```

Things that combine to cause the problem:
1. `Vector#[]` returns a value when it receives a single integer index
2. `GroupBy#apply_method`, for this example, `@groups` is set to `{[:a]=>[0], [:b]=>[1, 3], [:c]=>[2]}
` so the call below to `slice = vec[*indexes]` just passes one value.
3. The next line, `daru/core/group_by.rb @ line 228`: `single_row << (slice.is_a?(Numeric) ? slice : slice.send(method))` presumably assumes that if slice is not a numeric it must be an Enumerable? Regardless it explodes when it gets a single nil.

I'm not sure what the correct fix is. Some ideas: 
1. Vector#[] could always return a Vector. That's a consistency plus, but probably a usability hit since extracting the value easily is nice.
2. Guard against `nil` in `GroupBy#apply_method` and just append `nil` to `single_row` if `slice.nil?`. Or something along those lines?
3. It seems like the existing check, at `daru/core/group_by.rb @ line 228` is essentially a type check to see if the slice is an `Enumerable` or not. Maybe just have it be `single_row << (slice.is_a?(Enumerable) ? slice.send(method)) : slice`. 

Just getting started with Daru so not sure if `Vector#[]` returning a bare `nil` causes issues elsewhere. I've added fix 3. to my fork and will see if it ends up breaking other things as I do more stuff based on pivot tables. The tests are green for what that matters.

lokeshh:
Fixes #44 and update min, max, sum and mean to work with missing values.

### DIFF
--- a/lib/daru/accessors/array_wrapper.rb
+++ b/lib/daru/accessors/array_wrapper.rb
@@ -66,27 +66,31 @@ module Daru
         ArrayWrapper.new @data.dup, @context
       end
 
+      def compact
+        @data - @context.missing_values
+      end
+
       def mean
-        sum.quo(@size - @context.missing_positions.size).to_f
+        values_to_sum = compact
+        return nil if values_to_sum.empty?
+        sum = values_to_sum.inject :+
+        sum.quo(values_to_sum.size).to_f
       end
 
       def product
-        @data.inject(1) { |m,e| m*e unless e.nil? }
+        compact.inject :*
       end
 
       def max
-        @data.max
+        compact.max
       end
 
       def min
-        @data.min
+        compact.min
       end
 
       def sum
-        @data.inject(0) do |memo ,e|
-          memo += e unless e.nil? #TODO: Remove this conditional somehow!
-          memo
-        end
+        compact.inject :+
       end
 
      private
--- a/spec/math/statistics/vector_spec.rb
+++ b/spec/math/statistics/vector_spec.rb
@@ -6,12 +6,15 @@ describe Daru::Vector do
       before do
         @dv = Daru::Vector.new [323, 11, 555, 666, 234, 21, 666, 343, 1, 2], dtype: dtype
         @dv_with_nils = Daru::Vector.new [323, 11, 555, nil, 666, 234, 21, 666, 343, nil, 1, 2]
+        @dv_with_missing = Daru::Vector.new [1, 2, 3, 3], missing_values: [3], dtype: dtype
+        @dv_with_all_missing = Daru::Vector.new [3, 3], missing_values: [3], dtype: dtype
       end
 
       context "#mean" do
         it "calculates mean" do
           expect(@dv.mean).to eq(282.2)
-          expect(@dv_with_nils.mean).to eq(282.2)
+          expect(@dv_with_missing.mean).to eq(1.5)
+          expect(@dv_with_all_missing.mean).to eq(nil)
         end
       end
 
@@ -60,26 +63,58 @@ describe Daru::Vector do
 
       context "#max" do
         it "returns the max value" do
-          @dv.max
+          expect(@dv.max).to eq(666)
         end
+        
+        it "returns the max value without considering values set as missing" do
+          expect(@dv_with_missing.max).to eq(2)
+        end
+        
+        it "returns nil when all values are set missing" do
+          expect(@dv_with_all_missing.max).to eq(nil)
+        end          
       end
 
       context "#min" do
         it "returns the min value" do
-          @dv.min
+          expect(@dv.min).to eq(1)
         end
+        
+        it "returns the min value without considering values set as missing" do
+          expect(@dv_with_missing.min).to eq(1)
+        end
+        
+        it "returns nil when all values are set missing" do
+          expect(@dv_with_all_missing.min).to eq(nil)
+        end          
       end 
 
       context "#sum" do
         it "returns the sum" do
-          @dv.sum
+          expect(@dv.sum).to eq(2822)
         end
+        
+        it "returns the sum without considering values set as missing" do
+          expect(@dv_with_missing.sum).to eq(3)
+        end
+        
+        it "returns nil when all values are set missing" do
+          expect(@dv_with_all_missing.sum).to eq(nil)
+        end        
       end
 
       context "#product" do
         it "returns the product" do
           v = Daru::Vector.new [1, 2, 3, 4, 5], dtype: dtype
           expect(v.product).to eq(120)
+        end
+
+        it "returns the product without considering values set as missing" do
+          expect(@dv_with_missing.product).to eq(2)
+        end
+        
+        it "returns nil when all values are set missing" do
+          expect(@dv_with_all_missing.product).to eq(nil)
         end
       end
 
--- a/benchmarks/group_by.rb
+++ b/benchmarks/group_by.rb
@@ -7,11 +7,11 @@ data = Daru::DataFrame.from_csv 'TradeoffData.csv'
 
 Benchmark.bm do |x|
   x.report("Single column grouping") do
-    @single = data.group_by([:Treatment])
+    @single = data.group_by(['Treatment'])
   end
 
   x.report("Multi-column grouping") do
-    @multi = data.group_by([:Group, :Treatment])
+    @multi = data.group_by(['Group', 'Treatment'])
   end
 
   x.report("Single mean") do
@@ -24,9 +24,9 @@ Benchmark.bm do |x|
 end
 
 #                    ===== Benchmarks =====
-# 
+#
 #                          user     system      total        real
-# Single column grouping 0.000000   0.000000   0.000000   (0.000356)
-# Multi-column grouping  0.000000   0.000000   0.000000   (0.000958)
-# Single mean            0.000000   0.000000   0.000000   (0.000865)
-# Multi mean             0.000000   0.000000   0.000000   (0.002748)
+# Single column grouping  0.000000   0.000000   0.000000  (0.000340)
+# Multi-column grouping   0.000000   0.000000   0.000000  (0.000855)
+# Single mean             0.000000   0.000000   0.000000  (0.001208)
+# Multi mean              0.000000   0.000000   0.000000  (0.004892)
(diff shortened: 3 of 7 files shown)


lokeshh:
There's another bug to be solved:

```ruby
[7] pry(main)> df
=> 
#<Daru::DataFrame:21298420 @name = 5763c93c-46bb-4707-975e-ec0be84cec96 @size = 10>
                    a          b          c          d          e 
         0        foo        one      small          1          2 
         1        foo        one      large          2          4 
         2        foo        one      large          2          4 
         3        foo        two      small          3          6 
         4        foo        two      small          3          6 
         5        bar        one      large          4          8 
         6        bar        one      small          5         10 
         7        bar        two      large          6         12 
         8        bar        two      small          7         14 
         9        ice        one      small          8        nil 

[8] pry(main)> df.pivot_table(index: [:a], vectors: [:b], values: :e)
ZeroDivisionError: divided by 0
from /home/ubuntu/workspace/daru/lib/daru/accessors/array_wrapper.rb:70:in `/'
```

lokeshh:
Perhaps we should return nil instead of `ZeroDivisionError`.

v0dro:
Good idea. `mean` is returning a divide by zero error because it does not consider the number of missing elements when computing the mean.

Go ahead and change it.

v0dro:
Could you add a test that makes sure that the issue this PR fixes has been actually fixed? This will also serve for avoiding potential pitfalls for future contributors.

lokeshh:
Done. 

Maybe we should also have `min` and `max` methods working for vectors having nils.

v0dro:
Yes we should.

lib/daru/maths/statistics/vector.rb
```diff
@@ -7,6 +7,8 @@ module Statistics
       module Vector
         def mean
           @data.mean
+          rescue ZeroDivisionError
```
v0dro: A more robust way of handling this is to change the wrapper methods that handle basic stats operations like mean and product. See this: https://github.com/v0dro/daru/blob/master/lib/daru/accessors/array_wrapper.rb#L70

You can make it return a nil if the size of `@context.missing_positions` is `0`.

lokeshh: What is there on [line 87](https://github.com/v0dro/daru/blob/master/lib/daru/accessors/array_wrapper.rb#L87). Why should the conditional be removed?

v0dro: I wanted a method to calculate sum without have to check if the element is nil every time so that the method is not slowed down as a result of the conditional.

v0dro: An btw that looks like a bug. `sum` is supposed to skip over all missing elements in the `missing_values`.

lokeshh: Yes, it is. Have a look at changes I have made in the latest commit.

lokeshh:
I am about to add tests. I noticed no expect keyword in some of the [specs](https://github.com/v0dro/daru/blob/master/spec/math/statistics/vector_spec.rb#L57). Is that intentional?

v0dro:
Those specs are incomplete. I had added a method call there just to make sure that I've not made any syntax errors.

lib/daru/accessors/array_wrapper.rb
```diff
@@ -67,26 +67,24 @@ def dup
       end
 
       def mean
-        sum.quo(@size - @context.missing_positions.size).to_f
+        return nil if @data.compact.empty?
```
v0dro: You'll call `@data.compact` twice in case its not empty, which will be expensive. To top it all your code does not consider the fact that elements apart from `nil` can also be marked as missing.

v0dro: mean can be written like this:

``` ruby
def mean
  non_missing_size = @size - @context.missing_positions.size
  return nil if non_missing_size == 0
  sum.quo(non_missing_size).to_f
end

File: lib/daru/accessors/array_wrapper.rb [hunk 3]
```diff
@@ -67,26 +67,24 @@ def dup
...
       end
 
       def product
-        @data.inject(1) { |m,e| m*e unless e.nil? }
+        @data.compact.inject :*
       end
 
       def max
-        @data.max
+        @data.compact.max
```
v0dro: Again, the same issue as above. Factor missing elements apart from nil.

lib/daru/accessors/array_wrapper.rb
lokeshh: How about creating a custom function like compact to get rid of all missing values?

v0dro: There is `#only_valid` which does that, but it creates a new Vector and Index, which might be too expensive.

v0dro: You can add a `#compact` inside the wrappers though. Make sure you have an implementation for each wrapper.

lokeshh: `#only_valid` with `as_a: :array` returns an array without creating a vector. So, I can use `@context.only_valid(as_a: :array)` instead of `@data.compact`.

lokeshh:
I have created `compact` in array_wrapper. Does it look good? If it's alright then I will go write the same for other wrappers.

lib/daru/accessors/array_wrapper.rb
```diff
@@ -66,27 +66,32 @@ def dup
         ArrayWrapper.new @data.dup, @context
       end
 
+      def compact
+        new_index = (@context.index.to_a - @context.missing_positions)
+        new_index.map { |i| @context[i] }
```
v0dro: Shouldn't `#compact` be returning a data array with the missing elements removed?

How about `@data - @context.missing_values`?

lib/daru/accessors/array_wrapper.rb
```diff
@@ -66,27 +66,32 @@ def dup
...
+      end
+
       def mean
-        sum.quo(@size - @context.missing_positions.size).to_f
+        non_nils = compact
```
v0dro: `non_nils` is misleading. Missing need not be nil.

lib/daru/accessors/array_wrapper.rb
lokeshh: Fixed.

lib/daru/accessors/array_wrapper.rb
lokeshh: How about `elements_to_sum`?

lokeshh:
I was implementing the functions in gsl wrapper and found out that I couldn't create a vector with dtype gsl having nil values. Is it normal? This will lead to problem if we use pivot_table on a data frame having vectors of dtype gsl.
```ruby
[6] pry(main)> @dv_with_missing = Daru::Vector.new [1, 2, 3, 3, nil], missing_values: [3], dtype: :gsl                                                              
TypeError: no implicit conversion to float from nil
from /home/ubuntu/workspace/daru/lib/daru/accessors/gsl_wrapper.rb:86:in `alloc'
```

v0dro:
So `@data` is a `GSL::Vector` and it does not support the `#-` method with an `Array`. Same is the case with not supporting nils. User cannot use nils if he's using GSL as the internal storage type.

spec/math/statistics/vector_spec.rb
```diff
@@ -6,12 +6,15 @@
       before do
         @dv = Daru::Vector.new [323, 11, 555, 666, 234, 21, 666, 343, 1, 2], dtype: dtype
         @dv_with_nils = Daru::Vector.new [323, 11, 555, nil, 666, 234, 21, 666, 343, nil, 1, 2]
+        @dv_with_missing = Daru::Vector.new [1, 2, 3, 3], missing_values: [3], dtype: :gsl
+        @dv_with_all_missing = Daru::Vector.new [3, 3], missing_values: [3], dtype: :gsl
       end
```
lokeshh: Had to replace all `nil`s so that the same tests can work with `:gsl`. What do you suggest? Should I test the array wrapper separately to include `nil`?

lokeshh:
I updated the functions for gsl wrapper. I think there are also many other functions like median, mode which could also be updated. What do you say?

Also, I think I should split this PR into 2 PR's
- one to fix #44 
- another to update functions min, max, etc.

v0dro:
I'm concerned about the space overhead that we'll have per object if `@valid_data` is used too. So let it be for now. GC is a big concern when dealing with a lot of objects and it quickly becomes a bottleneck.

Also, could you try benchmarking with `#compact` versus writing a loop inside methods like `mean` and individually checking whether the value is `nil` or not to compute that particular stat?

spec/math/statistics/vector_spec.rb
v0dro: This is also fine.

v0dro: Btw it should be `dtype: dtype`

lokeshh:
Here are benchmarks:
```ruby
Sum of a vector using compact  0.000000   0.000000   0.000000 (  0.001734)
Sum of a vector without compact  0.010000   0.010000   0.020000 (  0.018389)
```
[Here's](https://gist.github.com/lokeshh/73ca4d3efa30a0e41a126f881bc28f9f) the code for benchmark and sum functions.

v0dro:
Alright looks good. Just add those benchmarks to the codebase and I'll merge.

v0dro:
Note: Modify benchmarks to reflect the real statistics methods being used in daru.

PR #99, opened 2016-04-15
added exponential moving standard deviation function

shahsaurabh0605:
Added the exponentially moving standard deviation function `emsd` for vectors.

### DIFF
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -565,6 +565,34 @@ module Daru
           Daru::Vector.new(var_base, index: @index, name: @name)	
         end
 
+        # Exponential Moving Standard Deviation.
+        # Calculates an exponential moving standard deviation of the series using a
+        # specified parameter. If wilder is false (the default) then the EMSD
+        # uses a smoothing value of 2 / (n + 1), if it is true then it uses the
+        # Welles Wilder smoother of 1 / n.
+        #
+        # @param [Integer] n (10) Loopback length.
+        # @param [TrueClass, FalseClass] wilder (false) If true, 1/n value is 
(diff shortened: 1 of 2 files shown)


lib/daru/maths/statistics/vector.rb
```diff
@@ -565,6 +565,34 @@ def emv(n = 10, wilder = false)
...
+        #
+        #   # first 9 observations are nil
+        #   ts.emsd   # => [ ... nil, 0.285... , 0.258..., 0.243..., ...]
+        #
+        # @return [Daru::Vector] contains EMSD
+        def emsd(n = 10, wilder = false)
+          result = []
+          vec = emv(n, wilder)
+          vec.data.each do |d|
+            result << (@missing_values.has_key?(d) ? nil : Math::sqrt(d))
```
v0dro: You have a vector in `vec` and you're using the `@missing_values` parameter of `self`. Use missing values of `vec` directly.

Also, `vec` is a very random name. Call it something specific.

lib/daru/maths/statistics/vector.rb
```diff
@@ -565,6 +565,34 @@ def emv(n = 10, wilder = false)
...
+        #            # => [0.400..., 0.727..., 0.862..., 0.013..., ...]
+        #
+        #   # first 9 observations are nil
+        #   ts.emsd   # => [ ... nil, 0.285... , 0.258..., 0.243..., ...]
+        #
+        # @return [Daru::Vector] contains EMSD
+        def emsd(n = 10, wilder = false)
+          result = []
+          vec = emv(n, wilder)
+          vec.data.each do |d|
```
v0dro: Why not just `vec.each`?

shahsaurabh0605:
Made the changes.

PR #102, opened 2016-04-17
covariance function for vectors

shahsaurabh0605:
Added `covariance_sample` and `covariance_population` methods for vectors.

### DIFF
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -135,7 +135,7 @@ module Daru
           end
         end
 
-        # Count number of occurences of each value in the Vector
+        # Count number of occurrences of each value in the Vector
         def value_counts
           values = {}
           @data.each do |d|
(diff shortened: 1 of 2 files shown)


lib/daru/maths/statistics/vector.rb
```diff
@@ -700,7 +724,8 @@ def cumsum
         alias :std :sds
         alias :adp :average_deviation_population
         alias :cov :coefficient_of_variation
-        alias :variance :variance_sample    
+        alias :variance :variance_sample   
+        alias :covariance :covariance_sample
```
v0dro: Remove aliases

lib/daru/maths/statistics/vector.rb
```diff
@@ -169,6 +169,30 @@ def variance_population m=nil
...
+        # Sample covariance with denominator (N-1)
+        def covariance_sample other
+          @size == other.size or raise ArgumentError, "size of both the vectors must be equal"
+          mean_x = self.mean 
+          mean_y = other.mean
+          sum = 0
+          (0...size).each do |i| 
+            sum = sum + ((@data[i].nil? || other.data[i].nil?) ? 0 : (@data[i] - mean_x) * (other.data[i] - mean_y))
+          end 
+          covar = sum / (n_valid - 1)
```
v0dro: Just write `sum / (n_valid - 1)`. No need for the variable.

lib/daru/maths/statistics/vector.rb
```diff
@@ -169,6 +169,30 @@ def variance_population m=nil
...
         end
 
+        # Sample covariance with denominator (N-1)
+        def covariance_sample other
+          @size == other.size or raise ArgumentError, "size of both the vectors must be equal"
+          mean_x = self.mean 
+          mean_y = other.mean
+          sum = 0
+          (0...size).each do |i| 
+            sum = sum + ((@data[i].nil? || other.data[i].nil?) ? 0 : (@data[i] - mean_x) * (other.data[i] - mean_y))
```
v0dro: Missing data need only consist of `nil` elements. It can be anything in `missing_values`, even numbers.

lib/daru/maths/statistics/vector.rb
shahsaurabh0605: Alias for `variance` was already present, so won't it be a good idea to add for `covariance` as well?

v0dro: ok fine

PR #103, opened 2016-04-20
Removes implicit to_hash conversions in favor of explicit to_h conversions

gnilrets:
I don't think we should be defining implicit hash conversion functions.  Here's an example where we pass a dataframe to a method and Ruby *implicitly* converts it to a hash and populates the keyword arguments (`kargs`) instead of the array arguments (`args`).  The changes in this commit correctly put the dataframe in `args`.

Unfortunately, this will likely break some client code that uses `to_hash`, but I think it's the right direction.  I can't think of use cases where having an implicit conversion is particularly helpful.

```ruby
require 'daru'

df = Daru::DataFrame.new({ a: [1,2], b: [3,4] })

def mymeth(*args, **kargs)
  puts "args: #{args}"
  puts "kargs: #{kargs}"
end

mymeth(df)
```

    args: []
    kargs: {:a=>
    #<Daru::Vector:70093300385580 @name = a @metadata = {} @size = 2 >
          a
      0   1
      1   2
    , :b=>
    #<Daru::Vector:70093300385140 @name = b @metadata = {} @size = 2 >
          b
      0   3
      1   4
    }

### DIFF
--- a/lib/daru/core/merge.rb
+++ b/lib/daru/core/merge.rb
@@ -29,7 +29,7 @@ module Daru
         end
 
         def hashify df
-          hsh = df.to_hash
+          hsh = df.to_h
           hsh.each { |k,v| hsh[k] = v.to_a }
           hsh
         end
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1065,7 +1065,7 @@ module Daru
         name = row[tree_keys.last]
         if !block
           current[name] ||= []
-          current[name].push(row.to_hash.delete_if { |key,value| tree_keys.include? key})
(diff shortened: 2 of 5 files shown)


v0dro:
@mrkn could you shed some light here? Do you think it's wise to remove implicit hash conversions? Could really use your insight into this given your deep experience with Ruby.

I'm OK with API changes since daru is a very young library and users expect minor API changes.

mrkn:
IMHO, dataframes aren't kind of association arrays, so it is appropriate to remove implicit to_hash conversions.

v0dro:
Alright then.

PR #106, opened 2016-04-27
FIX - Improvement to dup performance

gnilrets:
Vectors were being completely recreated because there was no check
to see if the indexes were the same when the index was not nil.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -296,10 +296,10 @@ module Daru
         when Hash
           create_vectors_index_with vectors, source
           if all_daru_vectors_in_source? source
+            vectors_have_same_index = all_vectors_have_equal_indexes?(source)
             if !index.nil?
               @index = try_create_index index
-            elsif all_vectors_have_equal_indexes?(source)
-              vectors_have_same_index = true
+            elsif vectors_have_same_index
               @index = source.values[0].index.dup
             else
               all_indexes = []
--- a/benchmarks/duplicating.rb
+++ b/benchmarks/duplicating.rb
@@ -1,0 +1,45 @@
+$:.unshift File.expand_path("../../lib", __FILE__)
+
+require 'benchmark'
+require 'daru'
+
+# Check scaling
+base_n = 10000
+0.upto(2) do |iscale|
+  n = base_n * 2**iscale
+
+  df_h = ('a'..'z').map { |v| v.to_sym }.reduce({}) do |h, v|
+    h[v] = Daru::Vector.new(1.upto(n).to_a)
+    h
+  end
+
+  df = Daru::DataFrame.new(df_h)
+
+  Benchmark.bm do |bm|
+    bm.report("dupe (n=#{n})") do
+      df.dup
+    end
+  end
+end
+
+#                   ===== Benchmarks =====
+# System: iMac Late 2013 3.5GHz Core i7
+#
+#       user     system      total        real
+#dupe (n=10000)  0.590000   0.020000   0.610000 (  0.613648)
+#       user     system      total        real
+#dupe (n=20000)  1.170000   0.040000   1.210000 (  1.236629)
+#       user     system      total        real
+#dupe (n=40000)  2.390000   0.070000   2.460000 (  2.511199)
+
+
+
+
+#                   ===== Prior Benchmarks (Daru 0.1.2 - 2707559369c03894a8394714820aabf116b99b20 - 2016-04-25) =====
+# Note that the n here is 100x smaller than above
+#       user     system      total        real
+#dupe (n=100)  0.220000   0.000000   0.220000 (  0.227924)
+#       user     system      total        real
+#dupe (n=200)  0.850000   0.000000   0.850000 (  0.856591)
+#       user     system      total        real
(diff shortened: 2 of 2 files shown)


lib/daru/dataframe.rb
```diff
@@ -295,11 +295,11 @@ def initialize source, opts={}
           end
         when Hash
           create_vectors_index_with vectors, source
+          vectors_have_same_index = all_vectors_have_equal_indexes?(source)
```
v0dro: I don't quite understand how moving the method above the if condition made it faster. In fact that method will now run even if all the values of the supplied Hash are not `Daru::Vector` objects. Could you explain?

Also, there's a problem here. Since the `all_vectors_have_equal_indexes?` method has been moved outside the `if all_daru_vectors_in_source? source` statement, the `all_vectors_have_equal_indexes?` can get an `Array` in the values of the Hash that is `source`, while it is expecting all the values to be `Daru::Vector` objects. For example:

``` ruby

df = Daru::DataFrame.new({a: [1,2,3], b: [2,3,4]})
```

The code passes because calling `index` on `Array` returns an `Enumerator`. See it [here](https://github.com/inside-track/daru/blob/FIX-DupePerformance/lib/daru/dataframe.rb#L2415).

v0dro: Also, the benchmarks for your script on my machine turn to be like so:

I'm using a 32 bit debian computer with Intel(R) Core(TM) i5-2410M CPU @ 2.30GHz and 4 GB RAM.
```
require 'benchmark'
require 'daru'

# Check scaling
base_n = 10000
0.upto(2) do |iscale|
  n = base_n * 2**iscale

  df_h = ('a'..'z').map { |v| v.to_sym }.reduce({}) do |h, v|
    h[v] = Daru::Vector.new(1.upto(n).to_a)
    h
  end

  df = Daru::DataFrame.new(df_h)

  Benchmark.bm do |bm|
    bm.report("dupe (n=#{n})") do
      df.dup
    end
  end
end

#        user     system      total        real
# dupe (n=10000)  0.470000   0.000000   0.470000 (  0.479517)
#        user     system      total        real
# dupe (n=20000)  0.950000   0.010000   0.960000 (  0.957177)
#        user     system      total        real
# dupe (n=40000)  1.920000   0.000000   1.920000 (  1.921814)
# [Finished in 13.1s]
```

gnilrets: `vectors_have_same_index` is referenced in the `if clone` block.  When a dataframe is duplicated, all vectors have the same index, but `vectors_have_same_index` is false because `!index.nil?` will be true.  So when a dataframe is duplicated, all of the vectors have to be recreated, rather than just using the `.dup` method on the vector (line 321 below).

I don't think I fully follow your second point.  Could you provide a test that would fail and I'll work to get it passing?

gnilrets: Are those benchmarks you get without this commit?

v0dro: Yes here's the script with the Ruby version. This is very strange :\

```

require 'benchmark'
require 'daru'

# Check scaling
base_n = 10000
print "daru version : #{Daru::VERSION}\n"
0.upto(2) do |iscale|
  n = base_n * 2**iscale

  df_h = ('a'..'z').map { |v| v.to_sym }.reduce({}) do |h, v|
    h[v] = Daru::Vector.new(1.upto(n).to_a)
    h
  end

  df = Daru::DataFrame.new(df_h)

  Benchmark.bm do |bm|
    bm.report("dupe (n=#{n})") do
      df.dup
    end
  end
end

# daru version : 0.1.2
#        user     system      total        real
# dupe (n=10000)  0.880000   0.000000   0.880000 (  0.879011)
#        user     system      total        real
# dupe (n=20000)  0.970000   0.000000   0.970000 (  0.974217)
#        user     system      total        real
# dupe (n=40000)  1.950000   0.000000   1.950000 (  1.943483)
# [Finished in 15.7s]
 
```

v0dro: I'll just try uninstalling and reinstalling daru hold on.

gnilrets: This commit doesn't change `Daru::VERSION`.  Also, I'm using MRI Ruby 2.2.

v0dro: My second point is that the `all_vectors_have_equal_indexes?` method is expecting a Hash of key value pairs where the values are strictly `Daru::Vector` objects, since this is ensured by the fact that it is inside the `if` block.

The program still works inspite of `source` being a Hash of key value pairs with `Array` objects for values because of duck typing. The [line in question](https://github.com/inside-track/daru/blob/FIX-DupePerformance/lib/daru/dataframe.rb#L2415) passes without error because `Array` has a method called `#index` defined on it.

So you can move the `vectors_have_same_index = all_vectors_have_equal_indexes?(source)` line below the first `if` statement (i.e. `if all_daru_vectors_in_source? source`) so that it is ensured that `source` always has key-value pairs where the values are `Daru::Vector` objects.

Also, I just reinstalled daru 0.1.2 from rubygems and am getting the same result. But I see your point and can merge this with the above changes.

gnilrets:
Ok, here's the change you requested.  

Very strange thing about the benchmarks.  I'm still getting O(N^2) scaling on the current master branch and O(N) with this commit.

PR #107, opened 2016-04-27
[WIP] Refactor dataframe index

lokeshh:
Improve functions related to `[]` and `[]=` and make them independent of any sort of indexing used whether it be `MultiIndex` or `DateTimeIndex`.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -2310,14 +2310,19 @@ module Daru
 
       # Update an existing vector based on its index
       if pos.is_a? Daru::Index
-        # If there are multiple vectors associated with this name
+        # If there are multiple vectors associated with this name modify all of them
+        # This could be in case of Index, MultiIndex as well as DateTimeIndex
         pos.each { |p| @data[@vectors[p]] = v }
       else
         @data[pos] = v
       end
 
     rescue IndexError
-      # Add a new vector
+      # IndexError tells that this vector isn't in the dataframe
+      # Therefore add a new vector with this name
+      # Perhaps not a good idea to use exception handling for this purpose
+      # Doing this because other method like @index.include? are insufficient
+      # for this purpose.
       @vectors |= name
       @data[@vectors[*name]] = v
     end
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -2264,7 +2264,6 @@ module Daru
     end
 
     def insert_or_modify_vector name, vector
-      name = name[0] unless @vectors.is_a?(MultiIndex)
       v = nil
 
       if @index.empty?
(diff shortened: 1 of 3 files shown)


lib/daru/dataframe.rb
```diff
@@ -2308,24 +2307,21 @@ def insert_or_modify_vector name, vector
     end
 
     def assign_or_add_vector name, v
-      #FIXME: fix this jugaad. need to make changes in Indexing itself.
-      begin
-        pos = @vectors[name]
-      rescue IndexError
-        pos = name
-      end
+      name = [name] unless name.is_a? Enumerable
```
v0dro: Why Enumerable?

lokeshh:
Added support to modify multiples vectors at a time. I'm looking forward to add support to add multiple vectors at a time now.

lib/daru/dataframe.rb
```diff
@@ -2302,30 +2301,25 @@ def insert_or_modify_vector name, vector
...
-        @data[pos] = v
-      elsif pos.kind_of?(Daru::Index)
-        pos.each do |p|
-          @data[@vectors[p]] = v
-        end
+      # Update an existing vector based on its index
+      if pos.is_a? Daru::Index
+        # If there are multiple vectors associated with this name
+        pos.each { |p| @data[@vectors[p]] = v }
       else
```
lokeshh: Need to assign correct name to each vector. Is there a function to change name of a vector?

v0dro: Use Vector#rename. You'll need to modify the function to use `Vector#set_name` internally. It's written quite hastily.

lib/daru/dataframe.rb
```diff
@@ -2302,30 +2301,25 @@ def insert_or_modify_vector name, vector
...
-        @vectors = @vectors | [name] if !@vectors.include?(name)
-        @data[@vectors[name]] = v
+        @data[pos] = v
       end
+
+    rescue IndexError
```
v0dro: Provide reasoning in form of comments as to why you're rescuing the IndexError. Its not very apparent.

lib/daru/dataframe.rb
```diff
@@ -2302,30 +2301,25 @@ def insert_or_modify_vector name, vector
...
-      if !pos.kind_of?(Daru::Index) and pos == name and
-        (@vectors.include?(name) or (pos.is_a?(Integer) and pos < @data.size))
-        @data[pos] = v
-      elsif pos.kind_of?(Daru::Index)
-        pos.each do |p|
-          @data[@vectors[p]] = v
-        end
+      # Update an existing vector based on its index
+      if pos.is_a? Daru::Index
+        # If there are multiple vectors associated with this name
```
v0dro: Multiple vectors with same name as in a MultiIndex I assume?

lokeshh: Yes. In addition to that it also assigns multiple vectors in case of Index and DateTimeIndex.

v0dro: Please put that in the comments.

lib/daru/dataframe.rb
lokeshh: Rescuing IndexError here is for the situation when the index which the user specified isn't in the data frame. This suggests that the user intended to add a new vector rather than modify an existing vector. I know it's kind of not a good practice to use error handling in this way but in current scenario nothing else was working and was resulting into making the code more complex. For example initially I had used `@index.include? name` to see if an index exists or not but it was giving false for say `[:a, :foo]` when `[:a, :foo, :bar]` exists and `@index` is a MultiIndex. The same goes for DateTimeIndex.

v0dro: OK works. But put that in the comments for future reference. Also specify the reason for using error handling over the other approach.

zverok: I assume such problems suggest we need extend indexes API (so that there'll be an easy way to ask "if index includes this").
Also, I can't see much specs to cover this problem (variants of indexes).

v0dro: Point.

lokeshh: Agreed.

lokeshh: For now, I am going to stay here (dataframe.rb) and take a look at how index class is being used to come up with solutions for other places where index type is being checked.

v0dro:
whats the status of this PR?

lokeshh:
I will began working on it soon. I'm little caught up in exams: D

v0dro:
Haha sure. Take your time!

PR #116, opened 2016-05-05
Initialze Daru::Index from range

lokeshh:


lib/daru/index.rb
```diff
@@ -24,7 +24,7 @@ def self.new *args, &block
       source = args[0]
 
       idx =
-      if source and source[0].is_a?(Array)
+      if source.is_a?(Array) and source[0].is_a?(Array)
```
v0dro: why introduce this conditional?

spec/index_spec.rb
```diff
@@ -37,6 +37,12 @@
       expect(idx.to_a).to eq(['speaker', 'mic', 'guitar', 'amp'])
     end
 
+    it "creates an Index from Range" do
+      idx = Daru::Index.new 1..5
+
+      expect(idx).to eq(Daru::Index.new [1, 2, 3, 4, 5])
+    end
```
v0dro: everything else is fine.

lib/daru/index.rb
```diff
@@ -51,8 +51,8 @@ def map(&block)
 
     def initialize index
       index = 0                         if index.nil?
-      index = Array.new(index) { |i| i} if index.is_a? Integer
-      index = index.to_a                if index.is_a? Daru::Index
+      index = index.times.to_a          if index.is_a? Integer
+      index = index.to_a                if index.is_a? Enumerable
```
zverok: Maybe `case` will be cleaner here?

lib/daru/index.rb
```diff
@@ -51,8 +51,8 @@ def map(&block)
...
 
       @relation_hash = {}
       index.each_with_index do |n, idx|
```
zverok: BTW, as far as I can see, the code of this each_with_index could be replaces with just `@relations_hash = index.each_with_index.to_h`

lib/daru/index.rb
lokeshh: Range is not an array, so source[0] was producing an error.

zverok: `source.respond_to?(:first) && source.first.is_a?(Array)` maybe?..

lib/daru/index.rb
```diff
@@ -50,15 +50,17 @@ def map(&block)
...
-      index.each_with_index do |n, idx|
-        @relation_hash[n] = idx 
+      index = case
+        when index.nil?
+          []
+        when index.is_a?(Integer)
+          index.times.to_a
+        when index.is_a?(Enumerable)
+          index.to_a
       end
```
zverok:
```ruby
index = case index
  when nil
  when Integer
  when Enumerable
  else
  end
```

lib/daru/index.rb
lokeshh: Sure. It shows the intent better. Thanks

lib/daru/index.rb
lokeshh: Beautiful!

lib/daru/index.rb
```diff
@@ -50,15 +50,17 @@ def map(&block)
...
-      @relation_hash = {}
-      index.each_with_index do |n, idx|
-        @relation_hash[n] = idx 
+      index = case index
+        when nil
+          []
+        when Integer
+          index.times.to_a
+        when Enumerable
+          index.to_a
```
zverok: Sorry for being pedantic, but I think we still need `else` here, or we have pretty cryptic error:
```ruby
Daru::Index.new('foo')
# => NoMethodError: undefined method `each_with_index' for nil:NilClass
```

zverok: (And spec for this case would be great also)

lib/daru/index.rb
```diff
@@ -50,15 +50,17 @@ def map(&block)
...
       end
 
+      @relation_hash = index.each_with_index.to_h
+
```
zverok: Now we can add `.freeze` (from line 64) to the end of this statement, I think.

lib/daru/index.rb
lokeshh: No, it's good. Keep telling me more places where it can be improved. Thanks!

lib/daru/index.rb
```diff
@@ -24,7 +24,7 @@ def self.new *args, &block
       source = args[0]
 
       idx =
-      if source and source[0].is_a?(Array)
+      if source.respond_to?(:first) && source.first.is_a?(Array)
```
v0dro: I still don't understand the reason behind replacing the previous line with this one. Also, don't squash commits until the PR is ready. Once the commit is gone, the comments made on it cannot be seen.

lokeshh:
@v0dro This is ready to be merged.

lokeshh:
@v0dro The `to_h` works to seem fine in Daru even when using Ruby 2.0. I am currently looking at what the problem is.

lokeshh:
This is weird:

```ruby
[4] pry(main)> RUBY_VERSION
=> "2.0.0"
[5] pry(main)> a = [1, 2, 3].each_with_index
=> #<Enumerator: ...>
[6] pry(main)> a.method(:to_h).source_location
=> ["/home/ubuntu/.rvm/gems/ruby-2.0.0-p643/gems/backports-3.6.8/lib/backports/2.1.0/enumerable/to_h.rb", 4]
```

lokeshh:
Is there any way to check Daru version inside pry?

lokeshh:
Finally, here the fix:

Adding `  spec.add_development_dependency 'nmatrix', '~> 0.2.1'` to specfile of statsample-glm solves the issue.

v0dro:
Why was this happening?

lokeshh:
I am investigating.

lokeshh:
Here's a permanent fix and it explains the issue: https://github.com/v0dro/daru/pull/126

PR #119, opened 2016-05-06
[WIP] Applying Rubocop + code cleanup

zverok:
Hey.

Here's some work-in-progress I'd like to propose and discuss.

**The idea**

I am adding Rubocop to daru, and want to change code so that RuboCop will be happy with it. Some things, like personal code styles of maintainers, or other project guidelines, can be outlined in .rubocop.yml

**Reasons**

1. As daru becames mature project, and gets more and more contributors, it becames valuable to maintain common code style/standarts, and Rubocop allows to enforce them;
2. Many of Rubocop's suggestions are not only about style, but also about quality (too complex or unmaintainable code), and my experience says it is always good to at least try to follow all the suggestions.

**Plan**

Currently, I've added the Rubocop, and generated "todo" config from code base: https://github.com/zverok/daru/blob/rubocop/.rubocop_todo.yml

Also, I've already fixed one cop (pretty huge, it says "don't use reduce when you can use each_with_object", but while fixing it, I've also simplified several things where possible).

Then, I'm planning to go through that TODO cop-by-cop, and either fix them, or, if necessary, move to main `.rubocop.yml` as a daru accepted style. While doing so, I'll not change any external APIs, or even rewrite large portions of code (like renaming instance variables and so on), just cleanup and necessary small refactorings.

Then, I think Rubocop check should be added to Travis for CI checking on new pull requests.

So, my questions for @v0dro and other maintainers:
* should I proceed (or maybe it is not a best time to do this work?.. or it is generally considered not welcome?)
* whether there are some already known style preferences (like map over collect, spaces inside/around blocks, parens on method definitions) I can set up right now?
* whether it will be more wise to do One Huge pull request when I'm done (of course, with proper rebasing from what would be master state to that time) or ton of smaller ones from time to time?


lib/daru/maths/statistics/vector.rb
```diff
@@ -87,12 +87,8 @@ def max_index
...
 
         def frequencies
-          @data.inject({}) do |hash, element|
-            unless element.nil?
-              hash[element] ||= 0
-              hash[element] += 1
-            end
-            hash
+          @data.each_with_object(Hash.new(0)) do |element, hash|
+            hash[element] += 1 unless element.nil?
```
v0dro: what happens to the `hash[element] ||= 0`?

lib/daru/monkeys.rb
```diff
@@ -6,32 +6,19 @@ class Array
...
+    duplicated = group_by { |n| n}
+      .select { |_, g| g.size > 1}.map(&:first)
+    counter = duplicated.collect { |n| [n, 0] }.to_h
+
+    collect do |n|
+      if counter.key?(n)
+        counter[n] += 1
+        format('%s_%d', n, counter[n])
+      else
+        n
```
v0dro: great refactor!

v0dro:
Looks good. I think rubocop would do a fine job. You can send in new commits for now and then squash them later into a single commit.

lib/daru/maths/statistics/vector.rb
zverok: It went to default Hash value (`each_with_object(Hash.new(0))`). 
As far as I can see, effect should be exactly the same, and no specs were broken.

zverok:
OK, will do!

zverok:
So, what's done for now:
* I fixed almost all _non-substantial_ problems rubocop found ("non-substantial" means the fix didn't require some huge refactorings);
* substantial checks are left for the next round, they are listed in [.rubocop_todo.yml](https://github.com/zverok/daru/blob/rubocop/.rubocop_todo.yml); fixing of those would have a great value, but it will take time, maybe current work can be merged now, and then all those `Metric/` problems fixed in future? in meantime, new contributors will already be exposed to Rubocop checks;
* There is [.rubocop.yml](https://github.com/zverok/daru/blob/rubocop/.rubocop.yml) now, with several sections; most interesting (= to discuss and fix) is [here](https://github.com/zverok/daru/blob/rubocop/.rubocop.yml#L65)

If you are happy with current state of things, I'll squash it and it could be merged.

v0dro:
How about this: take your time and complete the rubocop changes and I'll merge them all in one go. I don't want the codebase to have TODO files and other extras that we'd rather not have. Also, since these are predominantly code style changes and do not affect the functionality as such, they can be safely done until the next release, which is not due for at least a month more.

v0dro:
About methods starting with `has_` or `is_`: yes they should be banned, now that I think of them.

We'll alias the ones that are currently there so that the API remains the same but we'll be better off without more of them.

v0dro:
Also, is it possible to include rubocop in the normal workflow of writing a change and submitting a commit as a PR? That way future contributors can correct themselves in case they go wrong.

zverok:
> How about this: take your time and complete the rubocop changes and I'll merge them all in one go.

No problem. It will take several more days, but I'll do it.

> Also, is it possible to include rubocop in the normal workflow of writing a change and submitting a commit as a PR? That way future contributors can correct themselves in case they go wrong.

Already did it in my PR. It is added to .travis.yml and CONTRIBUTING.md (That was one of reasons I wanted to have current work merged ASAP)

zverok:
BTW, you can preview amount of changes which are upcoming to this pull request (if we decide to squash both "style" and "refactor" rubocop fixes) in a separate brunch: https://github.com/zverok/daru/tree/rubocop2-refactor

TL;DR: pretty huge.

v0dro:
You can separate them into different commits according to what you're changing, if you want.

lib/daru/index.rb
```diff
@@ -23,19 +23,14 @@ class << subclass
...
-              source.all? { |e| e.is_a?(DateTime) }
-          Daru::DateTimeIndex.new(source, freq: :infer)
-        else
-          i = allocate
-          i.send :initialize, *args, &block
-          i
-        end
-
-      idx
+      if source && source[0].is_a?(Array)
```
v0dro: use `and` instead of `&&`?

lib/daru/maths/statistics/vector.rb
v0dro: ah

zverok:
See, that's up to you. 
My proposition is two big changes: current PR (which is ready to merge, and after that everyone would be obliged to respect Rubocop), and forthcoming, which would be HUGE refactoring we can discuss, I fear, through several weeks or even monthes. 
But you are the project owner, so judge by yourself.

lib/daru/index.rb
zverok: Nope, we should never. 
`&&` is logical operator, `and` is control flow operator (which, in fact, Rubocop bans even for control flow, but I'm not completely agree about it).

v0dro:
Ah. I thought the current PR will extended to include the refactoring too.

If you're sure this ready for merge, I'll do that. Also, is it possible to remove the TODO files from the repo?

lib/daru.rb
```diff
@@ -26,19 +26,20 @@ module Daru
     10 => 31,
     11 => 30,
     12 => 31
-  }
+  }.freeze
 
-  SPLIT_TOKEN = ','
-  class << self
-    @@lazy_update = false
+  @lazy_update = false
```
v0dro: This is a class variable because it can be changed by the user as `Daru.lazy_update = true`, which does missing data updating lazily.

lib/daru.rb
```diff
@@ -26,19 +26,20 @@ module Daru
...
+
+  SPLIT_TOKEN = ','.freeze
```
v0dro: This again can be changed by users with `Daru::SPLIT_TOKEN = '.'`

lib/daru/accessors/gsl_wrapper.rb
```diff
@@ -3,33 +3,33 @@ module Accessors
...
       def median
-        GSL::Stats::median_from_sorted_data(@data.sort)
+        GSL::Stats.median_from_sorted_data(@data.sort)
       end
-      
-      def variance_sample(m)
-        @data.variance_m
+
+      def variance_sample(_m)
+        @data.variance_m # FIXME: I (zverok) suppose it may be variance(m) meant here
```
v0dro: That's right please change it to `variance(m)`.

v0dro: The default of `m` should be the mean of `@data`.

lib/daru/accessors/mdarray_wrapper.rb
```diff
@@ -1,9 +1,7 @@
 module Daru
   module Accessors
-
```
v0dro: Lets just remove this file. MDArray is not being maintained anymore. jNMatrix will be the future for JRuby arrays.

lib/daru/accessors/nmatrix_wrapper.rb
```diff
@@ -32,47 +32,47 @@ def [] *index
         return @data[*index] if index[0] < @size
         nil
       end
- 
+
       def []= index, value
-        raise ArgumentError, "Index #{index} does not exist" if 
-          index > @size and index < @data.size
+        raise ArgumentError, "Index #{index} does not exist" if
+          index > @size && index < @data.size
```
v0dro: why favor `&&` over `and`? Its more intuitive.

lib/daru.rb
zverok: With my currenty `@lazy_update` class instance variable branch:
```
2.2.4 :001 > Daru.lazy_update
 => false 
2.2.4 :002 > Daru.lazy_update = true
 => true 
2.2.4 :003 > Daru.lazy_update
 => true 
```

I assume, that what I changed, should still work.

lib/daru/core/group_by.rb
```diff
@@ -29,13 +35,13 @@ def initialize context, names
...
-        if multi_indexed_grouping?
-          Daru::MultiIndex.from_tuples @groups.keys
-        else
-          Daru::Index.new @groups.keys.flatten
-        end
+          if multi_indexed_grouping?
+            Daru::MultiIndex.from_tuples @groups.keys
+          else
+            Daru::Index.new @groups.keys.flatten
+          end
```
v0dro: Are you sure about these indents here?

lib/daru.rb
zverok: If it can be changed by user, it should not be a constant, obviously.  Though, after `.freeze`, your example still working. What is not working, is things like `Daru::SPLIT_TOKEN << 'xxx'`

lib/daru/core/group_by.rb
```diff
@@ -267,11 +270,11 @@ def apply_method method_type, method
         multi_index = multi_indexed_grouping?
         rows, order = [], []
 
-        @groups.each do |group, indexes|
+        @groups.each do |_group, indexes|
```
v0dro: any specific reason for the `_`?

lib/daru/core/merge.rb
```diff
@@ -37,71 +37,67 @@ def hashify df
...
         end
 
         def arrayify_with_sort_keys(size, df_hash, on)
-
           # Converting to a hash and then to an array is more complex
           # than using df.to_a or df.map(:row).  However, it's
           # substantially faster this way.
 
-          idx_keys = on.map { |key| df_hash.keys.index(key) }
+          # idx_keys = on.map { |key| df_hash.keys.index(key) }
```
v0dro: ?

lib/daru/accessors/nmatrix_wrapper.rb
zverok: (Already answered this somewhere) 
Just because its like Ruby works. `&&` is logical and, `and` is control-flow and; it differs in all possible ways, including operator precedence.

lib/daru/core/group_by.rb
zverok: Yep.
```ruby
variable = 
  statement
```
is pretty common formatting, and with if's they are commonly turn into:
```ruby
variable = 
  if this
    foo 
  else
    bar
  end
```

lib/daru/core/group_by.rb
zverok: It is Rubocop-foced agreement for "not used argument". You can use `_` prefix or  name the whole argument `_` to mark it as skipped.

lib/daru/core/merge.rb
zverok: It is never used. I should've deleted it, in fact, but sometimes just commented out things that are not used.

zverok:
> If you're sure this ready for merge, I'll do that. Also, is it possible to remove the TODO files from the repo?

Currently, `.rubocop_todo.yml` plays two roles:
1. marks what should be fixed
2. makes those checks disabled (`.rubocop_todo.yml` is included in `.rubocop.yml` [here](https://github.com/v0dro/daru/pull/119/files#diff-11a0d7906801d9dea0eccb85667ad811R1))

If we'll just remove this file, corresponding checks would be red.

In fact, having `.rubocop_todo.yml` in repo is acceptable practice (even Rubocop itself have it), but I can move all checks from there into `# TODO` section in main `.rubocop.yml` if you think it would be cleaner.

lib/daru.rb
v0dro: Ok then let it be.

lib/daru/core/group_by.rb
v0dro: alright.

lib/daru.rb
v0dro: But having it as a class variable makes it explicit that the variable is specific to the whole class and not just one instance.

lib/daru/core/group_by.rb
v0dro: ok

lib/daru/core/merge.rb
v0dro: ok fine.

v0dro:
Ok this is fine except that class variable thing. I'll need some time to go over this PR though. Will merge once I'm done.

lib/daru.rb
zverok: Currently, almost ALL guides and tutorials explicitly bun class variable (in favor of "instance variable at module level"). Though, I'm not intending to fight for this one thing till the end of the world :) Chaning back.

v0dro: Ah. Any particular reason for this? I'm ready to keep it an instance variable if the reason is convincing enough.

gnilrets: One of the main arguments I've seen against class instance variables is that when they are modified in an inherited class, they also affect the parent class.

Here's one of my go-to articles on class variables that goes into more depth:  http://www.railstips.org/blog/archives/2006/11/18/class-and-instance-variables-in-ruby/

zverok: There are [many](http://archive.oreilly.com/pub/a/ruby/excerpts/ruby-best-practices/worst-practices.html) [rants](http://4thmouse.com/index.php/2011/03/20/why-class-variables-in-ruby-are-a-bad-idea/) on them. Neither of the problems listed are applicable to our case, but I think class variables are considered code smell nowadays unconditionally.

zverok: BTW, I've already changed this back: https://github.com/zverok/daru/commit/9d4d333095d1e7ea98318d8d7ff2ef0ab9570c25

Don't sure why it is not visible in pull request.

v0dro:
I read the stuff against class variables and I think it's justified to use a module level instance variable. Could you change it back?

zverok:
Did it. Also, I've merged latest master from main repo and fixed all problems, so it should be safe to merge now.

v0dro:
OK i've gone through the code and it looks good. Just give me the green signal once and I'll merge.

zverok:
Great!
I think you can merge it right now. 
It has no conflicts with current branch and introduces Rubocop for all upcoming contributions. 
Second level of refactoring I'll be doing in a separate branch, and it will be pretty slow, I fear.

PR #123, opened 2016-05-10
Large codebase refactoring

zverok:
**Warning!** This is definitely work-in-progress (and in early stages of it!). Pull request is created so @v0dro and others could observe the progress, comment and tell their preferences on code style.

Target of this pull request is to refactor daru codebase (**without** any changes in functionality, except maybe for small service methods additions) so that it conform to some pretty strict code quality metrics, both trivial (like method length/module length) and semantic (like perceived complexity and assignment branch condition size), with other eventual cleanups here and there.

Also, as a part of this pull request I propose (but can revoke this proposal) to update minimal Ruby version requirement to 2.1, which is lowest of currently supported versions, and unlike 2.0, provides proper handling of keyword arguments.

So, you can read through it, or ignore it for now :) I'll tell when I am finished!


v0dro:
See if you can find any speed bottlenecks through the course of your refactoring.

zverok:
> See if you can find any speed bottlenecks through the course of your refactoring.

BTW, one of targets of refactorin is increasing code reuse (so that complex things like `group_by` or `concat` could be expressed in terms of other methods). Than I'll profile it with existing benchmarks, and hope that greater code reuse means lesser bottlenecks, which could be clearly seen than. And once, in a month or so, I'll return with completely mergeable proposition :) Or so I hope.

lib/daru/dataframe.rb
```diff
@@ -2466,5 +2451,15 @@ def prepare_sort_block vector_order, opts
         left <=> right
       end
     end
+
+    def verify_error row, test, id, i
```
v0dro: This should be a private method.

v0dro:
@lokeshh keep having a look and reviewing the code in this PR. It will teach you many new things about good Ruby practices. You'll also help @zverok avoid potential pitfalls.

lokeshh:
Sure, that's a great opportunity.

lib/daru/vector.rb
```diff
@@ -1159,6 +1156,12 @@ def guard_sizes!
       end
     end
 
+    def guard_type_check value
```
v0dro: private again.

lib/daru/vector.rb
```diff
@@ -1159,6 +1156,12 @@ def guard_sizes!
...
+      @possibly_changed_type = true \
+        if @type == :object && (value.nil? || value.is_a?(Numeric)) ||
+           @type == :numeric && !value.is_a?(Numeric) && !value.nil?
```
v0dro: are you sure about the precedence here?

I think it should be:
``` ruby

 if (@type == :object && (value.nil? || value.is_a?(Numeric))) || (@type == :numeric && !value.is_a?(Numeric) && !value.nil?)
```

lib/daru/vector.rb
zverok: I believe you somehow comment on outdated diff, those methods are already private. Maybe some GitHub bug?..

v0dro: Oh my bad.

lib/daru/vector.rb
zverok: `&&` has higher precedence than `||`, so parentheses are unnecessary, and your statement is exactly equivalent to mine.

I believe parentheses previously was needed because code have used control flow operators `and` and `or` (which have same precedence), instead of boolean algebra operators `&&` and `||` (which have correct boolean algebra precedence).

lib/daru/vector.rb
```diff
@@ -531,36 +533,42 @@ def all? &block
...
-      index = index.map { |i| old_index[i] }
+      index.map! { |i| old_index[i] }
 
       Daru::Vector.new(vector, index: index, name: @name, metadata: @metadata.dup, dtype: @dtype)
     end
 
+    DEFAULT_SORTER = lambda { |(av, ai), (bv, bi)|
+      # FIXME: I suspect it can be replaced with just
+      #   [av || -Float::INFINITY, ai] <=> [bv || -Float::INFINITY, bi]
+      # but need more thorough check
```
lokeshh: `[av || -Float::INFINITY, ai] <=> [bv || -Float::INFINITY, bi]` would fail I think when `av` is something other than integer, say String and `bv` is nil.

zverok: Got it, thanks.

lib/daru/dataframe.rb
```diff
@@ -1492,39 +1396,19 @@ def report_building(b) # :nodoc: #
...
-        left <=> right
-      end
+      opts = prepare_sort_opts vector_order, opts
+      block = prepare_sort_block vector_order, opts
 
       idx = (0..@index.size-1).sort(&block)
-
-      old_index = @index.to_a
-      self.index = Daru::Index.new(idx.map { |i| old_index[i] })
+      new_index = @index.reorder(idx)
```
lokeshh: `idx = (0..@index.size-1).sort(&block)` can be replaced with `idx = @index.size.times.sort(&block)` I think.

zverok: Good catch!

lib/daru/dataframe.rb
```diff
@@ -2282,8 +2274,8 @@ def prepare_sort_opts vector_order, opts
         by: {}
       }.merge(opts)
 
-      opts[:ascending]   = coerce_sort_order vector_order, opts[:ascending]
-      opts[:handle_nils] = coerce_handle_nils vector_order, opts[:handle_nils]
+      opts[:ascending]   = coerce_boolean_array opts[:ascending], vector_order.size, ':ascending'
```
v0dro: Just supply `:ascending` there?

lib/daru/dataframe.rb
```diff
@@ -2282,8 +2274,8 @@ def prepare_sort_opts vector_order, opts
...
+      opts[:handle_nils] = coerce_boolean_array opts[:handle_nils], vector_order.size, ':handle_nils'
```
v0dro: Here too.

lib/daru/dataframe.rb
zverok: Ah, ok :) Went through set of changes here, just forgot the cleanup after last change. Though, current code looks like it WANTS to be simplified into something like `coerce_boolean_array opts, :ascending, vector_order.size` (and sets/gets the key inside).

zverok:
OK, guys, now let's talk!

The current version of code in this PR is "earliest possible version" of **ready refactoring**.
Main highlights:

1. There's ton of `FIXME`s in code here and there, which mark either places I'd want to discuss, or unfinished work (see below about my definition of "unfinished");
2. Current code is **working** in sense "it passes all specs";
3. As far as I understand, there are places either completely uncovered by specs (plotting, html formatting for IRuby), or with uncovered corner cases (for ex., see [this note](https://github.com/v0dro/daru/pull/123/files#diff-90f6d5516ed2b51bf8fd31c120e1ffa3R3) about my reimpl. of join)
4. So, the only way to have the refactoring finished (and that's my plan I'd like to discuss) is the following:
  * create new branch from current master;
  * extend specs coverage to some Really Large percents;
  * merge this branch into my current refactor branch and make sure that all new covered cases are working
  * PROFIT!
5. I've added `profile/` folder and several scripts to do detailed profiling of various aspects, now it should be pretty easy to do other;
6. As far as I know now (with the help of profiling), there are only two really serious bottlenecks:
  * `Index#[]` and `Vector#[]` -- they have so many features that when used in "simplest" form (I know an index, just give me a value), are still severely slowed down by all of those checks and options; **proposition**: semi-internal methods for "fast get by index", named like `Vector#_get(idx)`, and use only them in all complex algorithms;
  * (the biggest and the ugliest one) on any "row-scanning" dataframe tasks, on large dataframes there are tons of temporary vectors constructed (in `DataFrame#access_row`), and it is not cheap. I already have some solution in mind, just want to finish the "global refactoring" before trying it.
7. There are some utility methods added to several classes; they are needed for cleaner code, but in most cases seems of general interest/usefullness.

What I expect now?
* code review/discussions;
* FIXME discussions;
* nothing more, until p.(4) will be implemented.

It is not urgent at all, just communicating.

v0dro:
How about having specialized APIs for referring to a particular index  for power users and let the current `#[]` API be the way it is?

About row operations will it be faster if the whole dataframe is first transposed internally? If the number of objects created are lesser than that what's currently being made (and it's faster), I'm game!

v0dro:
Also it would be better if we keep the refactoring and code coverage PRs separate. I think the later commits here are the same as in #128 ?

zverok:
> How about having specialized APIs for referring to a particular index for power users and let the current #[] API be the way it is?

Yep, that was the kind of things I was thinking about.

> About row operations will it be faster if the whole dataframe is first transposed internally? If the number of objects created are lesser than that what's currently being made (and it's faster), I'm game!

My current plan (which in any case I will try only after my current PRs ready/merged) was something I've already described in emails: just have ONE row object for iteration, kinda like this (pseudo-code):

```ruby
class DataFrame
  def each_row
    row_accessor = RowAccessor.new(self)
    nrows.times do |i| 
      row_accessor.current_row = i 
      yield(row_accessor)
    end
  end
end
```

It needs further evaluation, but I suppose it could provide _dramatic_ speedup with almost now drawbacks.

zverok:
> Also it would be better if we keep the refactoring and code coverage PRs separate. I think the later commits here are the same as in #128 ?

I'm merging `coverage` branch here from time to time, just for move further in creating fully refactored and covered code. In fact, maybe it is not the best solution, so I can stop this premature merging, and just concentrate on `coverage` for now.

v0dro:
Yeah works. I like the `RowAccessor` solution, though I'm not sure how you'll serve it as a Daru::Vector inside the block of `each_row` and preserve indexing and if this will provide a speed boost.

zverok:
About RowAccessor -- in [this commit](https://github.com/zverok/daru/commit/7abf658f5e57b24b15c27a552a4eb7d1771628fb) I have (very rough) prototype, using this accessor idea, and some of `#at`, which already passes (most of) the tests, and performs much better:

* benchmarks/where_vs_filter.rb, current master:
```
where  0.000000   0.000000   0.000000 (  0.001856)
filter_rows  0.020000   0.000000   0.020000 (  0.014838)
```

* same benchmark, with `RowProxy`:
```
where  0.000000   0.000000   0.000000 (  0.001343)
filter_rows  0.010000   0.000000   0.010000 (  0.004331)
```
It is now 3 (not 8) times slower, and there is still huge room for improvement (as `filter_rows` currently uses `where` under the hoods, it definitely can't perform faster).

(To be honest, I'm pretty assured Ruby is too expressive to borrow bad Python idea about `df.where(df.eq() | dft.lt())`, so the ultimate goal is to make bare `filter_rows` fast enough to have `where` and entire BoolArray ideas completely unnecessary.)

v0dro:
Beautiful! I like your approach.

My only concern is the `#pos` and `#at_pos` method, which we're currently sorting out in Lokesh's PR about Categoricals in #134 

We'll until that is sorted, then this can be merged. It's mostly finalized as of now.

v0dro:
Are you sure that all methods applicable to Vector will be applicable here as well?

Also, there `where` clause is a Ruby idea borrowed from [Arel](https://github.com/rails/arel) =P

No python involved.

v0dro:
Now that I think of it I'm also a little concerned of yielding RowProxy objects. We could use some opinions here.

@gnilrets @mrkn @dansbits what are your thoughts? Do you it's alright to yield RowProxy objects instead of `Daru::Vector` objects in the `each_row`? It's really fast now that's for sure.

zverok:
`#at_pos` and `RowProxy` and stuff are in the separate branch, not included in this PR. That are too rough currently, just experimenting; I'll provide separate "speedup" PR with it, when (or "if") it would be production-ready. Current PR is just refactoring, without much functionality change.

And I'd like first https://github.com/v0dro/daru/pull/128 to be discussed and merged (so I could merge master with coverage here and recheck if I didn't break something during refactoring).

> No python involved.

Hm, I thought that trick with producing "boolean arrays" by applying operations to same dataframe resembles pandas' approach (that part, where constructing arguments for `where`, you need to call some methods on the same dataframe, not much DRY)

v0dro:
Yeah the internals are a pandas approach, but the interface is Ruby.

gnilrets:
I don't think I've made any use of having a row be a `Daru::Vector` when using `each_row`.  Due to the poor performance, I avoid using `each_row` even in cases where it makes the most sense.  Anything to improve this would be welcomed!

v0dro:
> I don't think I've made any use of having a row be a Daru::Vector when using each_row. Due to the poor performance, I avoid using each_row even in cases where it makes the most sense. Anything to improve this would be welcomed!

Right. Now that I think of it, you generally don't modify elements inside an `#each` block anyway, so I think we can go ahead with the proxy implementation. @zverok make sure you supply that in a separate PR. Also I think the Statistics and Arithmetic modules might need minor modifications.

gnilrets:
That being said, one bit of functionality that I'd like to see in `each_row` (or some equivalent), would be able to create a new vector that is based on some complex operation between two vectors on a row-by-row-basis.

zverok:
I think it is still possible to achieve with `RowProxy` (by wisely defined `[]=` method).
Though, that is not the most easy way, I'm dreaming of an API like 
```ruby
df.add_vector(:c) { |row| "#{row[:a]}-#{row[:b]}" }
# or even  
df.add_vector(:c) { |row| "#{row.a}-#{row.b}" }
```
...or something.

gnilrets:
What about multiple targets?  How would we do something like splitting up an array into separate vectors?

````ruby
df[:some_array].each_row do |row|
  row[:first_element] = row[:some_array].first
  row[:last_element] = row[:some_array].last
end
````

It would seem tedious and expensive separate this into multiple `df.add_vector(...)` steps, especially if the split operation is more expensive that a simple first/last.

**Comment by v0dro** [comment]
A major problem with modifying the dataframe inside an `each_row` block is that you're adding one element at a time to the DataFrame, and daru isn't designed to handle that. If you suddenly break out of the loop, it'll be left with an incomplete Vector, which will cause a malfunction. So in case of `each_row`-style looping we should either ensure that loops run to completion (I don't think that's possible) or not provide such a functionality altogether and let the user use methods like `DataFrame#add_vector` for adding new vectors, which are guaranteed to run to completion.

**Comment by gnilrets** [comment]
Ok, I'm fine with having to initialize empty vectors as placeholders.  But I would like to make sure we keep support for mapping one vector to multiple vectors and vice versa.

**Comment by v0dro** [comment]
Yes that will always be there.

**Comment by zverok** [comment]
OK, now I've remerged master with coverage here and fixed all errors occuring from that.
An I (with a bit of dread) assume it can be merged also.

Just review the changes I've made, there are tons of them.

**Comment by v0dro** [comment]
Cool. Will review and merge.

File: lib/daru/core/merge.rb [hunk 8]
```diff
@@ -1,205 +1,136 @@
...
-    # dataframes and vectors.
-    # @private
     module Merge
       class << self
         def join df1, df2, opts={}
-          on = opts[:on]
-
-          mf = MergeFrame.new df1, df2, on: on
-          mf.send opts[:how], {}
+          MergeFrame.new(df1, df2, opts).join
```
v0dro: Wow that's some great refactoring. Are you sure it's as fast/faster than the previous implementation? Also, what if we keep the profile directory instead of gitignoring it?

lib/daru/dataframe.rb
```diff
@@ -112,33 +112,17 @@ def from_plaintext path, fields
...
-        first = source.first
-
         raise SizeError, 'All vectors must have same length' \
-          unless source.all? { |v| v.size == first.size }
-
-        index = []
-        opts[:order] ||=
-          case first
-          when Daru::Vector # assume that all are Vectors
-            # FIXME: Not sure, why index is set from vectors
```
v0dro: Its because Vectors have names which can used for creating the index, while other data structures dont.

lib/daru/core/merge.rb
zverok: > Are you sure it's as fast/faster than the previous implementation? 

Yes, benchmarked it against current version thoroughly, it is _slightly_ faster, like this:
master:
```
Inner join (n=10000)  0.280000   0.010000   0.290000 (  0.288439)
Outer join (n=10000)  0.290000   0.000000   0.290000 (  0.297612)
Inner join (n=20000)  0.640000   0.000000   0.640000 (  0.637510)
Outer join (n=20000)  0.640000   0.000000   0.640000 (  0.644443)
Inner join (n=40000)  1.190000   0.000000   1.190000 (  1.186510)
Outer join (n=40000)  1.350000   0.000000   1.350000 (  1.352705)
```
my version:
```
Inner join (n=10000)  0.200000   0.000000   0.200000 (  0.193476)
Outer join (n=10000)  0.260000   0.000000   0.260000 (  0.269912)
Inner join (n=20000)  0.390000   0.010000   0.400000 (  0.409725)
Outer join (n=20000)  0.530000   0.010000   0.540000 (  0.535421)
Inner join (n=40000)  0.980000   0.000000   0.980000 (  0.983138)
Outer join (n=40000)  1.230000   0.000000   1.230000 (  1.233034)
```

> Also, what if we keep the profile directory instead of gitignoring it?

I'm gitignoring only profiler output (which is in any case temporary and local to my machine), you can reproduce it easily with `bundle exec ruby profile/<script-in-interest>rb`

lib/daru/dataframe.rb
zverok: What if rows are Vectors (= we can deduce index from them), but there IS explicitly provided order (`||=` condition doesn't work)?..

lib/daru/core/merge.rb
v0dro: Alright.

lib/daru/dataframe.rb
v0dro: The `:order` provided by the user should over ride the indexes pointed to by the Vectors. The `||=` in that case should not be there.

zverok: OK, I think I've got it now, will add the spec.

lib/daru/dataframe.rb
```diff
@@ -1322,34 +1116,24 @@ def rename_vectors name_map
     # Return the indexes of all the numeric vectors. Will include vectors with nils
     # alongwith numbers.
     def numeric_vectors
-      numerics = []
-
-      each_vector_with_index do |vec, i|
-        numerics << i if vec.type == :numeric
-      end
-      numerics
+      # FIXME: Why _with_index ?..
```
v0dro: It returns the index name so with index.

lib/daru/iruby/templates/dataframe.html.erb
```diff
@@ -0,0 +1,36 @@
+<table>
```
v0dro: This is awesome!

lib/daru/dataframe.rb
zverok: Ah, missed it!

zverok:
:+1:

mrkn:
Great job!!

PR #126, opened 2016-05-13
Include backports to add support for Ruby 2.0

lokeshh:
Solves the problem of build failing in statsample-glm with Ruby 2.0. 

The reason why tests are still passing is that backports gets added to Daru because it's a development time dependency but when Daru is installed on a machine with Ruby 2.0, it fails.

This fix includes backports as run time dependency to Daru and thus fixes the problem.

### DIFF
--- a/daru.gemspec
+++ b/daru.gemspec
@@ -51,6 +51,7 @@ EOF
 
   spec.add_runtime_dependency 'reportbuilder', '~> 1.4'
   spec.add_runtime_dependency 'spreadsheet', '~> 1.1.1'
+  spec.add_runtime_dependency 'backports'
 
   spec.add_development_dependency 'bundler', '~> 1.10'
   spec.add_development_dependency 'rake', '~>10.5'
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -80,3 +80,5 @@ require 'daru/core/merge.rb'
 
 require 'daru/date_time/offsets.rb'
 require 'daru/date_time/index.rb'
+
+require 'backports'


lokeshh:
I think we need minor release for statsample-glm build to pass.

lokeshh:
Give me some time before you make the release. There are three specs failing in Ruby 2.0 even after solving the `to_h`.

v0dro:
Take your time. Release will be first week of next month.

lokeshh:
I just realize there the three specs which were failing in Ruby 2.0, they are also failing in other Ruby versions like 2.2.1. They look something like this:

```ruby

  1) Statsample::GLM::Normal MLE algorithm reports correct values as a Daru::Vector
     Failure/Error: (data_set * coefficients[0..-2,0]).map { |xb| xb }
     
     TypeError:
       no implicit conversion of Range into Integer
     # /home/ubuntu/.rvm/gems/ruby-2.2.1/gems/backports-3.6.8/lib/backports/1.9.2/stdlib/matrix.rb:315:in `fetch'
     # /home/ubuntu/.rvm/gems/ruby-2.2.1/gems/backports-3.6.8/lib/backports/1.9.2/stdlib/matrix.rb:315:in `[]'
     # ./lib/statsample-glm/glm/mle/normal.rb:18:in `measurement'
     # ./lib/statsample-glm/glm/mle/base.rb:28:in `initialize'
     # ./lib/statsample-glm/glm/base.rb:32:in `new'
     # ./lib/statsample-glm/glm/base.rb:32:in `initialize'
     # ./lib/statsample-glm/glm/normal.rb:9:in `initialize'
     # ./lib/statsample-glm/glm.rb:32:in `new'
     # ./lib/statsample-glm/glm.rb:32:in `compute'
     # ./spec/normal_spec.rb:10:in `block (3 levels) in <top (required)>'
```

lokeshh:
I have a fix for this but this would require submitting at PR to statsample I think because 'backports' gem has to be loaded before 'extendmatrix' to solve the issue. See https://github.com/marcandre/backports/issues/101

lokeshh:
Wait there's one more solution which wouldn't involve making changes in Statsample. It would be add backports gem in statsample-glm and loading it before loading statsample. 

I have done this here: https://github.com/lokeshh/statsample-glm/commit/3925baf8efb0dde890177173c6e976d6b5911d4b and now all the tests are passing for all rubies.

v0dro:
backports should be loaded from daru itself since it's being used only in daru and not in statsample-glm.

daru can be the first gem you require in statsample-glm if that's an issue.

v0dro:
Its alright if the Travis builds fail until a daru release happens.

lokeshh:
Makes sense... I'm going to try it out. Thanks

Issue #124, opened 2016-05-12
All encompassing rake task for testing


v0dro:
We should ideally have a rake task that does the following, and automatically generates a report of failures in each case:

- Run rspecs for each ruby version (2.0,2.1,2.2,2.3).
- Run rubocop for each ruby version.
- Generate a report of test failures sorted by Ruby version.

PR #129, opened 2016-05-19
Add rake task to run spec for all Ruby versions

lokeshh:
For #124 

Create a rake task to test the specs in all Ruby versions automatically and display the result.

Tests for all versions are ran using `rake all`.


Rakefile
```diff
@@ -2,6 +2,16 @@ require 'rspec/core/rake_task'
 require 'bundler/gem_tasks'
 
 lib_folder = File.expand_path("../lib", __FILE__)
+
+RUBIES = ['ruby-2.0', 'ruby-2.1', 'ruby-2.2', 'ruby-2.3.0']
+
+task :all do |task|
+  RUBIES.each do |ruby_v|
+    puts "\n\n\nUsing #{ruby_v}\n\n\n"
+    puts `bash -lc "rvm use #{ruby_v}; rspec spec"`
```
v0dro: rspec provides a class for defining rake tasks. I think RVM has it's own Ruby API too. Use that. What if the user is not using bash? Or if he does not have bash?

lokeshh: Point. I will have a look what RVM API has to offer.

lokeshh: Checkout the new solution. It makes use of RVM generated wrappers.

v0dro:
OK so this is one part of the solution. But if there is a failure in Ruby 2.0.0, I will need to scroll all the way up to see that. It should ideally print the test report (number of tests passed and failures) for each ruby version at the bottom of the terminal output, so that in case there are failures, I know where to look instantly.

v0dro:
Btw it should also run rubocop.

lokeshh:
How about just printing the number of passed/failed for each Ruby? The user can just change Ruby version and run rspec if he wants to know about the failures.

v0dro:
Yeah that works too. Make sure you use appropriate colours (red for fail, green for pass).

lokeshh:
Ok. I did like I said but adding colors would need a little more work.

Should the rubocop also run in this same task or should I create another task for it?

v0dro:
You can add the :cop task as a subtask of :all.

lokeshh:
`:cop` doesn't need to run for all rubies right?

lokeshh:
Color support is added. Now `:all` firstly runs `rubocop` then it writes the summary for all rubies.

formatter.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def dump_summary(message)
+    colorizer = ::RSpec::Core::Formatters::ConsoleCodes
+
+    output.puts "\nFinished in #{message.formatted_duration} " \
+       "(files took #{message.formatted_load_time} to load)\n" \
+       "#{message.colorized_totals_line(colorizer)}\n"
+  end
+end
```
lokeshh: Thanks to [patkoperwas](http://stackoverflow.com/users/3295284/patkoperwas) for he made it possible just print the number of pass/failed examples with colors. [Here's](http://stackoverflow.com/questions/37553229/print-only-the-number-of-failed-examples-with-rspec?noredirect=1#comment62596136_37553229) his answer that helped me.

v0dro: How about rename this file to `.rspec_formatter.rb` so that its explicit that it's rspec and it stays hidden too.

lokeshh: Done

lokeshh:
@v0dro I think this is ready to be merged. Just a small thing. Please specify the complete versions names to be tested. Currently I have these: https://github.com/v0dro/daru/pull/129/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR7

PR #132, opened 2016-05-20
Group by

### LINKED ISSUE #120: Difference between single and multiple field selection
So something like this surprised me today.  I converted some code that I had written that used multiple vector references to one that only used one and it broke.  Something like the following.  I understand why this is happening (df[:a,:b] returns a dataframe and df[:a] returns a vector).  Any ideas for how we could clean it up to behave more naturally?

````ruby
df = Daru::DataFrame.new({ a: 1.upto(2).to_a , b: 1.upto(2).to_a })
df_ab = df[:a,:b]
df_a = df[:a]

df_ab.group_by([:a]).size
df_a.group_by([:a]).size # => ArgumentError: wrong number of arguments (1 for 0)
````

### PR DESCRIPTION
**Pull request by lokeshh**
Fixes #120 by creating a custom Daru::Vector::group_by which redirects to Daru::DataFrame where DataFrame is the one containing only that vector.

Also create a function `to_df` to convert a vector to a DataFrame containing only that vector.

### DIFF
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -1936,16 +1936,6 @@ describe Daru::Vector do
     end
   end
 
-  context "#to_df" do
-    it "converts vector to DataFrame containing that vector" do
-      dv = Daru::Vector.new 1..10, name: :a
-      df = Daru::DataFrame.new({
-        a: 1..10
-      })
-      expect(dv.to_df).to eq(df)
-    end
-  end
-
   context "#group_by" do
     it "redirects to DataFrame::group_by" do
       dv = Daru::Vector.new [1, 1, 2, 2, 3], name: :a
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -1936,4 +1936,21 @@ describe Daru::Vector do
     end
   end
 
+  context "#to_df" do
+    it "converts vector to DataFrame containing that vector" do
+      dv = Daru::Vector.new 1..10, name: :a
+      df = Daru::DataFrame.new({
+        a: 1..10
+      })
+      expect(dv.to_df).to eq(df)
+    end
+  end
+
+  context "#group_by" do
+    it "redirects to DataFrame::group_by" do
+      dv = Daru::Vector.new [1, 1, 2, 2, 3], name: :a
+      expect(dv.group_by([:a]).groups).to eq(dv.to_df.group_by([:a]).groups)
+    end
+  end
+
 end if mri?
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -1254,7 +1254,7 @@ module Daru
     end
 
     def group_by *args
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by v0dro** [comment]
Don't use shortcuts in Ruby methods. `to_dataframe` is better. Otherwise looks fine.

File: spec/vector_spec.rb [hunk 1]
```diff
@@ -1169,4 +1169,21 @@ def expect_correct_tokens hash
...
+        a: 1..10
+      })
+      expect(dv.to_df).to eq(df)
+    end
+  end
+
+  context "#group_by" do
+    it "redirects to DataFrame::group_by" do
+      dv = Daru::Vector.new [1, 1, 2, 2, 3], name: :a
+      expect(dv.group_by([:a]).groups).to eq(dv.to_df.group_by([:a]).groups)
```
v0dro: This test should have the actual result that the call to group_by will generate. Here you're just calling what's inside the `Vector#group_by` method.

spec/vector_spec.rb
```diff
@@ -1169,4 +1169,21 @@ def expect_correct_tokens hash
...
+      dv = Daru::Vector.new 1..10, name: :a
+      df = Daru::DataFrame.new({
+        a: 1..10
+      })
+      expect(dv.to_df).to eq(df)
+    end
+  end
+
+  context "#group_by" do
+    it "redirects to DataFrame::group_by" do
```
v0dro: It's an instance method so `DataFrame#group_by`. `::` is for class methods.

spec/vector_spec.rb
v0dro: A better API would be `dv.group_by.groups.

The group_by should of course be capable of accepting all the other arguments that `DataFrame#group_by` is capable of accepting.

v0dro:
Can you fix this PR? The `Vector#group_by` looks useful. `#to_df` has already been added.

lokeshh:
@v0dro Sorry, this wasn't ready. In particular one has to has to do `df[:a].group_by([:a]).groups` instead of just `df[:a].group_by.groups`.

lokeshh:
@gnilrets Which one is better `df[;a].group_by([:a])` or just `df[:a].group_by`. I'm asking this because the former one has the advantage that it is analogous to `df[:a, :b].group_by[:a]` but had the disadvantage that we have to specify `[:a]` when its obvious that its should be `[:a]`.

v0dro:
Reverted. Open a new PR for this one.

gnilrets:
@lokeshh - Perhaps we could support both if `#group_by` without an argument defaulted to the first vector of a dataframe?

lokeshh:
@gnilrets That seems like a brilliant idea. It would evade the need to construct all `group_by` methods for `Daru::Vector` and at the same time providing flexibility to the user. @v0dro What do you say?

v0dro:
Works for me.

PR #135, opened 2016-05-24
Update SqlDataSource to improve performance of DataFrame.from_sql

dansbits:
This updates the SqlDataSource so that it doesn't loop over the SQL result set when building data frames. There is a significant performance boost for DataFrame.from_sql

### DIFF
--- a/lib/daru/io/sql_data_source.rb
+++ b/lib/daru/io/sql_data_source.rb
@@ -9,16 +9,10 @@ module Daru
           @query = query
         end
 
-        def each_column_name
-          result.column_names.each do |column_name|
-            yield(column_name.to_sym)
-          end
-        end
-
-        def each_row
-          result.fetch do |row|
-            yield(row.to_a)
-          end
+        def result_hash
+          columns = result.column_names.map(&:to_sym)
+          data = result.to_a.map(&:to_a).transpose
+          columns.zip(data).to_h
         end
 
         private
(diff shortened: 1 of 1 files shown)


lib/daru/io/sql_data_source.rb
```diff
@@ -9,16 +9,10 @@ def initialize(dbh, query)
...
-        end
-
-        def each_row
-          result.fetch do |row|
-            yield(row.to_a)
-          end
+        def result_hash
+          columns = result.column_names.map(&:to_sym)
+          data = result.to_a.map(&:to_a).transpose
+          Hash[columns.zip(data)]
```
zverok: `columns.zip(data).to_h`, maybe?..

lib/daru/io/sql_data_source.rb
```diff
@@ -36,16 +30,11 @@ def initialize(conn, query)
...
-        end
+        def result_hash
+          columns = result.columns.map(&:to_sym)
+          data = result.cast_values.transpose
 
-        def each_row
-          result.each do |row|
-            yield(row.values)
-          end
+          Hash[columns.zip(data)]
```
zverok: Same as above.

dansbits:
@zverok Thanks for the Hash[] to to_h adjustment. That looks much better. :wink:  

I just pushed an update.

PR #149, opened 2016-06-01
Better `#to_html` for core classes

zverok:
Changes in the same fashion I did for `#inspect`. Mostly cosmetic, but data with multi indexes are now much more readable (for me, and may be only for me :)).

Before:

After:

(Still nothing done about multi-index dataframes using `MultiIndex` as order, though.)

### DIFF
--- a/lib/daru/index.rb
+++ b/lib/daru/index.rb
@@ -364,6 +364,11 @@ module Daru
         Formatters::Table.format([], row_headers: sparse_tuples, threshold: threshold)
     end
 
+    def to_html
+      path = File.expand_path('../iruby/templates/multi_index.html.erb', __FILE__)
+      ERB.new(File.read(path).strip).result(binding)
+    end
+
     # Provide a MultiIndex for sub vector produced
     #
     # @param input_indexes [Array] the input by user to index the vector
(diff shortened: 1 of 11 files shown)


gnilrets:
Like!

lokeshh:
Its amazingly great looking and much more readable from before. 

One thing worth mentioning is that when the index is not in proper order (eg `[[:a, :one], [:a, :two], [:b, :one], [:a, :three]]`) the output is a bit deceiving because in this display form one would expect all the `:a`'s and `:b`'s to lie together though this will not be the case.

lib/daru/index.rb
```diff
@@ -386,5 +391,32 @@ def sparse_tuples
         [nil] * (cur.size - left.size) + left.map(&:first)
       }
     end
+
+    def tuples_with_rowspans
```
v0dro: documentation for this method.

zverok: In fact, I thought about just making it `# @private`-ly documented, it is not that easy to explain what it does and why does somebody need it :)

zverok:
> In this case, the output will be bit deceiving because in this display form one would expect all the `:a`'s and `:b`'s to lie together though this will not be the case.

But that's how Daru stores them internally (if you'll enumerate tuples, you'll have the same result. I'm not aware of how multi-index-es are working in pandas or R, but current `#to_html` behavior just showing "the actual state of things", not trying to mask it.

v0dro:
Yeah that's how it should be. Order needs to be preserved.

PR #134, opened 2016-05-23
Categorical Index and Categorical Data

lokeshh:
The coding begins!

This PR adds `CategoricalIndex` to Daru.

PR #161, opened 2016-06-06
[WIP] Categorical Data

lokeshh:
Add support of categorical data to Daru.

### DIFF
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -1,5 +1,7 @@
 module Daru
   module Category
+    attr_accessor :coding_scheme, :base_category
+
     def initialize_catogory data, opts={}
       @type = :category
 
@@ -20,6 +22,15 @@ module Daru
       # Specify if the categories are ordered or not.
       # By default its unordered
       @order = opts[:order] || false
+
+      # The coding scheme to code with. Default is dummy coding.
+      @coding_scheme = :dummy
+
+      # Base category which won't be present in the coding
+      @base_category = @cat_hash.keys.first
+
+      # Stores the name of the vector
+      @name = opts[:name]
     end
 
     def each
@@ -42,6 +53,35 @@ module Daru
 
     def category
       @cat_hash.keys
+    end
+
+    def contrast_code full=false
+      send("#{coding_scheme}_coding".to_sym, full)
+    end
+
+    private
+
+    def dummy_coding full
+      categories = @cat_hash.keys
+      categories.shift unless full
+
+      df = categories.map do |category|
+        code_to_binary @cat_hash[category]
+      end
+
+      Daru::DataFrame.new df,
+        index: @index,
+        order: create_names(categories)
+    end
+
+    def code_to_binary positions
+      binary = Array.new(size, 0)
+      positions.map { |pos| binary[pos] = 1 }
+      binary
+    end
+
+    def create_names categories
+      categories.map { |cat| "#{name}_#{cat}" }
     end
   end
 end
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -78,6 +78,12 @@ module Daru
         order: create_names(categories)
     end
 
+    def dummy_code positions
+      code = Array.new(size, 0)
+      positions.each { |pos| code[pos] = 1 }
+      code
+    end
+
     def simple_coding full
       categories = @cat_hash.keys
       categories.delete(base_category) unless full
@@ -98,10 +104,55 @@ module Daru
       code
     end
 
-    def dummy_code positions
-      code = Array.new(size, 0)
-      positions.each { |pos| code[pos] = 1 }
-      code
+    def helmert_coding(*)
+      categories = @cat_hash.keys[0..-2]
+
+      df = categories.each_index.map do |index|
+        helmert_code index
+      end
+
+      Daru::DataFrame.new df,
+        index: @index,
+        order: create_names(categories)
+    end
+
+    def helmert_code index
+      n = (categories.size - index).to_f
+
+      @array.map do |cat_index|
+        if cat_index == index
+          (n-1)/n
+        elsif cat_index > index
+          -1/n
+        else
+          0
+        end
+      end
+    end
+
+    def deviation_coding(*)
+      categories = @cat_hash.keys[0..-2]
+
+      df = categories.each_index.map do |index|
+        deviation_code index
+      end
+
+      Daru::DataFrame.new df,
+        index: @index,
+        order: create_names(categories)
+    end
+
+    def deviation_code index
+      last = categories.size - 1
+      @array.map do |cat_index|
+        if cat_index == index
+          1
+        elsif cat_index == last
+          -1
+        else
+          0
+        end
+      end
     end
 
     def create_names categories
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -59,11 +59,59 @@ module Daru
       @cat_hash.keys
     end
 
+    def categories= new
+      # Change to SizeError
+      if new.size != categories.size
+        raise ShapeError, "New categories size (#{new.size}) does not match"\
+          "with old categories size (#{categories.size})"
+      end
+
+      @cat_hash = new.each_with_index.map do |new_cat, i|
+        @map_int_cat[i] = new_cat
+
+        old_cat = categories[i]
+        [new_cat, @cat_hash[old_cat]]
+      end.to_h
+    end
+
+    def order= new
+      if new.to_set != categories.to_set
+        raise ArgumentError, 'The context of new and old order must be the same.'
+      end
+
+      @cat_hash = new.map { |cat| [cat, @cat_hash[cat]] }.to_h
+    end
+
+    def min
+      assert_ordered 'min'
+
+      categories.first
+    end
+
+    def max
+      assert_ordered 'max'
+
+      categories.last
+    end
+
+    def sort
+      assert_ordered 'sort'
+
+      
+    end
+
     def contrast_code full=false
       send("#{coding_scheme}_coding".to_sym, full)
     end
 
     private
+
+    def assert_ordered operation
+      if !ordered?
+        # Change ArgumentError to something more expressive
+        raise ArgumentError, "Can not apply #{operation} when vector is unordered"
+      end      
+    end
 
     def dummy_coding full
       categories = @cat_hash.keys
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -80,7 +80,7 @@ module Daru
 
     def order= new
       if new.to_set != categories.to_set
-        raise ArgumentError, 'The context of new and old order must be the same.'
+        raise ArgumentError, 'The contents of new and old order must be the same.'
       end
 
       @cat_hash = new.map { |cat| [cat, @cat_hash[cat]] }.to_h
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -2,7 +2,7 @@ module Daru
   module Category
     attr_accessor :coding_scheme, :base_category
 
-    def initialize_catogory data, opts={}
+    def initialize_category data, opts={}
       @type = :category
 
       # Create a hash to map each category to positional indexes
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -51,7 +51,7 @@ module Daru
       @order
     end
 
-    def category
+    def categories
       @cat_hash.keys
     end
 
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -98,10 +98,26 @@ module Daru
       categories.last
     end
 
-    def sort
+    def sort!
       assert_ordered 'sort'
 
-      # TODO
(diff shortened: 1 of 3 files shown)


lokeshh:
Need some guidance here. Should I create a separate spec file for it or not?

v0dro:
Separate file is fine.

v0dro:
Isnt this PR supposed to depend on the previous one? As I said master should remain stable and up to date so we won't merge categorical index until all of categorical data is finished.

spec/categorical_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    subject { dv }
+    
+    it { is_expected.to be_a Daru::Vector }
+    its(:size) { is_expected.to eq 5 }
+    its(:type) { is_expected.to eq :category }
+    its(:ordered?) { is_expected.to eq false }
+    its(:to_a) { is_expected.to eq [:a, 1, :a, 1, :c] }
+  end
+  
+  context "#to_factor!" do
```
zverok: I know some statistic software like R uses this as a synonym of categories, but is it transparent for everyone?.. Like, you have two absolutely different words for the same thing....

lib/daru/vector.rb
```diff
@@ -170,21 +171,26 @@ def map!(&block)
     #   vecarr = Daru::Vector.new [1,2,3,4], index: [:a, :e, :i, :o]
     #   vechsh = Daru::Vector.new({a: 1, e: 2, i: 3, o: 4})
     def initialize source, opts={}
-      index, source = parse_source(source, opts)
-      set_name opts[:name]
+      if opts[:type] == :category
+        extend Daru::Category
```
zverok: For me, this part of code and overall logic looks like we just want a `CategoryVector` class, not some playing with module-inclusion-on-the-fly.

lokeshh: @v0dro what do you think?

spec/categorical_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+  
+  context "#contrast_code" do
+    let(:dv) { Daru::Vector.new [:a, 1, :a, 1, :c], type: :category, name: 'abc' }
+    subject { dv.contrast_code }
+    
+    it { is_expected.to be_a Daru::DataFrame }
+    its(:shape) { is_expected.to eq [5, 2] }
+    it { expect(subject['abc_1'].to_a).to eq [0, 1, 0, 1, 0] }
+    it { expect(subject['abc_c'].to_a).to eq [0, 0, 0, 0, 1] }
```
lokeshh: @zverok Any way to improve this line? The `its(:'abc_1.to_a')` doesn't work. I think its a bug.

zverok: Hm, as dataframes should respond to any `<vector_name>` as to method, I assume `its(:'abc_1.to_a')` should work. What does it say?

lokeshh:
```ruby
[2] pry(main)> df = dv.contrast_code
=> #<Daru::DataFrame(5x2)>
       abc_1 abc_c
     0     0     0
     1     1     0
     2     0     0
     3     1     0
     4     0     1
[3] pry(main)> df.abc_1
NoMethodError: undefined method `abc_1' for #<Daru::DataFrame:0x00000003729cc0>
from /home/ubuntu/workspace/daru/lib/daru/dataframe.rb:1644:in `method_missing'
```

zverok: Hm, I can suppose that the problem is here:
```ruby
df = Daru::DataFrame.new(a: [1,2,3], b: [4,5,6], c: [7,8,9])
df.a # => ok!
# => #<Daru::Vector(3)>
#       a
#   0   1
#   1   2
#   2   3 

df.respond_to?(:a) # not ok!!!
# => false 
```
rspec-its looks at `respond_to?` method, and in this case receives not the answer it wants.

zverok: Heh. That's another problem:
```ruby
df = Daru::DataFrame.new(a: [1,2], 'b' => [3,4])
df.a
# => #<Daru::Vector(2)>
#       a
#   0   1
#   1   2 
df.b
# NoMethodError: undefined method `b' for #<Daru::DataFrame(2x2)>
```

So, the `method_missing` only responds to Symbol vector names.

lokeshh: Thanks for looking into this. I think there's some more problem because I changed `contrast_code` function so that now the vector names are symbols but found out that its still not working.

lokeshh: > rspec-its looks at respond_to? method, and in this case receives not the answer it wants.

This doesn't seem to be the case because `its(:a)` works.

zverok: Hm, I've just pulled your branch, and using 
```ruby
its(:'abc_1.to_a') { is_expected.to eq [0, 1, 0, 1, 0] }
```
works perfectly well.

And yes, I was wrong about `respond_to?`, apparently (though, one day we should fix it, too, as well as a string/symbol confusion).

lokeshh: Oh yea, its working. Thanks. I was mistaken about it.

spec/categorical_spec.rb
```diff
@@ -0,0 +1,102 @@
+describe Daru::Vector do
+  context "initialize" do
+    let(:dv) { Daru::Vector.new [:a, 1, :a, 1, :c], type: :category }
+    subject { dv }
+    
+    it { is_expected.to be_a Daru::Vector }
+    its(:size) { is_expected.to eq 5 }
+    its(:type) { is_expected.to eq :category }
```
agisga: Maybe the type should be changed to `:categorical`? I think the term "categorical data" is a lot more common than "category data". Also the other possibility for type is `:numeric` and not `:number`, right?

lib/daru/category.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def size
+      @array.size
+    end
+
+    def ordered?
+      @order
+    end
+
+    def category
```
agisga: Maybe this should be called `categories` in plural? Because it returns an array of all available categories rather than just one category.

lib/daru/category.rb
```diff
@@ -0,0 +1,107 @@
...
+      # Inverse mapping of map_cat_int
+      @map_int_cat = map_cat_int.invert
+
+      # To link every instance to its category,
+      # it stores integer for every instance representing its category
+      @array = map_cat_int.values_at(*data)
+
+      # Specify if the categories are ordered or not.
+      # By default its unordered
+      @order = opts[:order] || false
```
agisga: Do you think `@order` should be reflected in `categories`, `@cat_hash`, `@base_category` (for example, the category which has order 1 should maybe be the default for `@base_category`)?

lib/daru/vector.rb
```diff
@@ -170,21 +171,26 @@ def map!(&block)
...
+        initialize_catogory source, opts
```
agisga: typo here (`initialize_category`) :)

lib/daru/category.rb
lokeshh: Yes, its better. Thanks

spec/categorical_spec.rb
lokeshh: Ok. Just want to confirm from @v0dro about this because he said once:

> I know I said :categorical on the listserv, but now that I think of it, we should be consistent with other tools and use :category.

lib/daru/category.rb
lokeshh: Um I didn't get what you said. `categories`, `@cat_hash.keys` all store the categories in the same sequence and is in the order in which they are encountered first in the vector. And `@base_category` chooses the first value of `@cat_hash.keys` so it is the one which occurs first in the vector.

lokeshh: And I will add a support such that one can reorder the categories in case the `@order = true` i.e. the categories have a order.

spec/categorical_spec.rb
agisga: Oh, if other tools use `category`, then it makes sense to stick to it here of course.

I see that pandas indeed uses dtype `category` (though their [help page](http://pandas.pydata.org/pandas-docs/stable/categorical.html) is titled "categorical"). Does type in daru play the same role as dtype in pandas?

lib/daru/category.rb
agisga: Oh never mind then. I didn't realize that `@order` is boolean. I though that `@order` is an array of category names put in a desired order.

zverok: Maybe it should be named `@ordered` to avoid such a confusion?

lokeshh: Yep. Done.

lib/daru/vector.rb
v0dro: If we do a `CategoryVector` class it will mean possibly further confusing users since they'll get a new data container just by specifying the index as a `CategoricalIndex`.

Keeping it as `Daru::Vector` and ensuring that the behavior remains similar for both categorical and normal data will make it easier to use and wrap your head around in general.

spec/categorical_spec.rb
v0dro: @lokeshh it's `:numeric` not `:number`.

I think pandas dtype reflects the numpy dtype that the data is stored as. daru has `:numeric` and `:object` only so that math and statistics methods can figure out if a computation is feasible on data.

I didn't want to reflect dtype as `int` or `float` since we use `Array` internally.

lib/daru/vector.rb
zverok: OK, let's leave it this way for now. For me, it looks like serious code smell curently, but in fact we can refactor it later as many times as we want to :)

spec/categorical_spec.rb
agisga: My point regarding "`:numeric` not `:number`" was that `:categorical` would match the existing `:numeric` better, because both are adjectives. I think that maybe a reason why pandas devs have chosen dtype `category` in favor of `categorical` is that all the other dtypes are nouns, like `float`, `int` or `object` (of course, I have no idea of the true reasons the pandas devs had).

But I'm fine with it, if we keep using `:category` and not `:categorical`, since pandas is using it. Though I think that in the english language describing data as being of "categorical type" is more correct than "category type".

v0dro: I think we'll shift to using nmatrix internally once it's compatible and stable with jruby. So it makes sense to use `:category`.

lib/daru/category.rb
```diff
@@ -0,0 +1,235 @@
...
+
+    def order= new
+      if new.to_set != categories.to_set
+        raise ArgumentError, 'The context of new and old order must be the same.'
+      end
+
+      @cat_hash = new.map { |cat| [cat, @cat_hash[cat]] }.to_h
+    end
+
+    def min
```
agisga: @lokeshh Not sure if you planned it already, but I think that if we have `#min` and `#max`, then we should also have comparison operators < and >. What do you think?

spec/categorical_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+    before { dv.categories = [1, 2, 3] }
+    
+    its(:to_a) { is_expected.to eq [1, 2, 1, 2, 3] }
+  end
+  
+  context "#order=" do
+    context "valid reordering" do
+      let(:dv) { Daru::Vector.new [:a, 1, :a, 1, :c], type: :category }
+      subject { dv }
+      before { dv.order = [:c, 1, :a] }
```
agisga: Is there a reason not to provide `order` as an option when the vector is initialized? I think this two step procedure of generating an ordered vector is slightly inconvenient.

lib/daru/category.rb
lokeshh: I'm thinking about extending the [Arel like query syntax](http://v0dro.github.io/blog/2015/08/03/finding-and-combining-data-in-daru/) which is already in Daru for that purpose.

agisga: Nice! :)

spec/categorical_spec.rb
lokeshh: Yes, I'm planning to introduce this feature of providing order while initializing the vector. I was a little confused on whether the option should be named `order` or `categories`. Also I was thinking that during initialization the user can specify categories which is not inside the vector already so he can have result like there are 0 instances of such and such category.

lokeshh: The cause of confusion is that I want both the things that is user specifying all categories and user specifying the order should be accomplished with the same option, so I'm not sure to name it `order` or `categories`.

agisga: If I understand you plan correctly, I think that `categories` is better, because the user might want to specify which categories are available (even if some have 0 instances), without making them ordered. Then the user could set `categories` to something and set `ordered: false`. On the other hand it would be counterintuitive to set `order` to something, but set `ordered: false` at the same time.

lokeshh: Ok then I will use `categories` to input the categories and their order. If `ordered` is false, then I will consider the input to `categories` just categories and if `ordered` is set to true, then I will consider the input to `categories` as well as their order.

lokeshh:
Moved to #134

PR #163, opened 2016-06-07
add rspec-guard to run specs automatically

### LINKED ISSUE #157: Use Guard for automatically running tests.
It would be great have something like [guard-spec](https://github.com/guard/guard-rspec) to automatically run the concerned specs whenever a change is made. This would **really** help to make TDD better.

lokeshh:
Addresses #157


lokeshh:
Need some help here to edit the `GuardFile`.

lokeshh:
Travis build is failing because `guard-rspec` doesn't work with old rubies.

v0dro:
Alright, but I cannot merge this in master unless we drop support for older rubies.

lokeshh:
Ok. Do you know a way to add dependency only if Ruby version is 2.3.0?

v0dro:
Alright. But it's a little hairy when using bundler for this. See my comments in #129.

v0dro:
Can you update this with the necessary changes?

lokeshh:
Sure. I've updated the requirement for `guard` so it will now only install when ruby version is >= '2.2.5'. Now all what remains is configuring the Guardfile.

lokeshh:
I've modified the Guardfile so that now if there's a change in say `lib/daru/abc.rb` it will run the tests automatically in `spec/abc_spec.rb`.

daru.gemspec
```diff
@@ -74,6 +74,8 @@ EOF
   spec.add_development_dependency 'ruby-prof'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'nokogiri'
-  spec.add_development_dependency 'guard-rspec'
   spec.add_development_dependency 'gruff'
+  if RUBY_VERSION >= '2.2.5'
```
v0dro: I'm not sure if this is valid for gemspecs. Have a look at the bundler docs for conditional dependencies. I vaguely remember running into problems once when I had done something similar.

lokeshh: I get its reference from [here](https://www.tiredpixel.com/2014/01/05/curses-conditional-ruby-gem-installation-within-a-gemspec/) (See at the end). And its working.

lokeshh: I get its reference from [here](https://www.tiredpixel.com/2014/01/05/curses-conditional-ruby-gem-installation-within-a-gemspec/) (See at the end). And its working.

lokeshh:
I've updated the CONTRIBUTING.md regarding automating test with guard. I had to split some spec files in order for guard to trigger right tests. This is ready to be merged.

Please see if you can merge this early as this would enable fixing other issues elegantly and rebasing with such huge PR sometimes confuses me.

PR #165, opened 2016-06-09
Remove monkey-patching

zverok:
Monkey-patching of core Ruby classes is a questionable practice that could be argued about.
But monkey-patching in gems for gem's internal needs (which, nevertheless, pollutes all the outside code) is definitely bad. So, let's get rid of this.

**Notes**
What was left in `monkeys.rb`:
* ability to convert core classes to `Daru::{Vector,DataFrame}` (though, I feel like it should be in optional `require 'daru/shortcuts'` or `daru/extensions`)
* `Object#itself` (it is introduced in Ruby 2.2, but somehow not in backports)
* `Matrix#elementwise_division` (I am afraid to break something by moving it)

The most painful was moving of `Array`s extensions (`recode_repeated` and `all_are?`), the code became not that pretty as it was... But we just should do it, anyways.

### DIFF
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -69,10 +69,12 @@ autoload :Spreadsheet, 'spreadsheet'
 autoload :CSV, 'csv'
 
 require 'matrix'
-require 'securerandom'
+#require 'securerandom'
 require 'reportbuilder'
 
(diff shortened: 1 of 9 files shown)


v0dro:
OK looks good.

PR #181, opened 2016-07-06
Remove runtime dependencies spreadsheet and reportbuilder 

### LINKED ISSUE #139: [Cleanup] Drop unnecessary runtime dependencies
There are two ugly fat dependencies which currently prohibit usage of daru in small-scale projects:
* reportbuilder (which tugs rtf, prawn, and whatnot);
* spreedsheet (which tugs ruby-ole, and is fat enough on its own).

Both of them are absolutely not necessary for core daru functionality, so they should be documented as optional dependencies (like nyaplot or gnuplot) and not go into mandatory ones.

v0dro:
Fixes #139

### DIFF
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -69,7 +69,7 @@ end
   begin
     gem lib[0], lib[1]
     require lib[0]
-  rescue LoadError => e
+  rescue LoadError
     STDERR.puts "\nInstall the #{lib[0]} gem version #{lib[1]} for using"\
     " #{lib[0]} functions."
   end
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -65,11 +65,18 @@ module Daru
   create_has_library :nyaplot
 end
 
-autoload :Spreadsheet, 'spreadsheet'
-autoload :CSV, 'csv'
+[['reportbuilder', '~>1.4'], ['spreadsheet', '~>1.1.1']].each do |lib|
+  begin
+    gem lib[0], lib[1]
+    require lib[0]
+  rescue LoadError => e
+    STDERR.puts "\nInstall the #{lib[0]} gem version #{lib[1]} for using"\
+    " #{lib[0]} functions."
+  end
+end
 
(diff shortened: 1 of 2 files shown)


lib/daru.rb
```diff
@@ -65,11 +65,18 @@ def create_has_library(library)
   create_has_library :nyaplot
 end
 
-autoload :Spreadsheet, 'spreadsheet'
-autoload :CSV, 'csv'
+[['reportbuilder', '~>1.4'], ['spreadsheet', '~>1.1.1']].each do |lib|
```
zverok: BTW, you can write `each do |lib, version|` in such cases.

v0dro: Ah. Noted.

lib/daru.rb
```diff
@@ -65,11 +65,18 @@ def create_has_library(library)
...
+  begin
+    gem lib[0], lib[1]
+    require lib[0]
+  rescue LoadError
+    STDERR.puts "\nInstall the #{lib[0]} gem version #{lib[1]} for using"\
```
PikachuEXE: Can we only see this message if we want to use those features?
I now see it on every start up which is a bit annoying

Also RubyMine consider anything in STDERR as warning, not just info

v0dro: That's a good idea, but will involve shifting the `require` to when the method is being called, which can increase the time required to execute the method.

However, I think we can achieve something like what you're saying if daru can be modularized effectively as per this proposal: https://github.com/SciRuby/daru/pull/304

PikachuEXE: I will try to come up with a PR then :)
Something like `Daru::Extensions::Reporting`?

v0dro: Hmm so there is an ongoing issue for removing the reportbuilder dependency for good: https://github.com/SciRuby/daru/issues/221

Also, the spreadsheet gem is used for the `from_excel` and `write_excel` methods. They won't really fit into something like `Reporting` since they are more of IO functionality for dataframes.

PR #194, opened 2016-07-25
Add links to categorical data notebooks

lokeshh:


v0dro:
rubocop failures.

lokeshh:
Should I submit a PR to solve Rubocop offences? The failure seems to me as a result of Rubocop update. It wasn't raising those offences earlier.

v0dro:
That's weird. The [gem version of rubocop is locked](https://github.com/v0dro/daru/blob/master/daru.gemspec#L73).

zverok:
> That's weird. The gem version of rubocop is locked.

It is an optimistic lock (`>=`). I like to have most of gems locked this way to stay on-the-edge. Typically, for gems it is more sane approach than pessimistic (`~>`), which are used for Enterprise Rails Apps.

Issue #187, opened 2016-07-20
Get vector using instance method when name is string


lokeshh:
```ruby
> df = Daru::DataFrame.new(
>   :a => [1, 2, 3],
>   'b' => [4, 5, 6]
> )
> df.a
#=> #works
> df.b
#=> #doesn't work
```
I propose to let let vectors with string names similarly processed. WDYT?

gnilrets:
I can think of at least one edge case which would be weird:
````ruby
df = Daru::DataFrame.new(
:a => [1,2,3],
'a' => [4,5,6]
)
df.a #=> ????
````

Not sure about everyone else, but I strongly prefer the array accessor method getting vectors - `df[:a]`.

**Comment by gnilrets** [comment]
Something that's always bothered me about this is that when you use `df[:a]` you get a vector, but when you use `df[:a, :b]` you get a dataframe.  This could be a problem if I'm creating a new dataframe from a variable list of vectors and it's possible for that list to just have one element.  What would everyone think about using `df[:a]` to return a dataframe and `df.a` to return a vector?

**Comment by zverok** [comment]
This thing is something I've wanted to discuss a long time ago :) Considering not-so-easy Ruby's String/Symbol dychotomy, it could be (maybe) reasonable for all real-life cases to make them indistinguishable from internal point of view (like infamous `HashWithIndifferentAccess`).

Most of the times, when somebody has `:a` and `'a'` vector in the same dataset, it does NOT what he/she wanted, but rather the result of some logical error.

**Comment by zverok** [comment]
> Something that's always bothered me about this is that when you use `df[:a]` you get a vector, but when you use `df[:a, :b]` you get a dataframe. 

This thing is normal for Ruby:
```ruby
arr = ('a'..'e').to_a
# => ["a", "b", "c", "d", "e"] 
arr[3]
# => "d" 
arr[3,4]
# => ["d", "e"] 

```

lokeshh:
> Something that's always bothered me about this is that when you use df[:a] you get a vector, but when you use df[:a, :b] you get a dataframe. This could be a problem if I'm creating a new dataframe from a variable list of vectors and it's possible for that list to just have one element.

I've encountered the same problem. You can use `df[:a..:a]` if you want a dataframe and not a vector. Generalizing it you can do something like following:
```ruby
# Assume x contains list of vectors names
x.size == 1 ? df[x.first..x.first] : df[*x] # This will return a dataframe even when x has one element.
```
I think we can also create another method for this type of usage.

[...]

PR #195, opened 2016-07-27
solve rubocop offences occuring on bundle update

lokeshh:
After doing bundle update or bundle install on a new system (with no daru) one is getting rubocop failures on master branch. This PR solves those issues.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1822,7 +1822,7 @@ module Daru
     end
 
     def respond_to_missing?(name, include_private=false)
-      (name =~ /(.+)\=/) || has_vector?(name) || super
+      name.to_s.end_with?('=') || has_vector?(name) || super
     end
 
     def interact_code vector_names, full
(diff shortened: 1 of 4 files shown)


lokeshh:
There are two rubocop offences still unresolved. That's why this PR is in WIP. Resolving them now.

lokeshh:
Solved the last two offences but there's a [this](https://gist.github.com/lokeshh/699f98f36814b339cc9ec68ed3b62362) example failing.

lokeshh:
It's a result of calling `MultiIndex#include?` with a non array object. So, I've added a check in the method to resolve this.

lib/daru/dataframe.rb
```diff
@@ -1817,10 +1817,14 @@ def method_missing(name, *args, &block)
       elsif has_vector? name
         self[name]
       else
-        super(name, *args, &block)
+        super
```
v0dro: why dont you need to pass the arguments anymore?

zverok: @v0dro `super` without arguments is called implicitly with _exactly_ the arguments of current method

v0dro:
Yes. #187 can be addressed in a separate PR.

Issue #197, opened 2016-07-28
Get rid of vector metadata


gnilrets:
This was my half-baked idea.  It didn't work out quite the way I wanted it to in the project I thought I needed it for so I'm not using it.  Probably best to get rid of it to remove bloat and let users handle metadata on their own.  Unless there are any objections, I'll try to find some time to remove it.

zverok:
+1 for this. This was one of my first question when I've started to work with Daru.

v0dro:
Ok works. Lets remove it.

PR #201, opened 2016-07-29
197 - Removed all references to metadata

gnilrets:
... in the interests of simplifying the API.  This was just bloat.

### DIFF
--- a/lib/daru/core/group_by.rb
+++ b/lib/daru/core/group_by.rb
@@ -218,7 +218,7 @@ module Daru
       #   })
       #   df.group_by([:a]).reduce('') { |result, row| result += row[:c]; result }
       #   # =>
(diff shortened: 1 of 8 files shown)


gnilrets:
Resolves #197 

I think I spent almost a whole day implementing this.  20 minutes to purge.

lib/daru/core/group_by.rb
```diff
@@ -218,7 +218,7 @@ def get_group group
       #   })
       #   df.group_by([:a]).reduce('') { |result, row| result += row[:c]; result }
       #   # =>
-      #   # #<Daru::Vector:70343147159900 @name = nil @metadata = {} @size = 2 >
+      #   # #<Daru::Vector:70343147159900 @name = nil @size = 2 >
```
zverok: Hm, looks like this example of Vector's inspect is very outdated.

PR #202, opened 2016-07-30
Add a #to_df method to vectors and dataframes

gnilrets:
See #187

### DIFF
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -911,6 +911,27 @@ describe Daru::Vector do
         end
       end
 
+      context "#to_df" do
+        let(:dv) { Daru::Vector.new(['a','b','c'], name: :my_dv, index: ['alpha', 'beta', 'gamma']) }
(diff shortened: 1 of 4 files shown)


spec/vector_spec.rb
```diff
@@ -921,6 +921,18 @@
         end
       end
 
+      context "#to_df" do
+        let(:dv) { Daru::Vector.new(['a','b','c'], name: :my_dv, index: ['alpha', 'beta', 'gamma']) }
+        let(:df) { dv.to_df }
+        it 'converts the vector to a single-vector dataframe' do
+          expect(df[:my_dv]).to eq dv
+        end
```
zverok: Two considerations:
* in fact, you never check df is really DataFrame instance, not something other
* what about resulting dataframe name?..

v0dro:
Resolve build failure and @zverok 's comments and I'll merge.

gnilrets:
Thanks, this should resolve those issues.

Issue #137, opened 2016-05-27
Proposal: get rid of missing_values concept completely


zverok:
**Reasons**:

1. `Vector#set_missing_positions` is needed on any vector update and is slow (entire `Daru.lazy_update` trick exists to fight it);
2. The concept is pretty confusing (and, to be honest, I'm not sure even inside Daru always used correctly -- there are ton of simple `nil`-checks here and there, and eventual "right" missing-data checks);
3. Ruby has native `nil`, which has the exact "there is no value" meaning.

As far as I can understand, the concept of "user decides what is 'missing values'" is borrowed from environments where "no values" can be represented by strings like 'None' or 'Nothing', or 'Unknown' etc. I'm pretty sure that the only place in Daru such data should be supported is input and output. So, when fetching data from CSV with `"None"`s, we can provide some option (like `missing: "None"`) -- and they would be stored internally as `nil`s. And, when saving to CSV, we provide same option -- and again, `nil`s are converted to `"None"`s or whatever user wants.

This change will help to greatly simplify API, codebase, docs, and Daru user's mental model.

v0dro:
@gnilrets @mrkn any comments here? I've come across some use cases where missing data was something apart from just `nil`.

v0dro:
Btw I just remembered that some functionality in statsample depends on this particular feature. If we decide to support only `nil` as missing (and even maybe `Float::NAN`) we'll need changes in both daru and statsample.

v0dro:
This is how statsample looked before dataset was replaced with Daru::DataFrame - https://github.com/SciRuby/statsample/blob/63280a31849bb6320eecda04445cc43d34932ae8/lib/statsample/dataset.rb

zverok:
I've just `grep`ed through statsample code, and it is not much of `missing_<anything>` features usage. I assume it should not be hard to update, and we all will benefit from this... Though, I'm sometimes feeling like being too pushing in SciRuby matters and need to slow myself a little :)

mrkn:
I agree with that 'replacing missing values with nil or Float::NAN.

mrkn:
I think the concept of `set_missing_positions` can be divided to scanning whole dataframe and updating  index of missing values.
The current implementation does only scanning.
But updating index is enough for the almost point at which `set_missing_positions` is called.

v0dro:
OK as per what @zverok is saying, we will drop support for missing values apart from nil and Float::NAN in future versions.

lokeshh:
[...]

PR #208, opened 2016-08-03
Missing values

lokeshh:
Addresses #137


spec/vector_spec.rb
```diff
@@ -1641,11 +1697,11 @@ def expect_correct_tokens hash
...
-  context "#n_valid" do
-    it "returns number of non-missing positions" do
-      v = Daru::Vector.new [1,2,3,4,nil,nil,3,5]
-      expect(v.n_valid).to eq(6)
-    end
+  context '#count_values' do
+    let(:dv) { Daru::Vector.new [1, 2, 3, 1, 2, nil, nil] }
+    it { expect(dv.count_values 1, 2).to eq 4 }
+    it { expect(dv.count_values nil).to 2 }
+    it { expect(dv.count_values 3, Float::NAN).to 1 }
```
lokeshh: Is this sort of testing fine, i.e. writing multiple tests in the same context?

zverok: Absolutely.

lokeshh:
I'm not clear about is whether we are removing `set_missing_positions` or not? If we remove it then its not possible to do the caching for `nil` and `Float::NAN` and if we do not remove it then, how will we get rid of lazy update which makes Daru slow.

v0dro:
Let `set_missing_positions` remain for now. This will also involve keeping lazy update, but I really don't see a fast way of updating the missing values cache for now. We'll probably need to port it to C or something funky like that.

lokeshh:
We can also do something as follows:

We will store the missing values positions in two variables, one for `nil` and one for `Float::NAN`. They will store the positions of the missing data. Whenever an update is made instead of calling `set_missing_positions` we will set `@missing_positions_outdated = false`. And whenever in the future we require operation regarding missing value we will check `@missing_positions_outdated`, if its `false`, we need not recompute the missing positions and if its `true` we need to recompute the missing positions and update the appropriate variables.

WDYT about this approach?

v0dro:
> We will store the missing values positions in two variables, one for nil and one for Float::NAN.

Why do `nil` and `Float::NAN` need to be in separate variables?

> Whenever an update is made instead of calling set_missing_positions we will set @missing_positions_outdated = false.

Update to what? The vector or the missing positions cache?

> And whenever in the future we require operation regarding missing value we will check @missing_positions_outdated, if its false, we need not recompute the missing positions and if its true we need to recompute the missing positions and update the appropriate variables.

Works! Make sure  you reflect this everywhere in the code, though.

lokeshh:
> Why do nil and Float::NAN need to be in separate variables?

User might ask for `reject_values(nil)` or `reject_values(Float::NAN).

> Update to what? The vector or the missing positions cache?
I meant update to vector.

lokeshh:
I'm deprecating the functions `has_missing_data?` etc., so they will be working but I'm getting rid of their functionality of working with arbitrary assigned missing values like `1`, `2`, etc. which are other than `nil`, `Float::NAN`. In other words `missing_values = [1, 2]` would have no effect. This makes it easy for me to remove the `lazy_update` and `set_missing_positions`. `lazy_update` and `set_missing_positions` have to removed to see the expected speed up. Let me know if its fine. Otherwise keeping the old functionality as well as introducing new one at the same time is getting hard.

lokeshh:
Kindly have a look. I've made some radical changes like removed some methods like `exist?`, `missing_values=`, etc. to remove the `lazy_update`, `set_missing_positions` to speed up Daru. Some common methods like `has_missing_data?`, `only_valid` has been kept though. It might cause inconvenience for the next Daru release for methods like `missing_values` won't be working but I insist keeping them as they will lead to better performance. Now I will carry on caching the missing positions of `nil` and `Float::NAN`.

lokeshh:
Here are some benchmarks. The performance speed are staggering.

I ran below code for this branch and master branch:

```ruby
require 'benchmark'

n = 1000
dv = Daru::Vector.new 1..100000

Benchmark.bm do |x|
  x.report { 1000.times { dv[0, 10, 20, 30] = nil } }
end
```

Result for current branch:
```ruby
   0.010000   0.000000   0.010000 (  0.004940)
```

Result for master branch:
```ruby
  36.160000   0.260000  36.420000 ( 39.420752)
```

zverok:
> Here are some benchmarks. The performance gains are staggering.

Fascinating! It's a giant leap forward. Now it looks like we can even think about replacing `where` with normal `select` :)

Bravo, @lokeshh!

lokeshh:
Thanks @zverok.

I have been running benchmarks on implementation with caching of `nil`'s positions and implementation with no caching of `nil`s and surprisingly I'm getting more time with caching. I couldn't figure out whats going on. The code seems correctly. Anyway is there some gem to profile which lines in the method are taking a long time and which not?

lokeshh:
It seems to me that its because operations such as `#at`, `#[]` are more expensive such that finding missing values in a vector is almost free, therefore caching of `nil` and `Float::NAN` is not helping.

v0dro:
Hmmm it's interesting to see that caching isn't really helping much. Can you try with profiling and see what's actually going on? I think you can ruby-prof for this.

zverok:
> I'm looking at this in more detail. Is there some gem to profile which lines in the method are taking a long time and which not?

Yeah, like @v0dro says, ruby-prof is pretty easy to use. You just do something like this:
```ruby
RubyProf.start
(code you want to profile)
result = RubyProf.stop
printer = RubyProf::GraphHtmlPrinter.new(result)
printer.print(File.open('profile.html', 'w'))
```

And then that file would have pretty list of what've taken how many time, cross-linked to other methods it uses and stuff.

lokeshh:
@zverok In our case, the behavior is not exactly not the same, so I guess I will go with writing different specs. Right?

zverok:
@lokeshh Yep, while extracting the same behaviors into shared example groups.

lokeshh:
> 

@lokeshh Yep, while extracting the same behaviors into shared example specs and then using `include_context`?


@zverok Do you extracting the same behaviors by using `include_context`?

zverok:
Yep.

lokeshh:
Alright. Thanks!

spec/vector_spec.rb
```diff
@@ -1392,11 +1394,14 @@
...
-      it { is_expected.to be_a Daru::Vector }
-      its(:to_a) { is_expected.to eq [1, 3, :a, Float::NAN, Float::NAN, 1] }
-      its(:'index.to_a') { is_expected.to eq [11, 13, 14, 15, 17, 18] }
+      # it { is_expected.to be_a Daru::Vector }
+      # its(:to_a) { is_expected.to eq [1, 3, :a, Float::NAN, Float::NAN, 1] }
+      # its(:'index.to_a') { is_expected.to eq [11, 13, 14, 15, 17, 18] }
+      include_context 'reject values checker',
+        [dv, [nil]] => [[1, 3, :a, Float::NAN, Float::NAN, 1],
+          [11, 13, 14, 15, 17, 18]]
     end
```
lokeshh: @zverok This is driving me nuts. I'm trying to make this work but it isn't. It gives error something like `'dv' is not available on an example group`. Could you please have a look?

Another thing I'm thinking is that this code doesn't look readable at all. Passing the expected index, along with expected values of the vector and also passing along the type makes it hard to read what's going on. What do you think?

lokeshh:
The PR is complete. I will start adding documentation. @zverok I will take care of extracting the common behavior later on.

Here's the short summary:

Methods added in Daru::Vector (and category):
- reject_values
- include_values?
- indexes
- count_values
- replace values

Methods added in Daru::DataFrame:
- reject_values
- include_values?
- replace_values

[Here](https://gist.github.com/lokeshh/13ae3d100d6294c0c2d216c62170915f) are the benchmarks.

Methods removed in Daru::Vector:
- missing_values
- missing_values=
- update
- exists?
- set_missing_positions

and other methods have been deprecated.

lokeshh:
All done. This is ready to be merged.

v0dro:
Works. Something will probably break in statsample so you might need to fix it.

lokeshh:
@v0dro. Done. Modified the Statsample to work with this PR.

PR #212, opened 2016-08-13
Add to_nmatrix method in Daru::Vector

lokeshh:
Daru::DataFrame and Daru::Vector has `to_matrix` method to convert to `Matrix` object but no method in Daru::Vector to convert to `NMatrix` object. This PR accomplishes that. In addition it adds `is_numeric?` method to decide if vector contain all numeric values or not.


lokeshh:
This method will be useful to replace `Matrix` with `NMatrix` in Statsample-GLM.

v0dro:
In Ruby you can get away with just `numeric?` instead of `is_numeric?` since it allows question marks.

Make that change and I'll merge.

lokeshh:
I just found out that there's already `numeric?` implemented in Daru. So, I've removed `is_numeric?` completely and now using `numeric?` now.

lib/daru/vector.rb
```diff
@@ -830,6 +830,30 @@ def to_matrix axis=:horizontal
...
+    # @example
+    #   dv = Daru::Vector.new [1, 2, 3]
+    #   dv.to_nmatrix
+    #   # =>
+    #   # [
+    #   #   [1, 2, 3] ]
+    def to_nmatrix axis=:horizontal
+      raise ArgumentError, 'Can not convert to nmatrix'\
+        'because the vector is numeric' unless numeric? && !include?(nil)
+
```
lokeshh: @v0dro `numeric?` return true if vector contain `nil`s. Should this behavior be changed?

v0dro: No let it be. `nil` represents missing data, which can be numeric. Moreover, stats method work when missing data is present and they will malfunction if `numeric?` does not work the way it does.

v0dro:
Rebase against current master and I'll merge.

lib/daru/vector.rb
lokeshh: ok

lokeshh:
This can be merged now. Also could a Daru release be done now, so I can make sure Statsample-GLM build is passing and Formula Language PR can be merged.

PR #228, opened 2016-08-23
Replace is_nil?, not_nil?

### LINKED ISSUE #141: [API] Rename Vector#is_nil?
As the method not returns boolean (the only thing that predicate methods should return), it definitely shouldn't contain question mark in the name.

So, my proposal here is `#is_nil?` => `#nils`, `#not_nil?` => `#not_nils`.

lokeshh:
Fixes #141 

Replaces Vector#is_nil? and Vector#not_nil? with something else.

### DIFF
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -702,31 +702,6 @@ module Daru
       self
     end
 
-    # Returns a vector which has *true* in the position where the element in self
-    # is nil, and false otherwise.
-    #
-    # == Usage
-    #
-    #   v = Daru::Vector.new([1,2,4,nil])
-    #   v.is_nil?
-    #   # =>
-    #   #<Daru::Vector:89421000 @name = nil @size = 4 >
-    #   #      nil
-    #   #  0  false
-    #   #  1  false
-    #   #  2  false
-    #   #  3  true
-    #
-    def is_nil?
-      # FIXME: EXTREMELY bad name for method not returning boolean - zverok, 2016-05-18
-      recode(&:nil?)
-    end
-
-    # Opposite of #is_nil?
-    def not_nil?
-      recode { |v| !v.nil? }
-    end
-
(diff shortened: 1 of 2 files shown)


v0dro:
How about `include_values?` and `not_include_values?`? That way it's consistent with `Array#include?` and the extra `_values` conveys that the behavior differs a little from the `Array#include?`.  @zverok?

zverok:
What it will return? If arrays of booleans -- then I still don't think that "predicate method sign" (`?`) is appropriate.

v0dro:
@zverok it will return arrays of booleans. Yeah it makes sense to not have the `?`.

zverok:
Hey, guys, are we going to finalize this eventually? :) I think that we can at least go with `nils`/`not_nils` or `existance(*values)`/`non_existance(*values)`. Or whatever!

v0dro:
Yeah lets stick with `include_values`.

lokeshh:
We already have `include_values?` to return `true` and `false`. I think it will be confusing to have `include_values`. Lets's have `is_values`. It also signifies that we are checking if the value is the same as what we are mentioning.

v0dro:
> We already have include_values? to return true and false. I think it will be confusing to have include_values. Lets's have is_values. It also signifies that we are checking if the value is the same as what we are passing as argument, "include" could be confusing here. WDYT

Ah yes. `is_values?` seems like a good choice. @zverok ?

zverok:
To be honest, I like neither, but can't come with a better name :) So, let's go with `is_values`!

v0dro:
Works. `is_values?` is the way to go.

zverok:
Just without question mark, I thought we've agreed on that ;)

v0dro:
@lokeshh can you fix this? The PR is quite old.

lokeshh:
Done finally. No more delay :)

PR #241, opened 2016-08-31
fix documentation of contrast_code

lokeshh:
Fix the default value of `:full` option and make it option rather than parameter.


v0dro:
PR closed because mine was in more detail. Also elaborated the `option` tag.

PR #247, opened 2016-09-09
Refactoring `Daru::Index`

yui-knk:


lib/daru/index/index.rb
```diff
@@ -43,10 +43,6 @@ def each(&block)
       self
     end
 
-    def map(&block)
```
v0dro: The `Daru::Index` class is a super class for all other index classes (`MultiIndex`, `DateTimeIndex` and `CategoricalIndex`). The only way in which each element can be iterated over in these classes is by calling `to_a` on them since they utilize speciliazed data structures for storing their data. So `map` has been put here on purpose so that it can be shared by all the index classes that inherit from `Daru::Index`.

spec/index/index_spec.rb
```diff
@@ -280,4 +280,11 @@
       it { expect { idx.at 2, 3 }.to raise_error IndexError }
     end
   end
-end
\ No newline at end of file
+
+  context "Enumerable" do
```
v0dro: Can you follow the convention set in the rest of the file? The top level context is the method name in this case, followed by tests for that method that test particular functionality.

yui-knk: > The top level context is the method name in this case

I know this, but in this case I want to test `Daru::Index` to behave as like a Enumerable.
So I think we do not need to test all Enumerable method (`#map` and `#select` are sample methods to check `Daru::Index` is a Enumerable).

Do you think it is better to separate these test cases?

v0dro: Hmm yes that works. In that case, can you write an `it` block inside the `context 'Enumerable'` which specifies that the test is for all Enumerable methods? That way future contributors won't be confused.

yui-knk:
I updated. How do you think about it?

v0dro:
OK works. Can you fix failing tests?

yui-knk:
tests are fixed :)

PR #249, opened 2016-09-09
Filtering out 'vendor' from coverage targets

yui-knk:


v0dro:
What does this change do exactly?

yui-knk:
I first `bundle install --path vendor/bundle`. And then I run tests locally.
In this situation, before this commit, simplecov also calculates coverages for files
which are located under `vendor/bundle`.



This commit prevent simplecov calculating to coverages for files located under `vendor/bundle`.

zverok:
@v0dro I think I can understand it. In some Bunlder _global_ settings, it creates local gem cache in _current_ project, and it is in `./vendor/bundle` folder. (I have this setting some time ago, it is somewhat convenient, but creates unexpected complications sometimes, like described above)

v0dro:
Ok works.

PR #253, opened 2016-09-10
Test about `#on_offset?`

yui-knk:


spec/date_time/offsets_spec.rb
```diff
@@ -266,6 +267,16 @@
           DateTime.new(2015,1,31))
       end
     end
+
+    context "#on_offset?" do
+      it "returns false if date is on the offset" do
```
v0dro: The date you've specified is not on the offset so should this test not be `"returns false if date is not on the offset"`?

yui-knk: Nice catch!
I fixed.

PR #257, opened 2016-09-11
Fix https://github.com/v0dro/daru/issues/250

zverok:


spec/dataframe_spec.rb
```diff
@@ -2927,6 +2927,19 @@
...
+    it 'works' do
+      categories = %i[jan feb mar apr may jun jul aug sep oct nov dec]
+      df = Daru::DataFrame.rows([
+        [2014, 2, 1600.0, 20.0],
+        [2014, 3, 1680.0, 21.0],
+        [2016, 2, 1600.0, 20.0],
+        [2016, 4, 1520.0, 19.0],
+      ], order: [:year, :month, :visitors, :days])
+      df[:averages] = df[:visitors] / df[:days]
+      df[:month] = df[:month].map{|i| categories[i - 1]}
```
v0dro: Remove this test? :P

zverok: I'll rewrite it into something more test-alike :) Or we could lost the same functionality in the future.

spec/dataframe_spec.rb
```diff
@@ -2927,6 +2927,32 @@
...
+      actual = df.pivot_table(index: :month, vectors: [:year], values: :averages)
+      expected =
+        Daru::DataFrame.new(
+          [
+            [80.0, 80.0, nil],
+            [nil, 80.0, 80.0],
+          ], index: Daru::MultiIndex.from_tuples([[:apr], [:feb], [:mar]]),
+          order: Daru::MultiIndex.from_tuples([[:averages, 2016], [:averages, 2014]])
+        )
```
zverok: OK, Now with this "better" test we see some problems with pivot (messed up months and years order + unnecessary multiindex from monthes). I'll take a look at evening.

zverok: @v0dro OK, I've investigated the matters deeper and added some explanatory comments here and there. I assume, we should go with this "fix major bug" PR ASAP, and then take a look at `group_by` refactoring, maybe.

v0dro:
Read your comments. Could you please open an issue about this so it can be discussed further?

Issue #268, opened 2016-09-24
`Daru::DataFrame.rows` regression: cannot create empty data frames.


yemartin:
Commit 760fb19 broke creating empty data frames with `Daru::DataFrame#rows`. The following used to work up to commit 2ee4c4c.

Note: this is important because if `source_data` is retrieved dynamically from a database, we have no control when the result may have no rows.

``` ruby
require 'spec_helper.rb'

describe "Daru::DataFrame#rows" do
  it 'can create an empty data frame' do
    source_data = [] # potential result of SomeModel.where(some_condition).pluck(:a, :b, :c)
    expect do
      Daru::DataFrame.rows source_data, order: [:a, :b, :c]
    end.not_to raise_error
  end
end
```

zverok:
I'm on it.
I'm a big fan of breaking things (and probably fix them afterwards)

v0dro:
@zverok any update?

zverok:
Oh damn. I've fixed the issue the next day after it was reported and just forgot to do a PR (some family problems distracted me). Shame on me.

PR #269, opened 2016-09-24
Fix DataFrame.rows with empty argument (#268)

zverok:


v0dro:
Failing tests. Please make them pass and I'll merge.

zverok:
Oh, there is Rubocop fails, will fix them tomorrow. On my way home from EuRuKo in Bulgaria, have seen Matz and even managed to give a lightning talk about SciRuby and related matters :)

v0dro:
Do post the talk link on the mailing list when it is up!

zverok:
OK, I've just rebased from current master and now everything should be OK with Rubocop too :)

> Do post the talk link on the mailing list when it is up!

To be honest, it was an epic fail :) I wanted to talk about just too much, and had not enough time (lightning talks was 5 min there), so... I think I've just failed. But not before stressing on several important points and praising SciRuby :)

PR #270, opened 2016-10-08
261 - Resolve a bug with false values on dataframe create

### LINKED ISSUE #261: Joins turn `false` into `nil`
````ruby
left = Daru::DataFrame.new({ key: [1,2,3], left_value: [true, false, true] })
right = Daru::DataFrame.new({ key: [1,2,3], right_value: [true, false, true] })

left.join(right, on: [:key], how: :inner)
````

All of the values that should have stayed `false` have been converted to `nil`.

gnilrets:
This resolves #261.  The core issue was that creating a
dataframe from an array of hashes wasn't working properly
when some of the values were `false`.

### DIFF
--- a/lib/daru/core/merge.rb
+++ b/lib/daru/core/merge.rb
@@ -78,7 +78,7 @@ module Daru
       end
 
       def sanitize_merge_keys(merge_keys)
-        merge_keys.map { |v| v || NilSorter.new }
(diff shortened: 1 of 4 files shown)


gnilrets:
Booleans are so central to data analysis.  We need to be really careful using the common Ruby idioms `a || b` and `a ||= b` when working with data values so we don't confuse `false` and `nil`.

PR #273, opened 2016-10-15
Fix DataFrame.rows with empty argument (#268)

zverok:


zverok:
Yep, it is a consequence of Rubocop update to 0.44 Bozhidar have released on his birthday :) (though, both new "offences" are pretty relevant to code quality)

Issue #272, opened 2016-10-14
Return a Daru::Vector if multiple values present in mode.


dylandy:
I found that mode function can only return a single value even when the input vector has multiple modes. Is it a better choice for return a list of mode values rather than return a single value ?

v0dro:
Good idea. It will return a `Daru::Vector` incase of multiple values.

v0dro:
Example:
``` ruby
require 'daru'

v = Daru::Vector.new([1,2,3,4,4,5,5])
v.mode #=> Daru::Vector with [4,5]
```

db579:
I've implemented this to return a `Daru::Vector` as suggested but would it not be better to return an array of 1 or more values so that the return type is always consistent as doesn't need to be checked for on the calling side?

zverok:
@db579 Let's postpone this question or put it in a separate GitHub issue for discussion (I believe, we have several places with the same concept). In fact, the idiom is not unusual for Ruby:
```ruby
array = [1,2,3]
array[1] # => 2 # single value
array[1..2] # => [2, 3] # array of values
```
Though, in this case user explicitly requires one value or range of values, unlike the `#mode`.

PR #274, opened 2016-10-15
Return a Daru::Vector if multiple values present in mode.

dylandy:
Reference to  issue #272. Not sure if I wrote it correctly. :P

### DIFF
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -5,6 +5,8 @@ module Daru
     # the computationally intensive tasks.
     module Statistics
       module Vector # rubocop:disable Metrics/ModuleLength
+        extend Gem::Deprecate
+
         def mean
           @data.mean
         end
@@ -30,7 +32,10 @@ module Daru
         end
 
         def mode
-          frequencies.max { |a,b| a[1]<=>b[1] }.first
+          freq = frequencies.to_h
+          max = freq.max { |a,b| a[1] <=> b[1] }.last
+          output = freq.select { |_k,v| v == max }.keys
+          output.length == 1 ? output.first : Daru::Vector.new(output)
         end
 
         # Create a summary of count, mean, standard deviation, min and max of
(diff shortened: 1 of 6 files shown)


v0dro:
Can you make Travis pass?

lib/daru/maths/statistics/vector.rb
```diff
@@ -30,7 +30,9 @@ def median
         end
 
         def mode
-          frequencies.max { |a,b| a[1]<=>b[1] }.first
+          max = frequencies.values.max
```
v0dro: Don't call `values` use `frequencies.max { |a,b| a[1] <=> b[1] }.last`

lib/daru/maths/statistics/vector.rb
```diff
@@ -30,7 +30,9 @@ def median
...
+          output = frequencies.select { |_k,v| v == max }.to_a.map(&:first)
```
v0dro: You're calling `frequencies` twice. That's expensive. Cache it .

v0dro: No need to call `to_a.map`. Just `.values` will suffice since `select` will return a `Hash`.

dylandy:
Hello, I have passed the Travis test after the commit this time.

dylandy:
I have no idea why I failed on CI test after merging this time.

lokeshh:
See https://github.com/v0dro/daru/pull/279#issuecomment-258661492

v0dro:
I think if you rebase from the master branch your tests should pass now. Can you please do that?

lib/daru/maths/statistics/vector.rb
```diff
@@ -32,7 +32,10 @@ def median
         end
 
         def mode
-          frequencies.to_h.max { |a,b| a[1]<=>b[1] }.first
+          freq = frequencies.to_h
+          max = freq.max { |a,b| a[1] <=> b[1] }.last
```
zverok: Use `max_by`, please.

v0dro:
The build is still failing. Please make it pass.

dylandy:
Sorry for disappear too much time. I was too busy to do open source project recently. 

I just came up with following problem when I tried  ```rake spec```. I have no idea to fix this problem.

zverok:
@dylandy GSLWrapper is not defined if you have no GSL library installed: https://github.com/SciRuby/daru/blob/master/lib/daru/accessors/gsl_wrapper.rb#L1

I believe you need to install GSL to be fully able to take a part in development. Though, as a temporary solution, you can add the same `if Daru.has_gsl?` at the top of spec/accessors/gsl_wrapper_spec.rb

dylandy:
well, I finally fixed my library issue and it seems like somebody just solved this pr.
Thanks a lot.

PR #275, opened 2016-10-18
Add from_tsv and write_tsv

mrkn:
TSV format is popular to export/import tabular data as well as CSV format.
So I think we should have `write_tsv` and `from_tsv`.


gnilrets:
Seems a little redundant to me.  Whenever I hear "CSV" I automatically think any delimited text file, whether it be comma, tab, pipe, semicolon, whatever.....

zverok:
@gnilrets As I am saying in the review, convenience method `DataFrame#from_tsv` underlining "we can do this" is OK, but I am not sure about adding new functionality to `IO`

v0dro:
@mrkn don't you think its best to use `col_sep:` option for `from_csv`?

zverok:
@v0dro In public interface, I think it is OK to have `from_csv`/`from_tsv` just as an indicator "we can do that" (not everyone aware of the fact that Ruby's CSV library in fact is able to handle {Anything}SeparatedValues)

mrkn:
There are two options, I think.

1. Following @zverok opinion, remove new methods from `IO` and rewrite new methods in `DataFrame`.
2. Withdraw this pull-request.

I want to follow the development policy of this project.
I like the former, but the latter is also OK for me.

@v0dro How do you think?

v0dro:
There can be a shortcut method called `from_tsv` but I think all the major logic should lie in `from_csv` with a `col_sep:` option.

zverok:
@mrkn I believe we have a consensus here. Would you mind finalizing the PR so we can merge and prosper?

PR #276, opened 2016-10-24
correct the behavior of Vector#reorder! method

genya0407:
`Vector#reorder!` method overrode `@data` with `Array`. 
`@data` must be instance of `Accessors::ArrayWrapper`.


genya0407:
See #277

PR #277, opened 2016-10-27
correct the behavior of Vector#reorder! method

genya0407:
`Vector#reorder!` method overrode `@data` with `Array`. 
`@data` must be instance of `Accessors::ArrayWrapper`.

Same for #276 , only branch name differs.

### DIFF
--- a/lib/daru/index/multi_index.rb
+++ b/lib/daru/index/multi_index.rb
@@ -17,12 +17,10 @@ module Daru
     def initialize opts={}
       labels = opts[:labels]
       levels = opts[:levels]
-      raise ArgumentError,
-        'Must specify both labels and levels' unless labels && levels
-      raise ArgumentError,
-        'Labels and levels should be same size' if labels.size != levels.size
-      raise ArgumentError,
-        'Incorrect labels and levels' if incorrect_fields?(labels, levels)
+
+      raise ArgumentError, 'Must specify both labels and levels' unless labels && levels
+      raise ArgumentError, 'Labels and levels should be same size' if labels.size != levels.size
+      raise ArgumentError, 'Incorrect labels and levels' if incorrect_fields?(labels, levels)
 
       @labels = labels
       @levels = levels.map { |e| e.map.with_index.to_h }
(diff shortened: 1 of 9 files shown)


lib/daru/vector.rb
```diff
@@ -964,7 +964,7 @@ def reindex! new_index
     #   #   c   3
     def reorder! order
       @index = @index.reorder order
-      @data = order.map { |i| @data[i] }
+      @data = Accessors::ArrayWrapper.new(order.map { |i| @data[i] }, self)
```
v0dro: Good catch.

The type of `@data` should depend on the value of the of the `@dtype` variable. Could you just add this line instead of your change:
``` ruby
cast_vector_to @dtype, @data, @nm_dtype
```

v0dro:
Also make sure that the Travis CI build passes.

genya0407:
@v0dro: On Travis CI, some rubocop checks have failed. 
However, in my local environment, no rubocop check fails.
In fact, these checks didn't failed in the previous Travis CI check, and I haven't modify that file since then.

Do you have any ideas why these rubocop checks fail?

lokeshh:
Try `bundle update`.  It could be that your gems are not up to date.

genya0407:
@lokeshh Thanks! `bundle update` updated my rubocop, and same rubocop checks failed.

@v0dro Should I include commits that fixes rubocop checks, or make another Pull Request that fixes rubocop check?

lib/daru/index/multi_index.rb
```diff
@@ -17,12 +17,10 @@ def levels
...
       levels = opts[:levels]
-      raise ArgumentError,
-        'Must specify both labels and levels' unless labels && levels
-      raise ArgumentError,
-        'Labels and levels should be same size' if labels.size != levels.size
-      raise ArgumentError,
-        'Incorrect labels and levels' if incorrect_fields?(labels, levels)
+
+      raise ArgumentError, 'Must specify both labels and levels' unless labels && levels
+      raise ArgumentError, 'Labels and levels should be same size' if labels.size != levels.size
```
v0dro: Technically no. But its not a hard rule and can be relaxed if its going slightly more than 80 and readability is enhanced as a result.

PR #282, opened 2016-11-23
Fix #281

### LINKED ISSUE #281: Serious group_by bug
Looks like there's a pretty serious bug with group_by when the index isn't a sequential set of integers.

````
wtdf = Daru::DataFrame.new({ g: ['a','a','a'], num: [1,2,3]}, index: [2,12,23])
puts wtdf.sum.inspect
#=> #<Daru::Vector(1)>
#     sum
# num   6

puts wtdf.group_by([:g]).sum.inspect
#=> #<Daru::DataFrame(1x1)>
#     num
#   a   4
````

I think the problem is the way that group_by is making use of indexes:
````
wtdf.group_by([:g])
#=> #<Daru::Core::GroupBy:0x007f9997672950 @groups={["a"]=>[0, 1, 2]}, @non_group_vectors=[:num], @context=#<Daru::DataFrame(3x2)>
#       g num
#   2   a   1
#  12   a   2
#  23   a   3>
````

The groups for the `"a"` category reference indexes `[0, 1, 2]`.  The group_by code is confusing order in the dataframe with the index.  So it first attempts to find index 0, which doesn't exist so it assumes it's the just the first element of the data frame.  Likewise for index 1.  But then it looks for index 2, which does exist, but it's the first record in the dataframe and that's already been counted.

zverok:
Well, it seems to work... But I am afraid this liberalism in accepting both order number and index in same contexts will fire our feet off several more times.

### DIFF
--- a/lib/daru/core/group_by.rb
+++ b/lib/daru/core/group_by.rb
@@ -302,7 +302,11 @@ module Daru
             indexes << index
           end
         end
-        indexes
+        if indexes.count == 1
+          [@context.index.at(*indexes)]
+        else
+          @context.index.at(*indexes).to_a
+        end
       end
 
       def multi_indexed_grouping?
--- a/spec/core/group_by_spec.rb
+++ b/spec/core/group_by_spec.rb
@@ -405,9 +405,17 @@ describe Daru::Core::GroupBy do
 
   context 'groups by first vector if no vector mentioned' do
     subject { @df.group_by }
-    
+
     it { is_expected.to be_a Daru::Core::GroupBy }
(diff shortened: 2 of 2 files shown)


v0dro:
@gnilrets  your thoughts on this?

gnilrets:
Looks like it will solve the immediate issue, but I agree that it will probably come up again.

v0dro:
Hmmm we can look at this again later. I will merge this for now.

v0dro:
Merging for now. We'll see if a bug crops up later.

PR #288, opened 2016-12-25
DataFrame summary without ReportBuilder fixes #221

### LINKED ISSUE #221: Better DataFrame summary without reportbuilder
The `DataFrame#summary` method uses the `reportbuilder` gem for creating a summary of the data. This gem, in addition to being a dependency, was designed keeping `Statsample::Dataset` in mind, which daru moved away from a long time ago.

Thus, a custom implementation for summarizing dataframe contents should be created that does not depend on reportbuilder and shows ample information about the dataframe at a glance.

For reference, here's the output of `#summary` for a data frame:

```
2.2.1 :015 > source_df
 => #<Daru::DataFrame(1x2)>
      col1 col2
    0    1    2 
2.2.1 :016 > puts source_df.summary
= 
  Number of rows: 1
  Element:[col1]
  == col1
    n :1
    n valid:1
  Element:[col2]
  == col2
    n :1
    n valid:1
 => nil 
2.2.1 :017 > 
```

EDIT:
The above example is only a demonstration. The reportbuilder gem is capable of many more things that just the above summary, and before sending a PR for implementing this feature you should go through the documentation of reportbuilder and see what it is capable of and port most (if not all) of those functions to daru w/o introducing new dependencies.

krishnak9:
#221  fixes

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1407,7 +1407,10 @@ module Daru
 
     # Generate a summary of this DataFrame with ReportBuilder.
     def summary(method=:to_text)
-      ReportBuilder.new(no_title: true).add(self).send(method)
+      #uncomment this line to use ReportBuilder gem to generate data frame summary
+      #ReportBuilder.new(no_title: true).add(self).send(method)
+      #comment below line when using Reportbuilder gem
+      g_summary()
     end
 
     def report_building(b) # :nodoc: #
@@ -1419,6 +1422,29 @@ module Daru
         end
       end
     end
+    #custom function for dataframe summary
+    def g_summary()
+       frame_summary = "= \t\n"
+       frame_summary << "\tNumber of rows: #{nrows}\n"
+       @raw = self
+       #no_of_vect = self.vectors.count
+       
+       @raw.each do |v|
+        frame_summary << "\tElement:[#{v.name}]\n"
+        frame_summary << "\t== #{v.name}\n"
+        count = v.size
+        valid_count = 0
+        v.to_a.each do |val_n|
+           if (val_n!=nil && !(val_n.is_a?(Float) && val_n.nan?))
+              valid_count = valid_count + 1 
+           end
+        end
+       
+        frame_summary << "\t  n : #{count}\n"
+        frame_summary << "\t  n valid : #{valid_count}\n"
+       end
+       frame_summary
+     end
 
     # Sorts a dataframe (ascending/descending) in the given pripority sequence of
     # vectors, with or without a block.
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1405,46 +1405,21 @@ module Daru
       Daru::DataFrame.new(arry, clone: cln, order: order, index: @index)
     end
 
-    # Generate a summary of this DataFrame with ReportBuilder.
+    # Generate a summary of this DataFrame .
     def summary(method=:to_text)
-      #uncomment this line to use ReportBuilder gem to generate data frame summary
-      #ReportBuilder.new(no_title: true).add(self).send(method)
-      #comment below line when using Reportbuilder gem
-      g_summary()
+      summary_building()
     end
 
-    def report_building(b) # :nodoc: #
-      b.section(name: @name) do |g|
-        g.text "Number of rows: #{nrows}"
+    def summary_building # :nodoc: #
+      df_summary = "=\n"
+        df_summary << "  Number of rows: #{nrows}\n"
         @vectors.each do |v|
-          g.text "Element:[#{v}]"
-          g.parse_element(self[v])
+          df_summary << "  Element:[#{v}]\n"
+          df_summary << self[v].summary_building(2) 
+        
         end
-      end
+      df_summary
     end
-    #custom function for dataframe summary
-    def g_summary()
-       frame_summary = "= \t\n"
-       frame_summary << "\tNumber of rows: #{nrows}\n"
-       @raw = self
-       #no_of_vect = self.vectors.count
-       
-       @raw.each do |v|
-        frame_summary << "\tElement:[#{v.name}]\n"
-        frame_summary << "\t== #{v.name}\n"
-        count = v.size
-        valid_count = 0
-        v.to_a.each do |val_n|
-           if (val_n!=nil && !(val_n.is_a?(Float) && val_n.nan?))
-              valid_count = valid_count + 1 
-           end
-        end
-       
-        frame_summary << "\t  n : #{count}\n"
-        frame_summary << "\t  n valid : #{valid_count}\n"
-       end
-       frame_summary
-     end
 
     # Sorts a dataframe (ascending/descending) in the given pripority sequence of
     # vectors, with or without a block.
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1406,18 +1406,17 @@ module Daru
     end
 
     # Generate a summary of this DataFrame .
-    def summary(method=:to_text)
-      summary_building()
+    def summary()
+      summary_building
     end
 
     def summary_building # :nodoc: #
       df_summary = "=\n"
-        df_summary << "  Number of rows: #{nrows}\n"
-        @vectors.each do |v|
-          df_summary << "  Element:[#{v}]\n"
-          df_summary << self[v].summary_building(2) 
-        
-        end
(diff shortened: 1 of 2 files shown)


krishnak9:
#221  fixes

lib/daru/dataframe.rb
```diff
@@ -1407,7 +1407,10 @@ def only_numerics opts={}
 
     # Generate a summary of this DataFrame with ReportBuilder.
     def summary(method=:to_text)
-      ReportBuilder.new(no_title: true).add(self).send(method)
+      #uncomment this line to use ReportBuilder gem to generate data frame summary
+      #ReportBuilder.new(no_title: true).add(self).send(method)
+      #comment below line when using Reportbuilder gem
```
v0dro: There's no need for these comments or even then `g_summary` method. ReportBuilder should go.

lib/daru/dataframe.rb
```diff
@@ -1419,6 +1422,29 @@ def report_building(b) # :nodoc: #
         end
       end
     end
+    #custom function for dataframe summary
+    def g_summary()
+       frame_summary = "= \t\n"
```
v0dro: `df_summary` is better. Why the `\t\n` at the end? Have you seen how exactly ReportBuilder prepares reports?

lib/daru/dataframe.rb
```diff
@@ -1419,6 +1422,29 @@ def report_building(b) # :nodoc: #
...
+       frame_summary << "\tNumber of rows: #{nrows}\n"
+       @raw = self
```
v0dro: Why not use `self` directly? Using instance variables like this is not acceptable since it stays for absolutely no reason in the object and consumes resources. IVs should not be initialized unless you absolutely must.

lib/daru/dataframe.rb
```diff
@@ -1419,6 +1422,29 @@ def report_building(b) # :nodoc: #
...
+        valid_count = 0
+        v.to_a.each do |val_n|
+           if (val_n!=nil && !(val_n.is_a?(Float) && val_n.nan?))
+              valid_count = valid_count + 1 
+           end
+        end
+       
+        frame_summary << "\t  n : #{count}\n"
+        frame_summary << "\t  n valid : #{valid_count}\n"
+       end
```
v0dro: You should go through the code of the report builder gem and see ALL  the data that it outputs. It exists for more reasons than just telling `n` and `n_valid`.

The patch you submit must emulate that output as much as possible. Try running reportbuilder through different examples and see their respective outputs.

krishnak9: I will do all the changes as you mentioned in the suggestions and will go through the reportbuilder gem code and try to emulate all of its features.

krishnak9:
Can I use text-table gem to create table for distribution in vector.

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
-          s.text  "mode: #{mode}"
-
-          s.table(name: 'Distribution') do |t|
-            frequencies.sort_by(&:to_s).each do |k,v|
-              key = @index.include?(k) ? @index[k] : k
-              t.row [key, v, ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))]
+    def summary_building level=0 # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
+      
+        v_summary = ""
+        v_summary << "  "*2*level+"n :#{size}\n"
```
zverok: This `"  "*2*level` repeating on each line is pretty ugly. Could it be moved to variable?..

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
+        valid_count = 0
+        self.to_a.each do |val_n|
```
zverok: `self.` is unnecessary

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
+            if (val_n!=nil && !(val_n.is_a?(Float) && val_n.nan?))
```
zverok: * `()` around condition is unnecessary
* we never do `!= nil` in Ruby; you can do `if !val.nil?`, or, in the most cases, `if val && ....`
* ...or even hide it in helper method `if valid_value?(val)`

lib/daru/vector.rb
zverok: ...and we never use `each` to count something, use `Array#count`.
...and, finally, I believe vector already has method "number of valid values"

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
-          end
         end
+        v_summary << "  "*2*level+"n valid:#{valid_count}\n"
+        if @type == :object
+          v_summary  << "  "*2*level+"factors: #{factors.to_a.join(',')}\n"
+          v_summary  << "  "*2*level+"mode: #{mode}\n"
 
-        s.text "median: #{median}" if @type==:numeric || @type==:numeric
+        end
+        if @type==:numeric || @type==:numeric
```
zverok: 1. `elsif`
2. why the `||` of same condition?

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
+          v_summary << "  "*2*level+"median: #{median}\n"
+        end 
         if @type==:numeric
```
zverok: why the `if` of same condition as above?

lib/daru/vector.rb
```diff
@@ -885,37 +885,41 @@ def to_s
...
-        s.text "n valid:#{count_values(*Daru::MISSING_VALUES)}"
-        if @type == :object
-          s.text  "factors: #{factors.to_a.join(',')}"
-          s.text  "mode: #{mode}"
-
-          s.table(name: 'Distribution') do |t|
-            frequencies.sort_by(&:to_s).each do |k,v|
-              key = @index.include?(k) ? @index[k] : k
-              t.row [key, v, ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))]
+    def summary_building level=0 # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
```
krishnak9: Actually it was already disabled in the previous code.

lib/daru/vector.rb
krishnak9: Okay i will move that to variable.

lib/daru/vector.rb
zverok: Well, yes :) But daru was evolving quickly, and now we try to make the code quality better on each PR, without compromises!

zverok:
Well, looking closer, I see most of my notes are related to pre-existing code... But still, the library becomes mature now, and so we are struggling for better code quality with each commit. Report building was one of the parts sitting here since ancient times, and the idea of rewriting it is not only drop ReportBuilder, but also make the code shiny and readable.

krishnak9:
I will try to implement as much as good as I can make.I will try to improve on my next commits. And I will make a new pull request with new branch when changes are final so that these confusing commits doesn't show up on commit history.

krishnak9:
ohh okay i will make the changes. Are vector_summary and vector_indent better ?

zverok:
Well, we are already inside `Vector`s method, why we need any prefix at all?..

lib/daru/vector.rb
```diff
@@ -884,40 +884,29 @@ def to_s
...
-    # :nocov:
-    def report_building b # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-      b.section(name: name) do |s|
-        s.text "n :#{size}"
-        s.text "n valid:#{count_values(*Daru::MISSING_VALUES)}"
-        if @type == :object
-          s.text  "factors: #{factors.to_a.join(',')}"
-          s.text  "mode: #{mode}"
-
-          s.table(name: 'Distribution') do |t|
```
krishnak9: To implement table can i use text-table gem ,any suggestion for that.

zverok: Well, the idea of "drop reportbuilder" is to make the list of dependencies smaller. Will text-table make the code significantly simpler? And if so, is it the better "table formatting" gem selection? It was not updated for two years.

krishnak9: Well then I will try to make a custom implementation but i do not understand what this table actually does . I checked with some example and the thing is `frequencies.sort_by(&:to_s).each do |k,v|` returns a vector , so why use two variable k,v. Please help me in this

v0dro:
Closing this. Send a new PR with fixes.

PR #290, opened 2017-01-03
Comparison Operator for Daru::Vector

genya0407:
For https://github.com/v0dro/daru/issues/287 .

I didn't create method-alias for `==` and `!=`, because they conflict with `==` and `!=` that check equivalence of two vectors.


genya0407:
The second commit is for nokogiri's version issue.
nokogiri 1.7 doesn't support ruby 2.1.0.

PR #292, opened 2017-01-08
fix typo

genya0407:
`DataFrame#vectors=` checks if the argument is a `Daru::Index`.
However, because of a typo, it hasn't worked.


lib/daru/dataframe.rb
```diff
@@ -1346,7 +1346,7 @@ def index= idx
     #   df.vectors = Daru::Index.new([:foo, :bar, :baz])
     #   df.vectors.to_a #=> [:foo, :bar, :baz]
     def vectors= idx
-      unless index.is_a?(Daru::Index)
+      unless idx.is_a?(Daru::Index)
```
v0dro: The original version is correct. Its checking if the argument `idx` is not of type `Daru::Index`.

genya0407: Do you mean that `idx` should be `Daru::Index` ?

v0dro: The above code is invalid. The RHS must be a `Daru::Index`.

I don't get this... the statement is `unless idx.is_a?(Daru::Index)`. That means that the error should be thrown iff the object passed is anything but a `Daru::Index`.

genya0407: Please let me confirm:

The following code is **valid**:
```ruby
df.vectors = Daru::Index.new [:a, :b, :c]
```

And, the following code is **invalid** and should **throw error**:

```ruby
df.vectors = [:a, :b, :c]
```
Is this ok? Am I correct?

In addition, let me confirm one more thing.
The original code is `index.is_a?(Daru::Index)`,
and the code I wrote is `idx.is_a?(Daru::Index)`.

v0dro: You're correct. My bad.

PR #295, opened 2017-01-17
Return vector for multiple modal values

db579:
PR to resolve https://github.com/SciRuby/daru/issues/272

image


db579:
All specs passing, Travis failure seems to be due to Rubocop complaining about the use of `Marshall.load` elsewhere in the code.

PR #296, opened 2017-01-21
Add imagemagick in daru development guide

lokeshh:
In order to install rmagick gem (dependency for daru) I had to execute the following command

`sudo apt-get install libmagickwand-dev imagemagick`

I think it will be good to be added it in the contributing guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,7 @@ To install dependencies, execute the following commands:
 *  `sudo apt-get update -qq`
 *  `sudo apt-get install -y libgsl0-dev r-base r-base-dev`
 *  `sudo Rscript -e "install.packages(c('Rserve','irr'),,'http://cran.us.r-project.org')"`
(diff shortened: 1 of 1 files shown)


zverok:
/me looks at sciryb-dev mailing list
Perfect timing :)

Though, what do you think of making all external library dependencies optional even for development? I can see both pro and contra this approach. 
Pro: some novice/student can try to fix some small bug/propose feature without thinking of "how do I install RMagick" (which, FYI, is pretty tricky on Windows).
Contra: this student, in fact, can break something that only on Gruff plotting is visible...

v0dro:
@zverok how would the dependencies be removed without depending on gruff? I say test the whole gem in one go so that everything stays in order and nothing breaks.

zverok:
Well, The Ultimate Solution will be gems separation into daru-core, daru-plotting, daru-gsl and so on :) But I am not sure whether it worth trying :)

Issue #289, opened 2016-12-29
Read data by URL


zverok:
Another one stolen from our competitors (as #127). They can do this:

```r
df = pandas.read_csv('https://download.data.world/sql_query_result_download/producthunt/product-hunt-research?filename=product-hunt-research-QueryResult&mimetype=text%2Fcsv&query=--%20NOTE%3A%20from%20a%20sample%20of%205000%20records%20in%20the%20full%20dataset%0ASELECT%20name%2C%20time_of_day%2C%20PostsForExploration.date%0AFROM%20PostsForExploration%20&auth=<USER-FILE-SPECIFIC-TOKEN>')
```

With Ruby's `open-uri`, it would not be too hard to reproduce, and, again, not that useful -- but "cool looking" on demos.

v0dro:
Works.

Varunram:
@v0dro Do we add this as an extension?

v0dro:
Extension as in?

anshuman23:
Submitted the PR for this issue. I will also work on #127 

Please let me know if I there is anything that I can improve on in the PR.

athityakumar:
Fixed with PR #334 - I think this issue can be closed now. :smile:

zverok:
:+1:

PR #300, opened 2017-02-04
Added functionality as per Issue #289

anshuman23:
Image Screenshot of two dataframes being loaded from csv files hosted on these links -> 
1. https://raw.githubusercontent.com/anshuman23/ruBayes/master/test/sample.csv
2. https://raw.githubusercontent.com/anshuman23/ruBayes/master/train/train2.csv

The second csv file is the Iris Dataset.


anshuman23:
Hi!
I'm kind of new to this, so I'm sorry if I did not do the obvious. 
I just committed a small test to check url loading capability and had also checked styling using rubocop. I also made the code a little simpler and wrote it around from_csv.
I had already committed these changes before I saw your suggestions. I will make the required changes soon and update my PR accordingly to address the issues raised.  
Thank you!

Gemfile
```diff
@@ -1,2 +1,2 @@
-source 'https://rubygems.org'
+source 'https://ruby.taobao.org/'
```
zverok: ???

lib/daru.rb
```diff
@@ -96,6 +96,8 @@ def plotting_library= lib
 require 'erb'
 require 'date'
 
+require 'open-uri'
```
zverok: Better to require it where it is really needed (in io). This way the separation of concerns would be clearer.

spec/io/io_spec.rb
```diff
@@ -34,6 +34,16 @@
       end
     end
 
+    context ".from_csv_url" do
+      it "loads from a CSV file" do
+        df = Daru::DataFrame.from_csv_url('https://raw.githubusercontent.com/anshuman23/ruBayes/master/test/sample.csv')
```
zverok: Better use some mock library (webmock or VCR) to test network interactions. Imagine somebody developing new features and restarting specs every now and then -- and each time waiting for this small spec to connect to network and download test data (which also could change or vanish in future).

spec/io/io_spec.rb
```diff
@@ -34,6 +34,16 @@
...
+
+        df.vectors = [:country, :region].to_index
```
zverok: Should not be necessary, as source data already have column headers?..

spec/io/io_spec.rb
anshuman23: I have made all the other changes, but this is the only thing I'm a little bit unsure of in terms of implementation.

I know how to use webmock and vcr together to create an offline response, however, I'm not sure where I will 'require' all the modules (vcr and webmock here) as all the test files essentially cannot have modules being called inside (?)
This is also related to another problem - where do I write the initial configuration for the vcr cassette being used? 

If you could give me some insight, I'll make the required additions ASAP

anshuman23:
Hi @zverok , I have made the appropriate changes. If possible I'd like to ask you a few simple questions -
1. I have created a vcr_config.rb file which I use in the io_spec.rb test -
```
require 'webmock'                                                                                                                                                        
require 'vcr'                                                                                                                                                            
                                                                                                                                                                         
VCR.configure do |config|                                                                                                                                                
  config.cassette_library_dir = '../fixtures/vcr_cassettes'                                                                                                              
  config.hook_into :webmock                                                                                                                                              
end     
```
However, I think the main problem here would still be that VCR would be able to give me the HTTP GET object offline from the URL, but that is being done internally by 'from_csv_url' using open-uri. So what do you advise regarding this?

2. To add the vcr and webmock dependencies should I add them to the Gemfile directly using gem 'webmock' and gem 'vcr'?

If you are okay with these changes, I'll commit them to finalize the PR.

Thank you!

spec/io/io_spec.rb
zverok: Typically, all requirements and configuration of testing libraries are done in `spec/spec_helper.rb`, and Daru is not an exception :)

zverok:
@anshuman23 Thanks for your work! But I still feel like code could be made a bit better. Note the two things:
* highly repetitive code in `from_csv`/`from_csv_url` and `from_csv_hash`/`from_csv_hash_url`
* `from_csv` and `from_csv_url` now have different amount of features (the latter doesn't have `:headers` option).

WDYT about just making `from_csv` just a bit "smarter" -- with `open-uri` it would be litterally just using `CSV.new(open(path_or_url))` here: https://github.com/SciRuby/daru/blob/master/lib/daru/io/io.rb#L206 and there: https://github.com/SciRuby/daru/blob/master/lib/daru/io/io.rb#L214, and make sure nothing broken in tests + providing tests for new behavior (reading from URL).

> However, I think the main problem here would still be that VCR would be able to give me the HTTP GET object offline from the URL, but that is being done internally by 'from_csv_url' using open-uri. So what do you advise regarding this?

Please look at VCR usage examples at its docs. VCR+webmock (or just webmock, for such a simple case) for making any "real" network calls mocked with test data.

> To add the vcr and webmock dependencies should I add them to the Gemfile directly using gem 'webmock' and gem 'vcr'?

Yes, just in proper groups (development/test, not main dependencies).

anshuman23:
Thank you so much for the advice and the feedback. I realized the messiness of the code even when I was coding but I was so impatient to get done that I overlooked the value of the work.
I'll take into consideration all that you've said and commit better changes soon. Thanks again!

Gemfile
```diff
@@ -1,2 +1,2 @@
-source 'https://rubygems.org'
+source 'https://rubygems.org/'
```
v0dro: why the `/`?

lib/daru/dataframe.rb
```diff
@@ -4,6 +4,7 @@
 require 'daru/plotting/gruff.rb'
 require 'daru/plotting/nyaplot.rb'
 require 'daru/io/io.rb'
+require 'open-uri'
```
v0dro: This is gem-wide dependency and does not belong in this file. Put in `daru.rb`.

lib/daru/dataframe.rb
```diff
@@ -48,6 +49,11 @@ def from_csv path, opts={}, &block
         Daru::IO.from_csv path, opts, &block
       end
 
+      # Load a dataframe using a csv url
+      def from_csv_url url, opts={}, &block
```
v0dro: No need for new methods.

v0dro:
I am against introducing any new methods for reading a CSV. You should be able to figure out whether a string contains a URL or a file path by parsing it. All these features should be inside `from_csv` itself.

lib/daru/dataframe.rb
anshuman23: Oh okay! I'd initially put it in daru.rb but zverok thought that it should reflect where its being used. I'll put it back right away

lib/daru/dataframe.rb
anshuman23: Understood. I will make the changes and get back with better code

Gemfile
anshuman23: I'd changed it to the Asia specific server for rubygems because rubygems.org was down momentarily or something. When I pushed the commit I think I left a '/' there by mistake. 
Will edit this too.

anshuman23:
@zverok , @v0dro  - I have made changes as guided. I hope the latest commits are not lacking in any way. 
Please let me know your reviews. Thank you :)

lib/daru/io/io.rb
```diff
@@ -202,23 +200,32 @@ def from_csv_prepare_opts opts
...
+        if %w(http https).include?(url.scheme)
+          ::CSV
+            .parse(open(path), opts)
+            .by_col.map { |col_name, values| [col_name, values] }.to_h
+        else
+          ::CSV
+            .read(path, 'rb',opts)
+            .tap { |c| yield c if block_given? }
+            .by_col.map { |col_name, values| [col_name, values] }.to_h
+        end
```
zverok: The whole idea of `open-uri` library is to uniformely open URIs and local paths. I believe, the code in lines 205-207 should be enough for both cases (local file and URI).

lib/daru/io/io.rb
```diff
@@ -202,23 +200,32 @@ def from_csv_prepare_opts opts
...
-          .tap { |c| yield c if block_given? }
-          .to_a
-
+          if %w(http https).include?(url.scheme)
+            ::CSV.parse(open(path), opts)
+          else
+            ::CSV
+              .open(path, 'rb', opts)
+              .tap { |c| yield c if block_given? }
+              .to_a
```
zverok: Same as above.

spec/spec_helper.rb
```diff
@@ -67,3 +68,10 @@ def unindent
 end
 
 Dir[File.expand_path('../support/**/*.rb', __FILE__)].each {|f| require f }
+
+RSpec.configure do |config|
+  config.before(:each) do 
+    stub_request(:get,'https://raw.githubusercontent.com/anshuman23/ruBayes/master/test/sample.csv').
+      to_return(status: 200, body: "a,b,c,d,e\n4.8,3.4,1.9,0.2,Iris-setosa\n5.0,3.0,1.6,0.2,Iris-setosa\n5.0,3.4,1.6,0.4,Iris-setosa\n5.2,3.5,1.5,0.2,Iris-setosa\n6.6,3.0,4.4,1.4,Iris-versicolor\n6.8,2.8,4.8,1.4,Iris-versicolor\n6.7,3.0,5.0,1.7,Iris-versicolor\n6.0,2.9,4.5,1.5,Iris-versicolor\n6.4,3.1,5.5,1.8,Iris-virginica\n6.0,3.0,4.8,1.8,Iris-virginica\n6.9,3.1,5.4,2.1,Iris-virginica\n6.7,3.1,5.6,2.4,Iris-virginica\n")
+  end
```
zverok: This stub implies _most_ of your tests needs it, which is not true. The stub should be in io_spec.rb, near the actual test. And contents of "stabbed" csv should not be pasted in code, but stored in `spec/fixtures` folder (or rather use one of existing there fixtures).

anshuman23:
@zverok - Thanks again for the feedback. I've made these changes as well and pushed the commits. Please let me know if there is still something to be done.

anshuman23:
Hi @zverok @v0dro  - Is there something wrong with the commits I had pushed? If there are existing problems I'll rectify them although I did take into account all the suggestions and wrote the test for the functionality too

lib/daru/io/io.rb
```diff
@@ -202,23 +200,16 @@ def from_csv_prepare_opts opts
...
-          .tap { |c| yield c if block_given? }
+          .parse(open(path), opts)
           .by_col.map { |col_name, values| [col_name, values] }.to_h
       end
 
       def from_csv_hash(path, opts)
         csv_as_arrays =
-          ::CSV
-          .open(path, 'rb', opts)
-          .tap { |c| yield c if block_given? }
```
v0dro: If you remove the `yield` the user loses the functionality to manipulate rows with some provided block right?

So something like:
``` ruby
DataFrame.from_csv('file.csv') do |row|
  row[0] = row[0].to_i
end
```

Will not work..

v0dro: This reminds that we should probably be having a test where a block is passed to the `from_csv` function.

anshuman23: Yes, this does make sense. So what I was doing in the previous commit is what should be done, right? Differentiating between a URL and a local csv should be the way to go like this? -
```
def from_csv_hash_with_headers(path,opts) 
      url = URI.parse(path)
      if %w(http https).include?(url.scheme)
         ::CSV
           .parse(open(path), opts)
           .by_col.map { |col_name, values| [col_name, values] }.to_h
      else
         ::CSV
           .read(path, 'rb',opts)
           .tap { |c| yield c if block_given? }
           .by_col.map { |col_name, values| [col_name, values] }.to_h
      end
end
```

And similarly for the function  ```from_csv_hash```

I will write the test for passing the block too along with this. :+1:

athityakumar: As the `from_html` module uses a similar logic to differentiate between local and url files, I was thinking about this flow - 

```
def from_html path
    if File.exists? path
        # Parse from File.read 
    else
        begin
            # Parse from URI.parse()
        rescue
            # File not found - raise error / display error message
        end
    end
end
```

(The `if-else` can be converted to `? :` or other forms, if required.)
Do comment your insights on this, @anshuman23 @zverok @v0dro :smile:

zverok: @v0dro I am wondering, if anybody relies on this behavior (e.g. an ability to pass block for preprocessing the data)? It is untested and [not really documented](http://www.rubydoc.info/gems/daru/0.1.5/Daru/DataFrame#from_csv-class_method) so probably we could consider just dropping it?..

spec/io/io_spec.rb
```diff
@@ -32,6 +32,15 @@
         df = Daru::DataFrame.from_csv 'spec/fixtures/sales-funnel.csv'
         expect(df.vectors.to_a).to eq(%W[Account Name Rep Manager Product Quantity Price Status])
       end
+
+      it "works for urls (header comparison for mock data)" do
+        WebMock.stub_request(:get,'https://raw.githubusercontent.com/anshuman23/ruBayes/master/test/sample2.csv').
+          to_return(status: 200, body: File.read('spec/fixtures/scientific_notation.csv'))
```
zverok: Please, context + stubbing in before of that context. BTW, with stubs there's no need to use real (and long URL), some `http://dumm/some.csv` will be enough.

zverok: (And in any case it is better to have URL in separate `let` statement.

anshuman23:
@zverok - I have made the changes. Please review. :)

I will write the tests for #308 in a separate PR

spec/io/io_spec.rb
```diff
@@ -33,7 +33,17 @@
...
     end
-
+    
+    context "+ stubbing" do
+      it ".from_csv works for urls (header comparison for mock data)" do
+        WebMock.stub_request(:get,'http://dumm/some.csv').
+          to_return(status: 200, body: File.read('spec/fixtures/scientific_notation.csv'))
+        df = Daru::DataFrame.from_csv('http://dumm/some.csv', col_sep:',', headers: true)
+        expect(df.vectors).to eq([:x, :y].to_index)
+      end
```
zverok: It doesn't look like a good test. Please, move `WebMock` stuff to `before` block

v0dro:
Quick aside: The title of your PR should clearly state what issue it address explicitly. Listing just the issue number is not very useful for a reader.

v0dro:
@anshuman23 whats the status of this PR?

v0dro:
Has been resolved by #334

Issue #298, opened 2017-01-29
Open SQL databases by specifying file name


v0dro:
Currently the `.from_sql` method requires a `DBI::DatabaseHandle` object for loading data from an SQL database. In case you have a `database.db` file in your system, it becomes cumbersome to load this data.

There should be support for loading a database from such a file with the `Daru::DataFrame.from_sql` method.

Example:
``` ruby
Daru::DataFrame.from_sql '/path/to/db/database.db', 'some SQL query string`
```

zverok:
Probably related: https://github.com/SciRuby/daru/issues/138

athityakumar:
@v0dro @zverok - Can we use `ActiveRecord` gem here to connect with SQL database and collect table responses as an `object` / `'list`? After receiving an `object` of a table, we can probably then tweak the `from_sql` function of `Daru::DataFrame` module.

zverok:
@athityakumar Well, I am definitely against adding ActiveRecord as a dependency for a Daru core. In attempt to support any DB possible, with direct dependency on some abstraction layer/ORM, probably companion gem like `daru-db` could be created. @v0dro, WDYT?..

(And, BTW, if to decide on what to depend for database access, I'd prefer much more lightweight Sequel library.)

v0dro:
@zverok is correct. We should not add extra dependencies.

For this particular issue I think you can get away with just checking if the extension of the file name is `.db`. If that is the case, you can a library for loading the database from the file.

Daru currently uses `dbi` and `sqlite3` as development dependencies. `dbi` can use different database adapters for loading data. If you can use `dbi` for loading from file that would be great so that the user only needs to install the appropriate adapter.

dshvimer:
I am going to see what can be done with DBI. If that does not work, I think between sqlite3 and [this dbf gem](https://github.com/infused/dbf) we can have a large portion of database file types covered. 

In regards to a companion gem, how big would ```daru-db``` have to be? Would it need to support relational, no sql, and file based db's?

zverok:
As I've already wrote in PR,

> you should first clarify its scope (by investigating "which formats of binary DB files are available on popular scientfic/governmental/business datasources), and only then the coding attempt.

E.g., it is a necessary step to check:
* what are most common formats database files are published?
* what database files format are supported by popular competitors (pandas and R's database-handling libraries)?

We'd be really happy if you could spend some (not much, yet decent) effort to answer and dicsuss those questions.

dshvimer:
[...]

PR #301, opened 2017-02-05
Can open SQL databases by specifying file name 

dshvimer:
Let me know if anything looks unsightly

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -63,7 +63,7 @@ module Daru
 
       # Read a database query and returns a Dataset
       #
-      # @param dbh [DBI::DatabaseHandle, String] A DBI connection to be used to run the query OR Path to a SQlite3 database.
+      # @param dbh [DBI::DatabaseHandle, String] A DBI connection OR Path to a SQlite3 database.
       # @param query [String] The query to be executed
       #
       # @return A dataframe containing the data resulting from the query
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -63,7 +63,7 @@ module Daru
 
       # Read a database query and returns a Dataset
       #
-      # @param dbh [DBI::DatabaseHandle] A DBI connection to be used to run the query
+      # @param dbh [DBI::DatabaseHandle, String] A DBI connection to be used to run the query OR Path to a SQlite3 database.
       # @param query [String] The query to be executed
       #
       # @return A dataframe containing the data resulting from the query
@@ -72,6 +72,11 @@ module Daru
       #
       #  dbh = DBI.connect("DBI:Mysql:database:localhost", "user", "password")
       #  Daru::DataFrame.from_sql(dbh, "SELECT * FROM test")
+      #
+      #  #Alternatively
+      #
+      #  require 'dbi'
+      #  Daru::DataFrame.from_sql("path/to/sqlite.db", "SELECT * FROM test")
       def from_sql dbh, query
         Daru::IO.from_sql dbh, query
       end
--- a/lib/daru/io/sql_data_source.rb
+++ b/lib/daru/io/sql_data_source.rb
@@ -73,6 +73,8 @@ module Daru
         query = String.try_convert(query) or
           raise ArgumentError, "Query must be a string, #{query.class} received"
 
+        db = attempt_sqlite3_connection(db) if db.is_a?(String) && Pathname(db).exist?
+
         case db
         when DBI::DatabaseHandle
           DbiAdapter.new(db, query)
@@ -81,6 +83,14 @@ module Daru
         else
           raise ArgumentError, "Unknown database adapter type #{db.class}"
         end
+      end
+
+      def attempt_sqlite3_connection(db)
+        DBI.connect("DBI:SQLite3:#{db}")
+      rescue SQLite3::NotADatabaseException
+        raise ArgumentError, "Expected #{db} to point to a SQLite3 database"
+      rescue NameError
+        raise NameError, "In order to establish a connection to #{db}, please require 'dbi'"
       end
     end
   end
--- a/spec/io/sql_data_source_spec.rb
+++ b/spec/io/sql_data_source_spec.rb
@@ -62,5 +62,22 @@ RSpec.describe Daru::IO::SqlDataSource do
         }.to raise_error(ArgumentError)
       end
     end
+
+    context 'with path to sqlite3 file' do
+      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(db_name, query) }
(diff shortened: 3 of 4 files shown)


lib/daru/io/sql_data_source.rb
```diff
@@ -78,10 +97,19 @@ def init_adapter(db, query)
           DbiAdapter.new(db, query)
         when ActiveRecord::ConnectionAdapters::AbstractAdapter
           ActiveRecordConnectionAdapter.new(db, query)
+        when String
```
zverok: I am not sure that `String` is enough condition to use SQLite adapter. If the API accepts strings at all, somebody could (rightfully) expect that `make_dataframe('mysql://user:password@database')` should also work.

spec/io/sql_data_source_spec.rb
```diff
@@ -62,5 +66,19 @@
...
     end
+
+    context 'with database file path' do
+      it 'returns a dataframe' do
+        result = Daru::IO::SqlDataSource.make_dataframe(db_file, query)
+        expect(result).to be_a(Daru::DataFrame)
+        expect(result.row[0][:id]).to eq(1)
+        expect(result.row[0][:name]).to eq('Alex')
+      end
+      it 'raises ArgumentError' do
```
zverok: It is better to have in separate context. In `--document` mode of RSpec, lines 70-82 will generate strangely
```
with database file path
  it returns a dataframe
  it raises ArgumentError
```
...which is not a good description of what's going on.

lib/daru/io/sql_data_source.rb
dshvimer: Thank you for the feedback. I think that I will instead implement a method that looks at the db argument and does something intelligent with it before passing it on to the case statement. This way we have a hook for implementing connection strings or files etc.

dshvimer:
@zverok 
This PR is in response to [#298](https://github.com/SciRuby/daru/issues/298)

I will refactor the work later tonight and try to simplify things.

zverok:
> This PR is in response to #298

Well, I am not sure that SQLite is the only type of database that could be distributed as a file. At least I've seen some governmental data to be distributed in [DBFs](https://en.wikipedia.org/wiki/.dbf) and MS Access, for example. I am not sure it is a frequent case, but not sure about SQLite databases either. So, I feel like to solve #298 you should first clarify its scope (by investigating "which formats of binary DB files are available on popular scientfic/governmental/business datasources), and only then the coding attempt.

dshvimer:
@zverok I refactored what I had. 
Changes:
1. I am reusing the DBI adapter instead of implementing a SQLite3 adapter. 
2. Stronger condition for whether a file path is being passed in.

Let me know how it looks from your perspective.

In regards to DBF's: I am wrestling with support for SQL queries. It seems that [this gem](https://github.com/infused/dbf) does not support them. Thoughts?

spec/io/sql_data_source_spec.rb
```diff
@@ -62,5 +62,24 @@
         }.to raise_error(ArgumentError)
       end
     end
+
+    context 'with path to sqlite3 file' do
+      it 'returns a dataframe' do
+        result = Daru::IO::SqlDataSource.make_dataframe(db_name, query)
```
v0dro: Will passing the file name also work with `DataFrame.from_sql`?

dshvimer: It does work. I can write a test but it seems redundant. Should I move this test into the dataframe_spec.rb?

v0dro: No thats okay. Can you change the docs of `from_sql` to reflect the fact that passing a string also works now?

dshvimer: Updated the docs, but I was unsure if this was the correct way to do so:
https://monosnap.com/file/fgILJV5wPNLvpbeXEsIs7f0GRdLYxq

lib/daru/dataframe.rb
```diff
@@ -64,6 +64,7 @@ def from_excel path, opts={}, &block
       # Read a database query and returns a Dataset
       #
       # @param dbh [DBI::DatabaseHandle] A DBI connection to be used to run the query
```
v0dro: This should be specified as `@param dbh [DBI::DatabaseHandle, String] A DBI connection to be used to run the query OR Path to a SQlite3 database.`

dshvimer: Made the correction

spec/io/sql_data_source_spec.rb
```diff
@@ -62,5 +62,25 @@
...
+        # result = Daru::DataFrame.from_sql(db_name, query)
+        expect(result).to be_a(Daru::DataFrame)
+        expect(result.nrows).to eq(2)
+        expect(result.row[0][:id]).to eq(1)
+        expect(result.row[0][:name]).to eq('Homer')
+      end
```
zverok: Let's make a proper "modern" specs for a new functionality (like http://betterspecs.org/ modern). With subject and `its(...)` and stuff. For entire daru codebase it is a long road to better specs, but for new ones, it is reasonable to have them in decent form from the very beginning.

dshvimer: Just checking in to let you know that I saw this. I'll have a chance to take care of this at some point tomorrow.

dshvimer:
@zverok I made the changes to the test cases as you recommended. Let me if anything else needs to be done.

v0dro:
The Travis build is failing. Rubocop.

dshvimer:
Sorry about that. Rubocop is now all green

v0dro:
Good to go from my side. @zverok needs to approve and merge.

spec/io/sql_data_source_spec.rb
```diff
@@ -62,5 +62,22 @@
         }.to raise_error(ArgumentError)
       end
     end
+
+    context 'with path to sqlite3 file' do
+      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(db_name, query) }
+      it { is_expected.to be_a(Daru::DataFrame) }
+      it { expect(df.nrows).to eq 2 }
+      it { expect(df.row[0][:id]).to eq 1 }
+      it { expect(df.row[0][:name]).to eq 'Homer' }
```
zverok: Well, I'd say better way to write this would be:
```ruby
its(:nrows) { is_expected.to eq 2 }
its(:'rows.first') { is_expected.to eq {id: 1, name: 'Homer'} }
```

But it is a neatpeaking. I thinkg "enchance specs quality would be separated task, so we are good here.

dshvimer: @zverok I did not know that it could be written like that. Is it reasonable to open an issue to begin updating the tests throughout the library. I would be up to doing that next since I need to let the new syntax sink in.

Issue #246, opened 2016-09-07
where clause should support missing data.


v0dro:
When `#where` is called on a Vector/DataFrame, it should not fail if the data contains missing data.

lokeshh:
Could you share a case where it fails?

v0dro:
Here you go:
```
2.2.1 :011 >   v= Daru::Vector.new([1,2,3,Float::NAN, nil])
 => #<Daru::Vector(5)>
   0   1
   1   2
   2   3
   3 NaN
   4 nil 
2.2.1 :012 > v.where(v.lt(4))
NoMethodError: undefined method `<' for nil:NilClass
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/core/query.rb:42:in `block in apply_scalar_operator'
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/accessors/array_wrapper.rb:11:in `each'
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/accessors/array_wrapper.rb:11:in `each'
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/core/query.rb:42:in `map'
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/core/query.rb:42:in `apply_scalar_operator'
	from /home/sameer/.rvm/gems/ruby-2.2.1/gems/daru-0.1.4.1/lib/daru/vector.rb:353:in `block (2 levels) in <class:Vector>'
	from (irb):12
	from /home/sameer/.rvm/rubies/ruby-2.2.1/bin/irb:11:in `<main>'
```

athityakumar:
If this issue hasn't been fixed yet, I can work on this - if assigned.

Ping @lokeshh @v0dro

v0dro:
The expectation is that missing data (as in `nil` or `Float::NaN`) should be supported by the `where` clause irrespective of the kind of data in the vector/dataframe.

You can check out the README for more example of `where`.

PR #302, opened 2017-02-06
Adds support to missing data for where clause

athityakumar:
Fixes issue #246. Sample output for the building pass is shown below.

```
2.3.1 :002 > v= Daru::Vector.new([1,2,3,Float::NAN, nil])
 => #<Daru::Vector(5)>
   0   1
   1   2
   2   3
   3 NaN
   4 nil 
2.3.1 :003 > v.where(v.lt(4))

 => #<Daru::Vector(3)>
   0   1
   1   2
   2   3 
2.3.1 :004 > v.where(v.lt(2))

 => #<Daru::Vector(1)>
   0   1 
2.3.1 :005 > v.where(v.lt(3))

 => #<Daru::Vector(2)>
   0   1
   1   2 
2.3.1 :006 > v.where(v.gt(1))

 => #<Daru::Vector(2)>
   1   2
   2   3 
```

Ping @v0dro @zverok @lokeshh - Please review this PR when you're free. :smile:

### DIFF
--- a/lib/daru/core/query.rb
+++ b/lib/daru/core/query.rb
@@ -39,7 +39,7 @@ module Daru
 
       class << self
         def apply_scalar_operator operator, data, other
-          BoolArray.new data.map { |d| !!d.send(operator, other) }
+          BoolArray.new data.map { |d| !!d.send(operator, other) unless d.nil? }
         end
 
         def apply_vector_operator operator, vector, other
--- a/lib/daru/core/query.rb
+++ b/lib/daru/core/query.rb
@@ -39,7 +39,7 @@ module Daru
 
       class << self
         def apply_scalar_operator operator, data, other
-          BoolArray.new data.map { |d| !!d.send(operator, other) unless d.nil? }
+          BoolArray.new data.map { |d| !!d.send(operator, other) unless MISSING_VALUES.include? d }
         end
 
         def apply_vector_operator operator, vector, other
--- a/spec/core/query_spec.rb
+++ b/spec/core/query_spec.rb
@@ -267,10 +267,9 @@ describe "Arel-like syntax" do
           ).to eq(answer)
       end
 
-      let(:dv1) { Daru::Vector.new([1,11,32,Float::NAN,nil]) }
-      let(:dv2) { Daru::Vector.new([1,11]) }
+      let(:dv) { Daru::Vector.new([1,11,32,Float::NAN,nil]) }
       it "handles empty data" do
-        expect(dv1.where(dv1.lt(14))).to eq(dv2)
+        expect(dv.where(dv.lt(14))).to eq(Daru::Vector.new([1,11]))
       end
 
       it "does not give SystemStackError" do
--- a/spec/core/query_spec.rb
+++ b/spec/core/query_spec.rb
@@ -240,9 +240,9 @@ describe "Arel-like syntax" do
     context Daru::DataFrame do
       before do
         @df = Daru::DataFrame.new({
-          number: [1,2,3,4,5,6],
-          sym: [:one, :two, :three, :four, :five, :six],
-          names: ['sameer', 'john', 'james', 'omisha', 'priyanka', 'shravan']
+          number: [1,2,3,4,5,6,nil],
+          sym: [:one, :two, :three, :four, :five, :six, :seven],
+          names: ['sameer', 'john', 'james', 'omisha', 'priyanka', 'shravan', Float::NAN.to_s]
         })
       end
 
@@ -267,6 +267,16 @@ describe "Arel-like syntax" do
           ).to eq(answer)
       end
 
+      it "handles empty data" do
+        answer = Daru::DataFrame.new({
+          number: [nil],
+          sym: [:seven],
+          names: [Float::NAN.to_s]
+          }, index: Daru::Index.new([6])
+        )
+        expect(@df.where(@df[:sym].eq(:seven))).to eq(answer)
+      end
+
       it "does not give SystemStackError" do
         v = Daru::Vector.new [1]*300_000
         expect { v.where v.eq(1) }.not_to raise_error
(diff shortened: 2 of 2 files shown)


athityakumar:
Hmm, Travis CI takes "petty" C offenses pretty seriously. No offense, Travis CI. :wink:

athityakumar:
Tests, as in using Travis CI or `test-unit` gem?

zverok:
@athityakumar I've meant specs -- there are ton of them & you can run them locally by just `rspec` or `rspec path/to/your/spec.rb`

v0dro:
@athityakumar  I suggest you first go through the CONTRIBUTING file and understand our processes properly before submitting PRs. We're all volunteers and it saves our time if you abide by some formal processes.

v0dro:
Also, submitting a PR without tests is not acceptable. Especially if you want to apply for GSOC.

lib/daru/core/query.rb
```diff
@@ -39,7 +39,7 @@ def inspect
 
       class << self
         def apply_scalar_operator operator, data, other
-          BoolArray.new data.map { |d| !!d.send(operator, other) }
+          BoolArray.new data.map { |d| !!d.send(operator, other) unless d.nil? }
```
v0dro: Missing data is not just `nil`. What if the `Vector` contains strings combined with a `Float::NaN`? This will fail in that case.

v0dro: See this: https://github.com/SciRuby/daru/blob/master/lib/daru.rb#L33

athityakumar:
@v0dro : Thanks for mentioning about the `CONTRIBUTING.md` file - I just read through it. I'll add the tests in a day, after including the other missing data types.

athityakumar:
@v0dro  @zverok  : The PR is ready for another review. 

### Updates - 

(1) Thanks for the `rspec` and `robocup` references. They really made testing the changes simpler. 
(2) Support added for other missing data types like `Float:NAN` and `Float:NAN in string`.
(3) Tests have been written for `Daru::Vector#where` and `Daru::DataFrame#where`.

### Clarification required - 
Tests have been written for `Daru::Vector#where` and `Daru::DataFrame#where` as a simple single equality check. Should they be changed / new test be added for `SystemStackError` check? Or does the equality check do?

athityakumar:
@v0dro   @zverok   : The PR is ready for another review. 

### Updates - 

(1) Thanks for the `rspec` and `rubocop` references. They really made testing the changes simpler. 
(2) Support added for other missing data types like `Float:NAN` and `Float:NAN in string`.
(3) Tests have been written for `Daru::Vector#where` and `Daru::DataFrame#where`.

Any changes required?

spec/core/query_spec.rb
```diff
@@ -240,9 +240,9 @@
     context Daru::DataFrame do
       before do
         @df = Daru::DataFrame.new({
-          number: [1,2,3,4,5,6],
-          sym: [:one, :two, :three, :four, :five, :six],
-          names: ['sameer', 'john', 'james', 'omisha', 'priyanka', 'shravan']
+          number: [1,2,3,4,5,6,nil],
+          sym: [:one, :two, :three, :four, :five, :six, :seven],
+          names: ['sameer', 'john', 'james', 'omisha', 'priyanka', 'shravan', Float::NAN.to_s]
```
zverok: Why `Float::NAN.to_s` should be considered as a missing value?

spec/core/query_spec.rb
```diff
@@ -267,6 +267,16 @@
...
       end
 
+      it "handles empty data" do
+        answer = Daru::DataFrame.new({
+          number: [nil],
+          sym: [:seven],
+          names: [Float::NAN.to_s]
+          }, index: Daru::Index.new([6])
+        )
+        expect(@df.where(@df[:sym].eq(:seven))).to eq(answer)
```
zverok: There is an old saying "Don't believe test you've never seen failing". It means: you should first make sure this test is failing without your patch, then apply the patch, then check test is green. 
Hint: it doesn't check anything (on current daru master it will work).

zverok: Yep, think so.

spec/core/query_spec.rb
athityakumar: That was meant to be `Float::NAN` missing type as mentioned [here](https://github.com/SciRuby/daru/blob/master/lib/daru.rb#L33). I will change that in the next commit, along with other changes. :smile:

athityakumar:
@zverok : I've added the specific test that would have failed on the current master branch version of `daru`. As suggested, I've also added my test variables in `let` - however, it does seem a bit different / inconsistent with the rest of the tests in the `spec/core/query_spec.rb` file, as they don't use `let`.

zverok:
>  however, it does seem a bit different / inconsistent with the rest of the tests in the spec/core/query_spec.rb file, as they don't use let format.

Yes, part of the specs are just old and written without much attention. We want to have them rewritten in modern style with time, yet it was not done yet.

spec/core/query_spec.rb
```diff
@@ -267,6 +267,12 @@
           ).to eq(answer)
       end
 
+      let(:dv1) { Daru::Vector.new([1,11,32,Float::NAN,nil]) }
+      let(:dv2) { Daru::Vector.new([1,11]) }
+      it "handles empty data" do
+        expect(dv1.where(dv1.lt(14))).to eq(dv2)
```
zverok: I am not sure you need to create separate `dv`, `expect(dv1.where(dv1.lt(14))).to eq(Daru::Vector.new([1,11]))` should be enough.

athityakumar:
@v0dro @zverok : Made the required changes. :smile:

lib/daru/core/query.rb
lokeshh: @v0dro `Float::NaN` I think are only used in numerical domain. I don't think they ever occur with strings or any other object except numbers.

Issue #158, opened 2016-06-04
provide min/max with blocks


lokeshh:
`Daru::Vector#max` doesn't accept a block currently. For example with `Array#max` one can do:

``` ruby
>> dv = Daru::Vector.new ['c', 'bb', 'aaa']
>> dv.max { |i| i.length }
=> "c"     # Wrong because it hasn't been implemented yet
>> ['c', 'bb', 'aaa'].max { |i| i.length }
=> "aaa"

```

lokeshh:
Also I think we should have something like Pandas `argmax()` and `argmin()`, they return the index value at which the value is maximum.

v0dro:
Let's call those methods `index_of_max` and `index_of_min`. And the block is a good idea.

athityakumar:
I believe this issue can be closed now, as it has been taken care of by PR #303 . :smile:

Issue #206, opened 2016-08-02
Reset index function


dansbits:
Is there a function on the data frame similar to the pandas `reset_index` method for adding the values in the index to the dataframe. I often find myself wanting to re-incorporate the index back into my dataframe after using group_by.

I don't see anything and wanted to check before trying to add it.

v0dro:
Should be a good addition. How will the new vector that the index will create be named? Also, the new index that the dataframe will get will be numbered from zero right?

PR #303, opened 2017-02-06
Add new functions to Vector: max, min, index_of_max, index_of_min, max_by, min_by, index_of_max_by, index_of_min_by

athityakumar:
This Pull Request aims to add functions into `Daru::Vector` and `Daru::DataFrame` classes. 

- [x] `Daru::Vector#max` (for issue #158 )
- [x] `Daru::Vector#min` (for issue #158 )
- [x] `Daru::Vector#index_of_max` (for issue #158) 
- [x] `Daru::Vector#index_of_min` (for issue #158)
- [x] `Daru::Vector#max_by` (as alias, from this PR)
- [x] `Daru::Vector#min_by` (as alias, from this PR)
- [x] `Daru::Vector#index_of_max_by` (as alias, from this PR)
- [x] `Daru::Vector#index_of_min_by` (as alias, from this PR)

Do test the implemented functions. And suggest more functions to add.

### DIFF
--- a/example
+++ b/example
@@ -1,5 +1,0 @@
-require 'daru'
-dv = Daru::Vector.new ['c', 'bb', 'aaa']
-dv.max { |i| i.length }
-dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
-dv.index_of_max { |i| i.size }
--- a/example
+++ b/example
@@ -1,0 +1,5 @@
+require 'daru'
+dv = Daru::Vector.new ['c', 'bb', 'aaa']
+dv.max { |i| i.length }
+dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
+dv.index_of_max { |i| i.size }
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -560,6 +560,11 @@ module Daru
       self
     end
 
+    def reset_index!
+      @index = Daru::Index.new(Array.new(size) { |i| i })
+      self
+    end
+
     # Iterate over each index of the DataFrame.
     def each_index &block
       return to_enum(:each_index) unless block_given?
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -560,11 +560,6 @@ module Daru
       self
     end
 
-    def reset_index!
-      @index = Daru::Index.new(Array.new(size) { |i| i })
-      self
-    end
-
     # Iterate over each index of the DataFrame.
     def each_index &block
       return to_enum(:each_index) unless block_given?
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -72,14 +72,7 @@ module Daru
         #   to returning only the maximum number but passing *:vector* will return
         #   a Daru::Vector with the index of the corresponding maximum value.
         def max
-          if block_given?
-            arr = []
-            @data.data.each do |d|
-              arr.push(yield(d))
-            end
-          else
-            arr = @data.data
-          end
+          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
           @data.data[arr.index(arr.max)]
         rescue
           @data.to_a.max
@@ -87,14 +80,7 @@ module Daru
 
         # Get index of the above maximum element
         def index_of_max
-          if block_given?
-            arr = []
-            @data.data.each do |d|
-              arr.push(yield(d))
-            end
-          else
-            arr = @data.data
-          end
+          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
           @index.to_a[arr.index(arr.max)]
         end
 
@@ -104,14 +90,7 @@ module Daru
         #   to returning only the minimum number but passing *:vector* will return
         #   a Daru::Vector with the index of the corresponding minimum value.
         def min
-          if block_given?
-            arr = []
-            @data.data.each do |d|
-              arr.push(yield(d))
-            end
-          else
-            arr = @data.data
-          end
+          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
           @data.data[arr.index(arr.min)]
         rescue
           @data.to_a.min
@@ -119,14 +98,7 @@ module Daru
 
         # Get index of the above minimum element
         def index_of_min
-          if block_given?
-            arr = []
-            @data.data.each do |d|
-              arr.push(yield(d))
-            end
-          else
-            arr = @data.data
-          end
+          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
           @index.to_a[arr.index(arr.min)]
         end
 
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -71,16 +71,16 @@ module Daru
         # @param return_type [Symbol] Data type of the returned value. Defaults
         #   to returning only the maximum number but passing *:vector* will return
         #   a Daru::Vector with the index of the corresponding maximum value.
-        def max
-          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
+        def max(&block)
+          arr = block_given? ? @data.data.map(&block) : @data.data
           @data.data[arr.index(arr.max)]
         rescue
           @data.to_a.max
         end
 
         # Get index of the above maximum element
-        def index_of_max
-          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
+        def index_of_max(&block)
+          arr = block_given? ? @data.data.map(&block) : @data.data
           @index.to_a[arr.index(arr.max)]
         end
 
@@ -89,16 +89,16 @@ module Daru
         # @param return_type [Symbol] Data type of the returned value. Defaults
         #   to returning only the minimum number but passing *:vector* will return
         #   a Daru::Vector with the index of the corresponding minimum value.
-        def min
-          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
+        def min(&block)
+          arr = block_given? ? @data.data.map(&block) : @data.data
           @data.data[arr.index(arr.min)]
         rescue
           @data.to_a.min
         end
 
         # Get index of the above minimum element
-        def index_of_min
-          arr = block_given? ? @data.data.map { |d| yield(d) } : @data.data
+        def index_of_min(&block)
+          arr = block_given? ? @data.data.map(&block) : @data.data
           @index.to_a[arr.index(arr.min)]
         end
 
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -80,24 +80,23 @@ module Daru
         #   dv.max
         #   #=> "Tyrion"
         #
-        #   dv.max { |i| i.size }
-        #   #=> "Jon Starkgaryen"
+        #   dv.max(2) { |a,b| a.size <=> b.size }
+        #   #=> ["Jon Starkgaryen","Daenerys"]
         #
         #   dv.max(2) { |i| i.size }
         #   #=> ["Jon Starkgaryen","Daenerys"]
-        def max(size=nil, &block)
+        def max(*size, &block)
+          size = size.count.zero? ? nil : size[0]
           data = @data.data.to_a
           if block_given?
-            data.max(size,&block)
+            if block.parameters.count == 1 # Object block like { |x| x.size }
+              data.max_by(size,&block)
+            else # Comparative block like { |a,b| a.size <=> b.size }
+              data.max(size,&block)
+            end
           else
             data.max(size)
           end
-          # values = if block_given?
-          #            data.values_at(*data.map { |d| yield(d) }.to_a.each_with_index.max(size).map(&:last))
-          #          else
-          #            data.max(size)
-          #          end
-          # size_not_given ? values[0] : values
         end
 
         # Returns the index of the maximum value present in the vector.
@@ -114,10 +113,17 @@ module Daru
         #   dv.index_of_max
         #   #=> :t
         #
-        #   dv.index_of_max { |i| i.size }
-        #   #=> :j
-        def index_of_max(&block)
-          block_given? ? @index.to_a[@data.data.map(&block).to_a.each_with_index.max.last] : @index.max
+        #   dv.index_of_max(2) { |a,b| a.size <=> b.size }
+        #   #=> [:j, :d]
+        #
+        #   dv.max(2) { |i| i.size }
+        #   #=> [:j, :d]
+        def index_of_max(*size,&block)
+          size = size.count.zero? ? nil : size[0]
+          data = @data.data.to_a
+          indx = @index.to_a
+          vals = max(size,&block)
+          vals.is_a?(Array) ? (vals.map { |x| indx[data.index(x)] }) : indx[data.index(vals)]
         end
 
         # Returns the minimum value present in the vector.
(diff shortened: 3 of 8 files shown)


lokeshh:
You can add support for blocks to `index_of_max` and `index_of_min`.

For example

```ruby
[1] pry(main)> dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
=> #<Daru::Vector(3)>
   a   a
   b  bb
   c ccc
[2] pry(main)> dv.index_of_max { |i| i.size }
# This should return :c
```

v0dro:
Write tests for all new functions.

athityakumar:
@v0dro : Sure, I will include the tests for these functions first in a day or two, before adding other new functions - and let you know for another review of this PR.

athityakumar:
@v0dro @lokeshh : Thanks for your suggestions - the PR is ready for another review. :smile: 

### Updates 

(1) Functions `index_of_min` , `index_of_max` ,`max` , `min` are all placed inside blocks - hence, enabling features like 

```
dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
=> #<Daru::Vector(3)>
   a   a
   b  bb
   c ccc
dv.index_of_max { |i| i.size }
:c
``` 

(2) Tests have been added to all the 5 functions.

### Help required

After writing the `max` function, the tests of the following functions failed - possibly due to a function name clash : 

```
rspec './spec/maths/statistics/vector_spec.rb[1:2:11:1]' # Daru::Vector gsl #max returns the max value
rspec './spec/maths/statistics/vector_spec.rb[1:2:12:1]' # Daru::Vector gsl #min returns the min value
rspec './spec/maths/statistics/vector_spec.rb[1:2:17:1]' # Daru::Vector gsl #describe generates count, mean, std, min and max of vectors in one shot
```

I've temporarily commented those tests. But how do we proceed with this? Also, are there any other functions that need to be added?

lib/daru/dataframe.rb
```diff
@@ -560,6 +560,11 @@ def replace_values old_values, new_value
       self
     end
 
+    def reset_index!
+      @index = Daru::Index.new(Array.new(size) { |i| i })
+      self
+    end
```
zverok: A you sure it is what required by #206? Doesn't quite like pandas' [reset_index](https://github.com/SciRuby/daru/issues/206)

lib/daru/maths/statistics/vector.rb
```diff
@@ -75,19 +71,70 @@ def factors
         # @param return_type [Symbol] Data type of the returned value. Defaults
         #   to returning only the maximum number but passing *:vector* will return
         #   a Daru::Vector with the index of the corresponding maximum value.
-        def max return_type=:stored_type
-          max_value = @data.max
-          if return_type == :vector
-            Daru::Vector.new({index_of(max_value) => max_value}, name: @name, dtype: @dtype)
+        def max
```
zverok: Why have you removed `return_type` parameter?

lib/daru/maths/statistics/vector.rb
```diff
@@ -75,19 +71,70 @@ def factors
...
+          if block_given?
+            arr = []
+            @data.data.each do |d|
+              arr.push(yield(d))
+            end
```
zverok: Why don't you use `Array#map` here?

lib/daru/maths/statistics/vector.rb
```diff
@@ -75,19 +71,70 @@ def factors
...
+          @data.to_a.max
+        end
+
+        # Get index of the above maximum element
+        def index_of_max
+          if block_given?
+            arr = []
+            @data.data.each do |d|
+              arr.push(yield(d))
+            end
```
zverok: `Array#map`?

lib/daru/maths/statistics/vector.rb
```diff
@@ -75,19 +71,70 @@ def factors
...
+          end
+          @index.to_a[arr.index(arr.max)]
+        end
+
+        # Minimum element of the vector.
+        #
+        # @param return_type [Symbol] Data type of the returned value. Defaults
+        #   to returning only the minimum number but passing *:vector* will return
+        #   a Daru::Vector with the index of the corresponding minimum value.
+        def min
```
zverok: All the same notes as for `max`

lib/daru/maths/statistics/vector.rb
```diff
@@ -75,19 +71,70 @@ def factors
...
-            max_value
+            arr = @data.data
           end
+          @data.data[arr.index(arr.min)]
+        rescue
+          @data.to_a.min
+        end
+
+        # Get index of the above minimum element
+        def index_of_min
```
zverok: All the same notes as for `index_of_max`

spec/dataframe_spec.rb
```diff
@@ -3029,6 +3029,17 @@
...
+  context "#reset_index!" do
+    it "resets index array of dataframe" do
+      data = {x: [5,1,-6], y: [2,7,0], z: [4,3,8]}
+      idx = ["u","v","w"]
+      df1 = Daru::DataFrame.new(data, index: idx)     
+      df2 = Daru::DataFrame.new(data, index: Daru::Index.new(3))     
+      df1.reset_index!
+      expect(df1.index.to_a).to eq(df2.index.to_a)
+    end
+  end
```
zverok: I understand you are looking at other tests around, but we are trying to write proper tests for new functionality, with all modern RSpec approaches: http://betterspecs.org/ (E.g. `let`/`subject` and so on)

lib/daru/maths/statistics/vector.rb
athityakumar: @zverok : I had moved this functionality into `Daru::Vector#max_index` previously used to call `Daru::Vector#max` with `return_type` argument. And as it (`Daru::Vector#max_index`) is currently being used only from `Daru::DataFrame#max`, I thought it would be better to have this as `Daru::Vector#max_index` function - so that `Daru::Vector#max` could handle this with blocks :

```
dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
=> #<Daru::Vector(3)>
   a   a
   b  bb
   c ccc
dv.index_of_max { |i| i.size }
=> :c 
```

spec/dataframe_spec.rb
athityakumar: Thanks for this resource. I'll fix this by tomorrow. :smile:

lib/daru/dataframe.rb
athityakumar: Seems like I have implemented a very basic form of pandas' `reset_index` - of just resetting the `index`. I went through the complete doc and found that there are lots of functionalities of `reset_index`like taking care of `MultiIndex`, `Row`, `Column` and even `Series`. I'd prefer to isolate this function `reset_index` from this particular PR and continue in the later contributions as a separate PR.

Thanks for pointing this out! :smile:

athityakumar:
@zverok : Any other changes to be done / functions to be added? If not, can I start working on the `from_html` Nokogiri module? :smile:

example
```diff
@@ -0,0 +1,5 @@
+require 'daru'
+dv = Daru::Vector.new ['c', 'bb', 'aaa']
+dv.max { |i| i.length }
+dv = Daru::Vector.new ['a', 'bb', 'ccc'], index: [:a, :b, :c]
+dv.index_of_max { |i| i.size }
```
zverok: Please remove this file.

spec/maths/statistics/vector_spec.rb
```diff
@@ -6,12 +6,46 @@
...
         it "calculates mean" do
           expect(@dv.mean).to eq(282.2)
         end
       end
 
+
+      context "#max" do
+        it "gives the max of a vector" do
+          expect(dv2.max).to eq("tyrion lannister")
+          expect(dv2.max {|i| i.size}).to eq("tyrion lannister")
```
zverok: 1) `dv2.max(&:size)`. 

2) Typically, good specs contain one example per `it`. E.g.
```ruby
context '#max' do
  it 'returns max value' do
    expect(dv2.max).to eq("tyrion lannister")
  end

  it 'allows to provide block' do 
    # ...
  end
end
```

3) It would be better if the second example (with block) return something else than the first. (Now the second example will be "green" even if provided block just ignored inside `max`)

spec/maths/statistics/vector_spec.rb
```diff
@@ -6,12 +6,46 @@
...
+        end
+      end
+
+      context "#index_of_max" do
+        it "gives the index of max of a vector" do
+          expect(dv3.index_of_max).to eq(:tyrion)
+          expect(dv4.index_of_max {|i| -i}).to eq(2)
```
zverok: I'd say it will be better to show difference in behavior on a same dataframe.

athityakumar:
@zverok : Thanks for mentioning about the `YARD` resource. It's the first time I used, and it was just simplistically awesome. :smile: 

Any other changes to be made?

lib/daru/maths/statistics/vector.rb
```diff
@@ -70,24 +66,99 @@ def factors
...
+        #   dv.max
+        #   #=> "Tyrion"
+        #
+        #   dv.max { |i| i.size }
+        #   #=> "Jon Starkgaryen"
+        def max(&block)
+          arr = block_given? ? @data.data.map(&block) : @data.data
+          @data.data[arr.index(arr.max)]
+        rescue
+          @data.to_a.max
```
v0dro: Since a `Daru::Accessors::ArrayWrapper` class (which `@data` is an instance of) already has  a method called `max`.  (https://github.com/SciRuby/daru/blob/master/lib/daru/accessors/array_wrapper.rb#L89)

Also, why the `rescue`?

v0dro: Since a `Daru::Accessors::ArrayWrapper` class (which `@data` is an instance of) already has  a method called `max`.  (https://github.com/SciRuby/daru/blob/master/lib/daru/accessors/array_wrapper.rb#L89)

Also, why the `rescue`?

athityakumar:
@zverok @v0dro : I've pushed the changes as suggested above. :smile:

athityakumar:
I still have one more commit to push (reverting the changes that I earlier died in `dpescs/vector_spec.rb`) - however, the Travis CI build will anyway fail unless PR #306 is merged. Hence, I'll wait before pushing this commit. :smile:

athityakumar:
@zverok @v0dro : PR is ready for yet another review. :smile:

zverok:
Well, now when the code has decent quality, I have more conceptual question. Most of the time, we are trying to mimic/be close to Ruby's standard classes. Now, if you'll look at [Enumerable](http://ruby-doc.org/core-2.4.0/Enumerable.html)'s docs, you'll find
* `max`
* `max(n)`
* `max(n) { |a, b| ...`
* `max_by { |val| `
* `max_by(n) { |val| `

@v0dro, don't you think we should repeat this behavior? It looks somewhat reasonable for me.

athityakumar:
Sure. I have my mid-semester exams going on currently (till 25th Feb) - I'll push the changes after that, and let you know for the next PR review. :smile:

v0dro:
@athityakumar I've been monitoring your PRs and you are consistently requiring a lot of spoon feeding on part of @zverok to push you in the right direction.

If you want to apply as a GSOC student, we expect you to be an independent programmer and not make mistakes like renaming variables to something unreadable because of a Rubocop offense.

GSOC is for students who can work independently and turn for guidance only when they cannot figure out how to solve a problem. Helping you write better code after every other commit you push to a PR is not a mentor's responsibility. Just because @zverok is helpful does not mean he has a lot of time. He is not your professor.  All of us are volunteers and do this in our spare time. We don't have time to push you in the right direction after every other commit.

Carefully review your code yourself before you push it to a PR next time, or if its a WIP, mention so in a comment.

athityakumar:
> If you want to apply as a GSOC student, we expect you to be an independent programmer and not make mistakes like renaming variables to something unreadable because of a Rubocop offense.
> Carefully review your code yourself before you push it to a PR next time, or if its a WIP, mention so in a comment.

Sure, I usually @mention all the reviewers when the PR is a ready for a review. That commit (in which renaming variables to something unreadable because of a Rubocop offense) was an intermediate one - which I pushed to check for build reports on different ruby versions. And I accept that it was a bad call on my part to make the code unreadable - but this is something that happens very rarely.

> All of us are volunteers and do this in our spare time. We don't have time to push you in the right direction after every other commit.

Acknowledged.

> GSOC is for students who can work independently and turn for guidance only when they cannot figure out how to solve a problem. Helping you write better code after every other commit you push to a PR is not a mentor's responsibility. Just because @zverok is helpful does not mean he has a lot of time. 

I thought I was building a better rapport with the mentor, but I guess I didn't see it from the mentor's perspective - sorry for taking your time and thanks for helping, @zverok.

I'll make sure to review the code before pushing to the PR next time on wards. :+1:

v0dro:
Great! I appreciate that you're taking feedback positively.

zverok:
@athityakumar well I acknowledge I am forcing myself into "intermediary" commits; though, I somehow feel obliged to, just to make sure things go the right way. For me, the problem I can identify is your PRs is that you sometimes don't spend enough time to investigate "how the similar things are already done in daru" or "how this kind of things are typically done in good Ruby code". I don't think it is a fatal problem though, yet I still suggest to approach things with "something like that was probably done several times, how _they_ are usually doing it?" mindset. It is the only way to became an author of state-of-the-art solutions once ;)

athityakumar:
@v0dro @zverok - I require a small clarification. The [ruby doc](http://ruby-doc.org/core-1.9.3/Enumerable.html#method-i-max) specifies that `max` and `min` take a comparative block (like `|a,b| a <=> b`) whereas `max_by` and `min_by` take a single object in block (like `|x| x.size`).

Currently in the PR, `min` and `max` functions support the single object block rather than comparative block. That is, our `min` and `max` currently work like the inbuilt `min_by` and `max_by`. Should I change these functions accordingly to make these functions in `Daru::Vector` similar to that of Ruby's inbuilt enum?

zverok:
> The ruby doc specifies that max and min take a comparative block (like |a,b| a <=> b) whereas max_by and min_by take a single object in block (like |x| x.size).

Well, I pointed at that fact [above](https://github.com/SciRuby/daru/pull/303#issuecomment-280871341). Yet I have somewhat ambivalent feelings about this. As far as I can remember, Ruby's `max` vs `max_by` is "historical" (first it was only `max`, then somebody invented `max_by`). I have no clue if anybody uses `max` with block, as `max_by` always more superior/convenient. 

So, I can find good arguments for both ideas:
1. make the interface exactly as Array's
2. cleanup it a bit, having only `max` with 1-arg block (and probably `max_by` as just its alias, for those who will expect it to exist).

Currently, I like (2) a bit more (that's why I haven't forced `max`/`max_by` distinction, though pointed on it).

@v0dro @lokeshh WDYT?

lokeshh:
It's a hard choice. I would choose first option.

athityakumar:
I was thinking if we could use `max` to be compatible with both types of blocks by knowing the number of variables passed on (with `block.parameters`).
Source : http://stackoverflow.com/questions/31812486/check-how-many-parameters-block-being-passed-has-and-then-invoke-yield-with-dif

This way, `max` will support both `{|x| x.size}` (by calling `Array#max_by`) and `{|a,b| b.size <=> a.size}` (by calling `Array#max`) - and `max_by` can be aliased to `max`.

zverok:
> I was thinking if we could use max to be compatible with both types of blocks by knowing the number of variables passed

OK, if you are sure, let's go this way. May be the most reasonable, while not the most easy one.

athityakumar:
@zverok - I've amended the changes as suggested above. Functions `Daru:Vector#max`, `Daru:Vector#min`, `Daru:Vector#index_of_max` and `Daru:Vector#index_of_min` now can take both comparative as well as object blocks. Functions `Daru:Vector#max_by`, `Daru:Vector#min_by`, `Daru:Vector#index_of_max_by`, `Daru:Vector#index_of_min_by` have been aliased to the above mentioned functions, respectively.

athityakumar:
@zverok @v0dro - I've re-wrote the spec descriptions, as suggested above.

zverok:
@v0dro I think we can merge it, WDYT?

athityakumar:
Ping @v0dro

PR #305, opened 2017-02-09
Deprecate use of reportbuilder in DataFrame summary

### LINKED ISSUE #221: Better DataFrame summary without reportbuilder

ananyo2012:
Fixes #221 

This replaces the use of `reportbuilder` for `DataFrame` summary and uses Daru `DataFrame` and `Vector` methods in place.

### DIFF
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -80,14 +80,12 @@ module Daru
   create_has_library :gruff
 end
 
-[['reportbuilder', '~>1.4'], ['spreadsheet', '~>1.1.1']].each do |lib|
-  begin
-    gem lib[0], lib[1]
-    require lib[0]
-  rescue LoadError
-    STDERR.puts "\nInstall the #{lib[0]} gem version #{lib[1]} for using"\
-    " #{lib[0]} functions."
-  end
+begin
+  gem 'spreadsheet', '~>1.1.1'
+  require 'spreadsheet'
+rescue LoadError
+  STDERR.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
+  ' spreadsheet functions.'
 end
 
 autoload :CSV, 'csv'
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1413,19 +1413,16 @@ module Daru
       Daru::DataFrame.new(arry, clone: cln, order: order, index: @index)
     end
 
-    # Generate a summary of this DataFrame with ReportBuilder.
-    def summary(method=:to_text)
-      ReportBuilder.new(no_title: true).add(self).send(method)
-    end
-
-    def report_building(b) # :nodoc: #
-      b.section(name: @name) do |g|
-        g.text "Number of rows: #{nrows}"
-        @vectors.each do |v|
-          g.text "Element:[#{v}]"
-          g.parse_element(self[v])
-        end
+    # Generate a summary of this DataFrame based on individual vectors in the DataFrame
+    # @return [String] String containing the summary of the DataFrame
+    def summary
+      summary = "= #{name}"
+      summary << "\n  Number of rows: #{nrows}"
+      @vectors.each do |v|
+        summary << "\n  Element:[#{v}]\n"
+        summary << self[v].summary(1)
       end
+      summary
     end
 
     # Sorts a dataframe (ascending/descending) in the given pripority sequence of
--- a/lib/daru/formatters/table.rb
+++ b/lib/daru/formatters/table.rb
@@ -10,6 +10,7 @@ module Daru
         @data = data || []
         @headers = (headers || []).to_a
         @row_headers = (row_headers || []).to_a
+        @row_headers = [''] * @data.to_a.size if @row_headers.empty?
       end
 
       DEFAULT_SPACING = 10
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -879,41 +879,66 @@ module Daru
       to_html
     end
 
-    # Create a summary of the Vector using Report Builder.
-    def summary(method=:to_text)
-      ReportBuilder.new(no_title: true).add(self).send(method)
-    end
-
-    # :nocov:
-    def report_building b # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-      b.section(name: name) do |s|
-        s.text "n :#{size}"
-        s.text "n valid:#{count_values(*Daru::MISSING_VALUES)}"
-        if @type == :object
-          s.text  "factors: #{factors.to_a.join(',')}"
-          s.text  "mode: #{mode}"
-
-          s.table(name: 'Distribution') do |t|
-            frequencies.sort_by(&:to_s).each do |k,v|
-              key = @index.include?(k) ? @index[k] : k
-              t.row [key, v, ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))]
-            end
-          end
-        end
-
-        s.text "median: #{median}" if @type==:numeric || @type==:numeric
-        if @type==:numeric
-          s.text 'mean: %0.4f' % mean
-          if sd
-            s.text 'std.dev.: %0.4f' % sd
-            s.text 'std.err.: %0.4f' % se
-            s.text 'skew: %0.4f' % skew
-            s.text 'kurtosis: %0.4f' % kurtosis
-          end
-        end
+    # Create a summary of the Vector
+    # @params [Fixnum] indent_level
+    # @return [String] String containing the summary of the Vector
+    # @example
+    #   dv = Daru::Vector.new [1, 2, 3]
+    #   puts dv.summary
+    #
+    #   =
+    #     n :3
+    #     n valid:0
+    #     median: 2
+    #     mean: 2.0000
+    #     std.dev.: 1.0000
+    #     std.err.: 0.5774
+    #     skew: 0.0000
+    #     kurtosis: -2.3333
+    def summary(indent_level=0)
+      non_missing = size - count_values(*Daru::MISSING_VALUES)
+      summary = '  =' * indent_level + "= #{name}" \
+                "\n  n :#{size}" \
+                "\n  non-missing:#{non_missing}"
+      case type
+      when :object
+        summary << object_summary
+      when :numeric
+        summary << numeric_summary
       end
+      summary = summary.split("\n").join("\n" + '  ' * indent_level)
+      summary
     end
-    # :nocov:
+
+    # Displays summary for an object type Vector
+    # @return [String] String containing object vector summary
+    def object_summary
+      nval = count_values(*Daru::MISSING_VALUES)
+      summary = "\n  factors: #{factors.to_a.join(',')}" \
+                "\n  mode: #{mode}" \
+                "\n  Distribution"
+
+      data = frequencies.sort.each_with_index.map do |v, k|
+        [k, v, '%0.2f%%' % ((nval.zero? ? 1 : v.quo(nval))*100)]
+      end
+
+      summary << Daru::Formatters::Table.format(data, {headers: ['']*3, row_headers: ['']*frequencies.size})
+      summary
+    end
+
+    # Displays summary for an numeric type Vector
+    # @return [String] String containing numeric vector summary
+    def numeric_summary
+      summary = "\n  median: #{median}" +
+                "\n  mean: %0.4f" % mean
+      if sd
+        summary << "\n  std.dev.: %0.4f" % sd +
+                   "\n  std.err.: %0.4f" % se +
+                   "\n  skew: %0.4f" % skew +
+                   "\n  kurtosis: %0.4f" % kurtosis
+      end
+      summary
+    end
 
     # Over rides original inspect for pretty printing in irb
     def inspect spacing=20, threshold=15
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -888,7 +888,7 @@ module Daru
     #
     #   =
     #     n :3
-    #     n valid:0
+    #     non-missing:3
     #     median: 2
     #     mean: 2.0000
     #     std.dev.: 1.0000
@@ -915,14 +915,14 @@ module Daru
     def object_summary
       nval = count_values(*Daru::MISSING_VALUES)
       summary = "\n  factors: #{factors.to_a.join(',')}" \
-                "\n  mode: #{mode}" \
+                "\n  mode: #{mode.to_a.join(',')}" \
                 "\n  Distribution"
 
       data = frequencies.sort.each_with_index.map do |v, k|
         [k, v, '%0.2f%%' % ((nval.zero? ? 1 : v.quo(nval))*100)]
       end
 
-      summary << Daru::Formatters::Table.format(data, {headers: ['']*3, row_headers: ['']*frequencies.size})
+      summary << Daru::Formatters::Table.format(data, headers: ['']*3, row_headers: ['']*frequencies.size)
       summary
     end
 
@@ -933,8 +933,11 @@ module Daru
                 "\n  mean: %0.4f" % mean
       if sd
         summary << "\n  std.dev.: %0.4f" % sd +
-                   "\n  std.err.: %0.4f" % se +
-                   "\n  skew: %0.4f" % skew +
+                   "\n  std.err.: %0.4f" % se
+      end
+
+      if count_values(*Daru::MISSING_VALUES).zero?
+        summary << "\n  skew: %0.4f" % skew +
                    "\n  kurtosis: %0.4f" % kurtosis
       end
       summary
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -906,8 +906,7 @@ module Daru
       when :numeric
         summary << numeric_summary
       end
-      summary = summary.split("\n").join("\n" + '  ' * indent_level)
-      summary
+      summary.split("\n").join("\n" + '  ' * indent_level)
     end
 
     # Displays summary for an object type Vector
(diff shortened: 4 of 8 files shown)


v0dro:
ReportBuilder does much more than just the things that you've specified in your summary. Please install the reportbuilder gem, read the docs and see what it is capable of. Your PR should encapsulate all of that.

Here's a previous PR about this with some useful comments: https://github.com/SciRuby/daru/pull/288

ananyo2012:
@v0dro Can you explain what is an `object` type in a Daru vector with an example? It is used in [this line](https://github.com/SciRuby/daru/blob/master/lib/daru/vector.rb#L892).

ananyo2012:
Ok I see any non-numeric element is treated as an object. But if I create an vector with a non-numeric element like `Daru::Vector.new([1, 2, "xyz"])` it shows an error if I try to get the summary. The error says
```
NoMethodError: undefined method `quo' for nil:NilClass
```

v0dro:
Object is any kind of data that is not numeric or categorical. Basically all sorts of Ruby objects like strings, hashes, numbers, etc. are `object`.

You will not understand the full capabilities of reportbuilder if you use it through daru. You'll need to use it as a standalone gem first and pass it data in a format that is supported by reportbuilder. 

See this: https://github.com/clbustos/reportbuilder

Specifically look at `ReportBuilder::Table` in the README. The idea is to mimic that functionality in daru.

ananyo2012:
@v0dro I read through the `reportbuilder` code and implemented a `table` method that works for this case. Can you have a look and give your reviews?

v0dro:
Give me a day.

zverok:
@ananyo2012 The big problem of taking this task is that current state (with use of reportbuilder) has very loose tests, they literally just check that "well, `#summary` outputs some string".

What I assume could be reasonable approach to the task is:
* write detailed test for a current behavior;
* replace current behavior with a new one (reportbuilder-less);
* make sure the test works again (probably with some minor formatting changes, there is no need to follow reportbuilder's output symbol-by-symbol).

Currently, to tell whether what you did is a right thing, each of the reviewers should just checkout your branch and check the output with a bare eyes... Which is not the best approach possible.

I'll also review your actual code shortly.

lib/daru/dataframe.rb
```diff
@@ -1408,19 +1408,15 @@ def only_numerics opts={}
...
-
-    def report_building(b) # :nodoc: #
-      b.section(name: @name) do |g|
-        g.text "Number of rows: #{nrows}"
-        @vectors.each do |v|
-          g.text "Element:[#{v}]"
-          g.parse_element(self[v])
-        end
-      end
+    # Generate a summary of this DataFrame
```
zverok: While updating the code, please consider improving the documentation also. We use [YARD](http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md) as our documentation tool. I am not saying you should write 10 paragraphs of code, but at least adding `@return` tag, and probably an example of this method's output, will be useful.

lib/daru/vector.rb
```diff
@@ -879,41 +879,56 @@ def to_s
...
-          s.text  "mode: #{mode}"
-
-          s.table(name: 'Distribution') do |t|
-            frequencies.sort_by(&:to_s).each do |k,v|
-              key = @index.include?(k) ? @index[k] : k
-              t.row [key, v, ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))]
-            end
-          end
+    # Create a summary of the Vector
+    def summary(indent_level=0) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
```
zverok: Well, I understand it was disabled previously (I mean rubocop cops), but it was disabled exactly because "it is old code... once we'll rewrite it". In a new code it is not advised to use the disables, better to restructure code accordingly to what Rubocop wants.

lib/daru/vector.rb
```diff
@@ -879,41 +879,56 @@ def to_s
...
+        # Display in table format
+        summary << '  ' * indent_level + '  Distribution'
+        key = []
+        val = []
+        nval = []
+        frequencies.sort_by(&:to_s).each do |k,v|
+          key << (@index.include?(k) ? @index[k] : k)
+          val << v
+          nval << ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))
         end
```
zverok: I believe lines 893-900 could be better represented with `.map`

lib/daru/vector.rb
```diff
@@ -879,41 +879,56 @@ def to_s
...
+      if @type == :object
+        summary << "\n  factors: #{factors.to_a.join(',')}"
+        summary << "\n  mode: #{mode}"
+
+        # Display in table format
+        summary << '  ' * indent_level + '  Distribution'
+        key = []
+        val = []
+        nval = []
+        frequencies.sort_by(&:to_s).each do |k,v|
```
zverok: I you sure that sorting by string representation of frequencies is most natural?

lib/daru/vector.rb
```diff
@@ -879,41 +879,56 @@ def to_s
...
-              key = @index.include?(k) ? @index[k] : k
-              t.row [key, v, ('%0.2f%%' % (v.quo(count_values(*Daru::MISSING_VALUES))*100))]
-            end
-          end
+    # Create a summary of the Vector
+    def summary(indent_level=0) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
+      summary = '  =' * indent_level + "= #{name}"
+      summary << "\n  n :#{size}"
+      summary << "\n  n valid:#{count_values(*Daru::MISSING_VALUES)}"
+      if @type == :object
```
zverok: We have `object?` method for this.

lib/daru/vector.rb
```diff
@@ -879,41 +879,56 @@ def to_s
...
-          if sd
-            s.text 'std.dev.: %0.4f' % sd
-            s.text 'std.err.: %0.4f' % se
-            s.text 'skew: %0.4f' % skew
-            s.text 'kurtosis: %0.4f' % kurtosis
-          end
+        summary << table(key, val, nval)
+      end
+
+      if @type == :numeric
```
zverok: Why not `elsif`? Vector can't be object and numeric at the same time. Or, in fact, `case @type` would be more reasonable.

lib/daru/vector.rb
ananyo2012: Actually I didn't really change the output of the summary. The code is kept same for that. Just removed the usage of reportbuilder for showing the output.

lib/daru/vector.rb
ananyo2012: Ok working on it

ananyo2012:
@zverok Thanks for the review. I am working on the changes requested. Till now what I did was just replace the usage of reportbuilder with a reportbuilder-less approach without changing the output code itself following strictly the reportbuilder output conventions. Will work on restucturing the code now but I think I will follow the reportbuilder output format as its pretty reasonable. 

And yes I will write detailed test for the same since this is a manual approach now and needs to have adequate tests.

lib/daru/vector.rb
zverok: Well, ok, but I think any reasonable refactoring in fact _should_ question refactored code with this question ("what is the reason? is there any reason at all, or it is just historical?")
Otherwise it is pretty easy to fall into "useless refactorings tar-pit", rewriting the unnecessary or unreasonable code several times.

ananyo2012: Well I am not quite sure why this is done. Shall I use the default behaviour and sort them by their values?

zverok: I think let's do write some specs -- and then we'll immediately will see/document, what it outputs and in which order, and will be able to discuss which approach is reasonable.

ananyo2012:
@zverok @v0dro I have made the requested changes and added thorough tests. Have a look.

lib/daru/vector.rb
```diff
@@ -879,41 +879,86 @@ def to_s
...
+    #   =
+    #     n :3
+    #     n valid:0
+    #     median: 2
+    #     mean: 2.0000
+    #     std.dev.: 1.0000
+    #     std.err.: 0.5774
+    #     skew: 0.0000
+    #     kurtosis: -2.3333
+    def summary(indent_level=0) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
ananyo2012: These rubocop checks are related to the Assignment Branch Condition and the Method Length. I think we can safely ignore these checks. There are many places where these are ignored like [here](https://github.com/SciRuby/daru/blob/master/lib/daru/vector.rb#L1066).

zverok: > These rubocop checks are related to the Assignment Branch Condition and the Method Length. I think we can safely ignore these checks.

Which are the most important checks -- much more than spacing or argument naming. No, we can't ignore them in _new_ (or rewritten) code. There was a huge effort of making the code better & conforming to rubocop. Those that left with `disables` -- were left exactly because "this method should be completely rewritten, but for now let's leave it this way".

It is absolutely not acceptable for new/refactored code.

ananyo2012: Ok so to solve this. These are to be broken down into more elementary methods right?

zverok: Yep.

ananyo2012:
@zverok I have solved the rubocop offenses. Please have a look.

lib/daru/vector.rb
```diff
@@ -879,41 +879,114 @@ def to_s
...
+        body = ''
+        max_length = max_col_length(cols)
+        cols[0].each_index do |i|
+          body << "\n| "
+          cols.each_index do |j|
+            body << "#{cols[j][i].to_s.ljust(max_length[j])} | "
           end
         end
+        body
       end
```
zverok: Looks like there is no point having those methods in `Vector` class, they do nothing with concept of vector by themselves. We have `daru/formatters/` forlder for this kind of things... And even `daru/formatters/table.rb` there. I am not sure it suits your needs, but at least all dedicated formatting code should go there, not be mixed up in any class that needs fancy formatting.

spec/dataframe_spec.rb
```diff
@@ -3050,10 +3050,24 @@
...
 
   context "#summary" do
-    it "produces a summary of data frame" do
+    it "shows the dataframe name in summary" do
       expect(@data_frame.summary.match("#{@data_frame.name}")).to_not eq(nil)
       expect(@df_mi.summary.match("#{@df_mi.name}")).to_not eq(nil)
     end
+
+    it "shows the no of rows in dataframe" do
+      expect(@data_frame.summary.match("Number of rows: #{@data_frame.nrows}")).to_not eq(nil)
```
zverok:
```ruby
subject { @data_frame.summary }
it { is_expected.to include("Number of rows: #{@data_frame.nrows}") }
```

spec/dataframe_spec.rb
```diff
@@ -3050,10 +3050,24 @@
...
+    end
+
+    it "shows the individual vector names in the dataframe" do
+      expect(@data_frame.summary.match(/Element:\[#{@data_frame.numeric_vector_names[0]}\]/)).to_not eq(nil)
+    end
+
+    it "shows the summary of individual vectors in dataframe" do
+      expect(@data_frame.summary.match(@data_frame[:a].summary(1))).to_not eq(nil)
+      expect(@df_mi.summary.match(@df_mi[[:a, :one, :bar]].summary(1))).to_not eq(nil)
+    end
```
zverok: To be honest, I feel like those "match this"/"match that" tests are fragile and not very informative (you can't understand from tests what exactly summary outputs). Please take a look at how `#inspect` is tested: https://github.com/SciRuby/daru/blob/master/spec/dataframe_spec.rb#L3648

ananyo2012:
@zverok Ready for one more review.

lib/daru/vector.rb
```diff
@@ -1548,5 +1576,13 @@ def resort_index vector_index, opts
...
         .tap { |res| res.reverse! unless opts[:ascending] }
     end
+
+    def table data
+      Formatters::Table.format(
```
ananyo2012: I used the existing `Daru::Formatters::Table` class to generate the table but had to use it inside a wrapper method with private scope, since using it directly in the `object_summary` was increasing the ABC Metrics.

spec/dataframe_spec.rb
```diff
@@ -3050,9 +3050,34 @@
...
+            |  == b
+            |    n :3
+            |    n valid:0
+            |    factors: 1,2,string
+            |    mode: 1,2,string
+            |    Distribution                                
+            |                 1       1 100.00%
+            |                 2       1 100.00%
+            |            string       1 100.00%
+        }.unindent }
```
zverok: Well, now when we can read it, is it really a _good_ format? What are those `== a`, `== b` worth?

lib/daru/vector.rb
```diff
@@ -1548,5 +1576,13 @@ def resort_index vector_index, opts
...
+        data,
+        headers: ['']*3,
+        row_headers: ['']*frequencies.size
```
ananyo2012: Well I tested it but it is not showing the data without the headers. There is no such no header option. Without headers only an empty string is displayed. Also there are no tests to justify the same. Should I fix this in this PR itself?

spec/dataframe_spec.rb
ananyo2012: Well I followed the reportbuilder format. The `==a` is the name of the individual vectors in the dataframe and separates the display of each vector.

ananyo2012: For vectors the name of the vectors is shown at the top using `=name` format. Same is extended for individual vector elements for a dataframe. Refer to [this line](https://github.com/SciRuby/daru/pull/305/files#diff-8129d70abcb202c11af77e9269873527R900)

ananyo2012:
@zverok I left some comments. Can you have a look?

lib/daru/vector.rb
zverok: Yes, I think it is a good time to update helper class, as you are using it.

spec/dataframe_spec.rb
zverok: Well, but the fact that ReportBuilder uses it doesn't means it is a best readable representation. Though... Let it be, at least for now. Now when we have proper test :+1: we can update the test and rendering at any moment. Please fix `Table` to simplify code a bit and we are good.

lib/daru/vector.rb
ananyo2012: Ok working on this

ananyo2012:
@zverok Done with the table_formatters. Ready for another review!

lib/daru/vector.rb
```diff
@@ -879,41 +879,67 @@ def to_s
...
+    #     mean: 2.0000
+    #     std.dev.: 1.0000
+    #     std.err.: 0.5774
+    #     skew: 0.0000
+    #     kurtosis: -2.3333
+    def summary(indent_level=0)
+      nval = count_values(*Daru::MISSING_VALUES)
+      summary = '  =' * indent_level + "= #{name}" \
+                "\n  n :#{size}" \
+                "\n  n valid:#{nval}"
```
v0dro: `n valid` stands for non-missing data. You're counting missing data.

v0dro: Just call is `non-missing: #{nval}`

ananyo2012: Well I actually followed [this](https://github.com/SciRuby/daru/pull/305/files#diff-8129d70abcb202c11af77e9269873527L891). Changing it.

ananyo2012:
@v0dro done!

lib/daru.rb
```diff
@@ -80,7 +80,7 @@ def plotting_library= lib
   create_has_library :gruff
 end
 
-[['reportbuilder', '~>1.4'], ['spreadsheet', '~>1.1.1']].each do |lib|
+[['spreadsheet', '~>1.1.1']].each do |lib|
```
zverok: With only one library left, there is no need for this complicated machinery anymore ;)

lib/daru/formatters/table.rb
```diff
@@ -26,6 +26,8 @@ def format threshold=nil, spacing=nil
       private
 
       def build_rows threshold # rubocop:disable Metrics/AbcSize
+        @row_headers = [''] * @data.to_a.size if @row_headers.empty?
```
zverok: It should be in `initialize`, I believe.

lib/daru/vector.rb
```diff
@@ -879,41 +879,67 @@ def to_s
...
+    #
+    #   =
+    #     n :3
+    #     non-missing:3
+    #     median: 2
+    #     mean: 2.0000
+    #     std.dev.: 1.0000
+    #     std.err.: 0.5774
+    #     skew: 0.0000
+    #     kurtosis: -2.3333
```
zverok: Make an example output docs this way:
```ruby
# @example
#   dv = Daru::Vector.new [1, 2, 3]
#   puts dv.summary
#
#   # =
#   #   n :3
#   #   non-missing:3
#   #   median: 2
#   #   mean: 2.0000
#   #   std.dev.: 1.0000
#   #   std.err.: 0.5774
#   #   skew: 0.0000
#   #   kurtosis: -2.3333
```
Everything under the `@example` will be highlighted as a Ruby code in docs, and output will be highlighted pretty strangely, so we should make it look as a "comments".

zverok:
(And don't forget to rebase from master, it has a conflict now.)

ananyo2012:
@zverok Made the changes and rebased. Please have a look.

Issue #219, opened 2016-08-20
DataFrame.from_html to read HTML tables


v0dro:
Having a `DataFrame.from_html` method would be great for reading HTML tables into dataframes directly from URLs.

athityakumar:
@v0dro : Scraping gems like `mechanize` / `Nokogiri` can be used to get the data from html, but those html files can have any type of structure of tags where tables and data are located right? How can we get the exact data from html files - do we assume a a particular structure of the html files?

zverok:
As for me:

1. It seems like nokoigiri should be enough
2. Daru should not depend on Nokogiri (as it is large binary gem, not always easy to install). So it should probably be a part of separate "daru importers" effort, as discussed in mailing list.
3. I believe investigation and clarification of requirements is a natural part of this tickets, so the order of proceeding should probably be the following:
  * gather some (different, yet real) examples of table data on the web
  * propose & discuss some method and options to handle decent amount of them
  * then, implement.

athityakumar:
@zverok - I understood the part of using Nokogiri functionality and keeping Daru dependent on Nokogiri. However, the real issue is that - even if Nokogiri is used, we won't be able to get a proper desired output. Web scraping is done through specific searches for proper nodes that are parsed by Nokogiri from classes / divs that are actually present on the html file. And, for each website, this `searching through the nodes to get the desired data` approach willl be different and there is no automated / one single approach to get data in a reliable way, as opposed to extracting data from csv / json files.

zverok:
Well, that's why I am saying about **gather some (different, yet real) examples of table data on the web** above. I assume (maybe wrongly) that this thing _could_ be imagined:

```ruby
Daru.from_html('https://en.wikipedia.org/wiki/Chiang_Mai', '.wikitable.collapsible:first')
```
(it is a [climate table](https://en.wikipedia.org/wiki/Chiang_Mai) on Chiang Mai page) could produce some reasonable output. Just by:
* skipping table headers;
* considering all `<th>`s as a parts of indexes (columns/rows);
* considering all `<td>`s as data.

Here is my prototype code to do so (not the best code in my life, ok; and it ignores td/th thing, just takes first col/cell as an indexes):
```ruby
require 'open-uri'
require 'nokogiri'
require 'daru'

doc = Nokogiri::HTML(open('https://en.wikipedia.org/wiki/Chiang_Mai'))
[...]

PR #311, opened 2017-02-22
Adds support to from_html module

athityakumar:


athityakumar:
@zverok @v0dro : Please review this PR. Are there any code qualities to be refined, or new options to be provided as argument?

spec/io/io_spec.rb
```diff
@@ -86,6 +86,20 @@
       end
     end
 
+    context "#from_html" do
+      let(:path) { path = "https://en.wikipedia.org/wiki/Table_(information)" }
```
zverok: So... Each time somebody runs the specs, the data would be fetched from internet? It could not be done this way for several reasons:
* slow;
* what if my internet is down currently? or wikipedia is down?
* what if the wikipedia page will be changed? the test will fail for no internal reason

We use webmock/vcr for such cases, but for this particular test it is probably wiser to just save some test HTMLs into spec/fixtures, and use them.

athityakumar: Sure, can I save the HTML files into `spec/io/from_html/` directory?

zverok: `spec/fixtures` is right place for this kind of content.

lib/daru.rb
```diff
@@ -95,6 +95,7 @@ def plotting_library= lib
 require 'forwardable'
 require 'erb'
 require 'date'
+require 'mechanize'
```
athityakumar: According to our discussion on issue #219 , wasn't it decided that `Mechanize` should be a development dependency & not be installed along with `daru` on `gem install` right? I'm yet to catch the `LoadError` if `Mechanize` is not installed - I'll do this in the subsequent commit(s).

Specific comment : https://github.com/SciRuby/daru/issues/219#issuecomment-278578308

lib/daru/io/io.rb
```diff
@@ -221,6 +227,25 @@ def from_csv_hash(path, opts)
...
+        }.keep_if { |a| a['index'] != [] }
+      end
+
+      def from_html_get_df_list(list)
+        list.map { |ele|
+          begin
+            indexes, data = ele['index'], ele['data'].transpose
+            hash = Hash[indexes.map { |index| [index,data[indexes.index(index)]] }]
+            Daru::DataFrame.new(hash)
+          rescue # rubocop:disable Lint/HandleExceptions
```
athityakumar: Wikipedia has many tables (most of their links / navigation menus are also tables), and most of them don't have structured data of much importance. Hence, to differentiate the "useful" tables from the "other" tables, I had followed this `table > th` , `table > tr > td` node-paths. In case of the "other" tables, trying to access these node-paths will throw a variety of error(s) depending on the node structure, like `nil NilClass`.

So, I had to `rescue` in order to get the list of `useful` tables as `Daru::Dataframe`s. However, a simple `rescue` block with no code for exception shows a Rubocop offense. Hence, I had to rescue, and disable rubocop for this line.

lib/daru/io/io.rb
```diff
@@ -221,6 +227,25 @@ def from_csv_hash(path, opts)
...
+          {'index' => index, 'data' => data}
+        }.keep_if { |a| a['index'] != [] }
+      end
+
+      def from_html_get_df_list(list)
+        list.map { |ele|
+          begin
+            indexes, data = ele['index'], ele['data'].transpose
+            hash = Hash[indexes.map { |index| [index,data[indexes.index(index)]] }]
+            Daru::DataFrame.new(hash)
```
athityakumar: I'll have a look at the other constructing methods, and change this accordingly.

lib/daru/io/io.rb
```diff
@@ -221,6 +227,25 @@ def from_csv_hash(path, opts)
 
         headers.each_with_index.map { |h, i| [h, csv_as_arrays[i]] }.to_h
       end
+
+      def from_html_get_list(page)
+        page.search('table').map { |table|
+          index = table.search('th').map { |header| header.text.strip }
+          data = table.search('tr').map { |row| row.search('td').map { |val| val.text.strip } }.keep_if { |a| a!=[] }
+          {'index' => index, 'data' => data}
```
athityakumar: Are keys like `:index` and `:data` fine?

lib/daru/io/io.rb
```diff
@@ -221,6 +227,25 @@ def from_csv_hash(path, opts)
 
         headers.each_with_index.map { |h, i| [h, csv_as_arrays[i]] }.to_h
       end
+
+      def from_html_get_list(page)
+        page.search('table').map { |table|
+          index = table.search('th').map { |header| header.text.strip }
+          data = table.search('tr').map { |row| row.search('td').map { |val| val.text.strip } }.keep_if { |a| a!=[] }
```
athityakumar: Yeah, definitely - support for more types of table types and tag structures is required. I'll also have a look at `pandas` `from_html` module and discuss again regarding this.

lib/daru/io/io.rb
```diff
@@ -90,6 +90,12 @@ def from_csv path, opts={}
         Daru::DataFrame.new(hsh,daru_options)
       end
 
+      def from_html path
+        page = Mechanize.new.get(path)
+        list = from_html_get_list(page)
+        from_html_get_df_list(list)
```
athityakumar: I had to split them, due to the `maximum 15-line limit per method` rule of Rubocop. I'll change this accordingly in the next commit(s)

lib/daru/io/io.rb
zverok: Yes. And it allows shorter syntax: `{index: index, data: data}`

lib/daru/io/io.rb
zverok: Yes, splitting methods is a good thing. The problem is how to select right splitting point, and right names for methods. Typically, an approach "write long method, than divide it in 3" is not the most effective.

lib/daru/io/io.rb
zverok: Empty `rescue` of everything is not the most obvious solution. Suggestion: try to describe conditions of "good" table and drop off everything that not accords to this condition. While designing and testing it, you'll obtain much more clear understanding of your task.

Lazy "we'll just rescue everything that is broken" typically will lead you to "swallow" all kind of _useful_ exceptions (like errors in your algorithm).

lib/daru.rb
zverok: Please, when implementing new features, take a time to look around and understand, what approaches are taken on similar features implementations. For this task, you can look at `has_nyaplot?` and similar things in codebase.

lib/daru.rb
```diff
@@ -95,26 +95,20 @@ def plotting_library= lib
 require 'forwardable'
 require 'erb'
 require 'date'
-
```
v0dro: Any specific reason why these new lines were removed? They serve as placeholders since there are so many require statements.

athityakumar: No specific reason - but as they were already grouped properly, I thought that placeholders weren't necessary. I've revived these lines in the last commit. :smile:

athityakumar:
I was adding another test html file from [W3Schools](https://www.w3schools.com/html/html_tables.asp), which is successfully being parsed. However, the texts are truncated while being displayed as a `Daru::Vector`, or a `Daru::DataFrame` - which is not desirable. However, I can see that the `dv.data.to_a` does contain the full scraped text. However, I can't seem to find any truncating, used in the summary / inspect functions anywhere - any help would be appreciated. :smile: 

image

zverok:
> I was adding another test html file from W3Schools, which is successfully being parsed. 

What's the value of it? Daru is data processing library, not school-quality HTML processing. I'd prefer to see some examples of real data, published by some organizations and useful for analysis.

> However, the strings are truncated

Yes, that's the logic of dataframe's `inspect`, allowing reasonable output on any kind of data. It is desirable behavior. [spacing](https://github.com/SciRuby/daru/blob/master/lib/daru/dataframe.rb#L1904) parameter here is an explanation.

v0dro:
@athityakumar can you rebase with the current master?

athityakumar:
@v0dro - Sure, but this PR is still under progress - there are more tests to be added  from [here](https://github.com/pandas-dev/pandas/blob/648ae4f03622d8eafe1ca3b833bd6a99f56bece4/pandas/tests/io/test_html.py), and also provide more options to user like `tuples` , `thousands`, etc.

Ona side note, did you mean to approve the changes on PR #303 ?

v0dro:
No, I meant that the current progress on this PR is upto the mark. Updating the current master will prevent having to spend additional time later for making changes.

athityakumar:
@v0dro @zverok - I've rebased with the master, and added index-detection feature too. It's review time now :smile:

lib/daru.rb
```diff
@@ -80,12 +80,14 @@ def plotting_library= lib
   create_has_library :gruff
 end
 
-begin
-  gem 'spreadsheet', '~>1.1.1'
-  require 'spreadsheet'
-rescue LoadError
-  STDERR.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
-  ' spreadsheet functions.'
+[['spreadsheet', '~>1.1.1'], ['mechanize','~>2.7.5']].each do |lib|
```
zverok: `do |name, verson|` if we are returning to this idea.
And, probably:
```ruby
{
  'spreadsheet' => '~>1.1.1'
  'mechanize' => '~>2.7.5'
}.each do |name, version|
```

lib/daru/dataframe.rb
```diff
@@ -117,6 +117,21 @@ def from_plaintext path, fields
...
+      # the data should be present in a html file with proper combination of <table>, <th> and <td> tags.
+      #
+      # == Arguments
+      #
+      # * path - Path / URL of the html file to be read.
+      # * fields - A Hash, with match key to search. data, order, index keys can overwrite the scraped values from web.
+      #
+      # == Usage
+      #
+      # df = Daru::DataFrame.from_html(path, index: [:p1, :p2, :p3, :p4, :p5, :p6, :p7])[0]
```
zverok: Modern YARD docs, please. As per http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md

zverok: And why it ends up with `[0]`? (Well, I understand there could be several tables returned, but it is absolutely not obvious from docs or example).

lib/daru/io/io.rb
```diff
@@ -185,6 +185,18 @@ def load filename
         end
       end
 
+      def from_html path, opts
+        page = Mechanize.new.get(path)
+        page.search('table').map { |table| parse_html_table table }
+            .keep_if { |table| search_for_match table, opts[:match] }
+            .reject(&:nil?).reject(&:empty?)
```
zverok: `.reject(&:nil?)` => `.compact`

lib/daru/io/io.rb
```diff
@@ -220,6 +232,41 @@ def from_csv_hash(path, opts)
 
         headers.each_with_index.map { |h, i| [h, csv_as_arrays[i]] }.to_h
       end
+
+      # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
```
zverok: Unacceptable, sorry :(

zverok: Unacceptable, sorry :( You should try to write better code, not disable everything you dislike.

lib/daru/io/io.rb
```diff
@@ -185,6 +185,18 @@ def load filename
         end
       end
 
+      def from_html path, opts
+        page = Mechanize.new.get(path)
+        page.search('table').map { |table| parse_html_table table }
+            .keep_if { |table| search_for_match table, opts[:match] }
```
zverok: How this (`:match`) is supposed to be used? It is not documented anywhere.

lib/daru/io/io.rb
```diff
@@ -220,6 +232,41 @@ def from_csv_hash(path, opts)
...
+      end
+
+      def choose_value(scraped_val={}, user_val=nil)
+        unless user_val.nil?
+          user_val.each do |key,val|
+            scraped_val[key] = val
+          end
+        end
+        scraped_val
+      end
```
zverok: What it does and why? Neither of method or parameter names are informative, and it is hard to understand the reasons.

spec/io/io_spec.rb
```diff
@@ -291,6 +291,79 @@
...
     end
+
+    context "#from_html" do
+      let(:macau) { "file://#{Dir.pwd}/spec/fixtures/macau.html" }
+      let(:valid_markup) { "file://#{Dir.pwd}/spec/fixtures/valid_markup.html" }
+      let(:wiki_climate) { "file://#{Dir.pwd}/spec/fixtures/wiki_climate.html" }
+      let(:wiki_table_info) { "file://#{Dir.pwd}/spec/fixtures/wiki_table_info.html" }
+
+      it "returns default dataframe list from html tables" do
+        df = Daru::DataFrame.from_html(wiki_table_info)[0]
```
zverok: This should be `subject`, and different df examples should be different `context`s

lib/daru.rb
athityakumar: Sure, will make this change.

athityakumar:
@zverok @v0dro - I've updated the code, documentation and tests for the `Daru::DataFrame#from_html`module. Please review the changes. 😄

spec/io/io_spec.rb
```diff
@@ -269,6 +269,161 @@
...
+          expect(dfs.class).to eq Array
+          expect(dfs.first.class).to eq Daru::DataFrame
+          expect(dfs.first).to eq(Daru::DataFrame.new(
+              [["Tinu", "Blaszczyk", "Lily", "Olatunkboh", "Adrienne", "Axelia", "Jon-Kabat"],
+              ["Elejogun", "Kostrzewski", "McGarrett", "Chijiaku", "Anthoula", "Athanasios", "Zinn"],
+              ["14", "25", "16", "22", "22", "22", "22"]], 
+              order: ["First name","Last name","Age"]
+            )
+          )
+        end
```
zverok:
```ruby
context 'default output' do
  subject { Daru::DataFrame.from_html(path) }
  it { is_expected.to be_an(Array).and all be_a(Daru::DataFram) }
  its(:first) { is_expected.to eq ....
end

context 'with options provided by user' do 
....
```

athityakumar:
@zverok - I've made the spec changes. One more thing before merging. I see that the fixture HTML files added in this PR have increased the repository's size from ~3.4MB to ~4MB. Should I truncate these HTML files, keeping just relevant content like table tags?

lib/daru/io/io.rb
```diff
@@ -183,6 +183,18 @@ def load filename
...
+      def from_html path, opts
+        page = Mechanize.new.get(path)
+        page.search('table').map { |table| html_parse_table table }
+            .keep_if { |table| html_search table, opts[:match] }
+            .compact
+            .map { |table| html_decide_values table, opts }
+            .map { |table| html_table_to_dataframe table }
+      rescue LoadError
+        STDERR.puts '\nInstall the mechanize gem version 2.7.5 for using'\
+        ' from_html function.'
```
zverok: I believe it should not be `puts`, but rather failing with meaningful error. Because script relying on `from_html` should be terminated if `mechanize` is not accessible, not continue with undefined results and small warning message.

athityakumar: Acknowledged, does a run-time error like `raise RuntimeError, 'Install the mechanize gem 2.7.5 for using from_html function'` work?

zverok: Yep

zverok: Yep

athityakumar:
Thanks. I'll soon port all the IO modules( including `from_html`) to daru-io. If a `0.1.0` version of daru-io is released, then we can probably remove the existing IO modules from daru. Please let me know if this can be done after porting is done.

PR #312, opened 2017-03-01
where clause when data contains missing/undefined values

Shekharrajak:
Fixes https://github.com/SciRuby/daru/issues/246

```
irb(main):001:0> v= Daru::Vector.new([1,2,3,Float::NAN, nil])
=> #<Daru::Vector(5)>
   0   1
   1   2
   2   3
   3 NaN
   4 nil
irb(main):002:0> v.where(v.lt(4))
=> #<Daru::Vector(3)>
   0   1
   1   2
   2   3
irb(main):003:0> v.where(v.lt(3))
=> #<Daru::Vector(2)>
   0   1
   1   2
irb(main):004:0> v.where(v.lt(2))
=> #<Daru::Vector(1)>
   0   1
irb(main):005:0> v.where(v.lt(1))
=> #<Daru::Vector(0)>
```

### DIFF
--- a/lib/daru/core/query.rb
+++ b/lib/daru/core/query.rb
@@ -39,7 +39,7 @@ module Daru
 
       class << self
         def apply_scalar_operator operator, data, other
-          BoolArray.new data.map { |d| d.respond_to?(operator) && !!d.send(operator, other) }
+          BoolArray.new data.map { |d| !!d.send(operator, other) if d.respond_to?(operator) }
         end
 
         def apply_vector_operator operator, vector, other
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -352,7 +352,10 @@ module Daru
         if other.is_a?(Daru::Vector)
           mod.apply_vector_operator operator, self, other
         else
-          mod.apply_scalar_operator operator, @data,other
+          data_dup = @data.dup
+          missing_undefined_indexes = nil_positions + nan_positions
+          missing_undefined_indexes.each { |i| data_dup.delete_at(i) }
+          mod.apply_scalar_operator operator, data_dup, other
         end
       end
       alias_method operator, method if operator != :== && operator != :!=
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -352,10 +352,14 @@ module Daru
         if other.is_a?(Daru::Vector)
           mod.apply_vector_operator operator, self, other
         else
-          data_dup = @data.dup
           missing_undefined_indexes = nil_positions + nan_positions
-          missing_undefined_indexes.each { |i| data_dup.delete_at(i) }
-          mod.apply_scalar_operator operator, data_dup, other
+          if operator != :== && operator != :!= && missing_undefined_indexes.size
+            data_dup = @data.dup
+            missing_undefined_indexes.each { |i| data_dup.delete_at(i) }
+            mod.apply_scalar_operator operator, data_dup, other
+          else
+            mod.apply_scalar_operator operator, data, other
+          end
         end
       end
       alias_method operator, method if operator != :== && operator != :!=
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -352,14 +352,7 @@ module Daru
         if other.is_a?(Daru::Vector)
           mod.apply_vector_operator operator, self, other
         else
-          missing_undefined_indexes = nil_positions + nan_positions
-          if operator != :== && operator != :!= && missing_undefined_indexes.size
-            data_dup = @data.dup
-            missing_undefined_indexes.each { |i| data_dup.delete_at(i) }
-            mod.apply_scalar_operator operator, data_dup, other
-          else
-            mod.apply_scalar_operator operator, data, other
-          end
+          mod.apply_scalar_operator operator, @data, other
         end
       end
       alias_method operator, method if operator != :== && operator != :!=
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -1390,17 +1390,17 @@ describe Daru::Vector do
 
   context '#is_values' do
     let(:dv) { Daru::Vector.new [10, 11, 10, nil, nil] }
-    
+
     context 'single value' do
       subject { dv.is_values 10 }
       it { is_expected.to be_a Daru::Vector }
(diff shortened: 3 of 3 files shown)


Shekharrajak:
when I do ` bundle exec rspec` in my system I don't get any error. I didn't understand [this travis ](https://travis-ci.org/SciRuby/daru/jobs/206556696#L1126) failure. 

Also I don't find these errors   : 

```
/spec/vector_spec.rb:1397 # Daru::Vector#is_values single value to_a should eq [true, false, true, false, false]
rspec ./spec/vector_spec.rb:1403 # Daru::Vector#is_values multiple values to_a should eq [true, false, true, true, true]
```

in `spec/vector_spec.rb:1397` and `1403` line I don't find those lines.

zverok:
@Shekharrajak looks like some Travis flakiness (maybe due to those S3 problems, that ruined half-the-internet yesterday?) Now it shows "There was an error while loading data." :man_facepalming: 

Try rerun build? (Commit something dummy, or close and reopen PR)

Shekharrajak:
Thanks @zverok , I have closed and reopen the PR.

zverok:
Travis is ill now :( https://www.traviscistatus.com
Let's wait. I'll review the PR in the meantime

Shekharrajak:
@zverok , I have updated the PR branch(rebase with upstream/master) and fixed the error.

lib/daru/vector.rb
```diff
@@ -352,7 +352,14 @@ def == other
...
         else
-          mod.apply_scalar_operator operator, @data,other
+          missing_undefined_indexes = nil_positions + nan_positions
+          if operator != :== && operator != :!= && missing_undefined_indexes.size
+            data_dup = @data.dup
+            missing_undefined_indexes.each { |i| data_dup.delete_at(i) }
+            mod.apply_scalar_operator operator, data_dup, other
+          else
+            mod.apply_scalar_operator operator, data, other
+          end
```
zverok: I don't feel it as a readable code. 
First, what is the point? "When we are _applying scalar operator_,  we should ignore 'missing' values", right? (And only for _some_ of operators.) Why don't you think it is a responsibility of `apply_scalar_operator` itself to tell applicable elements from others, and the operators that are ok with "missing values" from those that are not? (And consider, BTW, the fact that `Float::NAN == Float::NAN # => false`)

spec/vector_spec.rb
```diff
@@ -1995,4 +1995,15 @@ def expect_correct_tokens hash
     end
   end
 
+  context '#where clause when Nan, nil data value is present' do
+    before do
+        @v = Daru::Vector.new([1,2,3,Float::NAN, nil])
```
Shekharrajak: Thanks for review , I will do required changes.

lib/daru/vector.rb
Shekharrajak: Thanks for review and telling me `Float::NAN == Float::NAN # => false`.This must be done in `apply_scalar_operator `. I will try it.

Shekharrajak: @zverok , according to this link : http://ruby-doc.org/core-2.0.0/NilClass.html 

the problem is in `nil` (it can handle the `NaN`) . `nil` is `0` so should we compare it as `0` ?

Shekharrajak:
```
irb(main):001:0> v= Daru::Vector.new([1,2,3,Float::NAN, nil])
=> #<Daru::Vector(5)>
   0   1
   1   2
   2   3
   3 NaN
   4 nil
irb(main):002:0> v.where(v.lt(4))
=> #<Daru::Vector(4)>
   0   1
   1   2
   2   3
   4 nil
```

`nil ` also is included in ans if we consider `nil` as `0` value.

Shekharrajak: I think we should compare it as `0` since 
```
irb(main):022:0> nil.to_f
=> 0.0

```

zverok: I believe that any "empty" (nil/NaN) value should not be included  in `lt`/`gt` comparison. I am just wondering, how `eq` comparison should work. Though, as `[1,2,3,Float::NAN, nil].index(Float::NAN)` in plain Ruby doesn't find `NaN`, I believe it is ok that `eq` will not find it either. 
So, just ignore "NaN problem", but code should be made more readable, as suggested.

Shekharrajak: All the operators (`< , > , ==, !=`) works for `Float::NAN` . Actually we need to handle `nil` only. I will add one condition in `apply_scalar_operator` .

spec/vector_spec.rb
```diff
@@ -1390,17 +1390,17 @@
 
   context '#is_values' do
     let(:dv) { Daru::Vector.new [10, 11, 10, nil, nil] }
-    
+
```
Shekharrajak: Just removed spaces.

lib/daru/vector.rb
```diff
@@ -352,7 +352,7 @@ def == other
         if other.is_a?(Daru::Vector)
           mod.apply_vector_operator operator, self, other
         else
-          mod.apply_scalar_operator operator, @data,other
+          mod.apply_scalar_operator operator, @data, other
```
Shekharrajak: just added space after the comma.

lib/daru/core/query.rb
```diff
@@ -39,7 +39,7 @@ def inspect
 
       class << self
         def apply_scalar_operator operator, data, other
-          BoolArray.new data.map { |d| d.respond_to?(operator) && !!d.send(operator, other) }
+          BoolArray.new data.map { |d| !!d.send(operator, other) if d.respond_to?(operator) }
```
Shekharrajak: first checking whether operator works for `d` or not, if yes then only do `!!d.send(operator, other)`. (Actually for `nil` `d.respond_to?(operator)` is false when operator is any of these : [ '>' '<' , '=<' ,'>='], for `NaN` all these already works correctly)

Shekharrajak: Ping @zverok , I think this logic is good.

Shekharrajak:
@zverok ,I think this PR is good to go.

lib/daru/core/query.rb
zverok: TBH, I preferred the first version. Though, let's move along.

PR #313, opened 2017-03-01
removed unnecessary spaces

Shekharrajak:
My text editor(vim/sublime) removes the extra spaces automatically whenever I save the files. So I get some extra changes in file(after saving my changes). 

I hope it is better to remove all extra spaces in one PR.


zverok:
Thanks!

Issue #159, opened 2016-06-05
Sort by index


v0dro:
A `sort_by_index` method in `DataFrame` and `Vector` that will sort based on the index and not values of data.

kojix2:
`sort_by_index` method is still not defined in `DataFrame`. 

```ruby
df.reindex df.index.sort
```

PR #315, opened 2017-03-04
defining sort for Index and sort_by_index for vector

Shekharrajak:
Fixes https://github.com/SciRuby/daru/issues/159

```
irb(main):007:0> dv = Daru::Vector.new [11, Float::NAN, nil], index: [23, 22, 21]
=> #<Daru::Vector(3)>
  23  11
  22 NaN
  21 nil
irb(main):008:0> dv.sort_by_index()
=> #<Daru::Vector(3)>
  21 nil
  22 NaN
  23  11
irb(main):009:0> dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
=> #<Daru::Vector(3)>
  23  11
  22  12
  21  13
irb(main):010:0> dv.sort_by_index()
=> #<Daru::Vector(3)>
  21  13
  22  12
  23  11
irb(main):011:0> dv.sort_by_index(ascending: false)
=> #<Daru::Vector(3)>
  23  11
  22  12
  21  13
```

### DIFF
--- a/lib/daru/index/index.rb
+++ b/lib/daru/index/index.rb
@@ -1,5 +1,5 @@
 module Daru
-  class Index
+  class Index # rubocop:disable Metrics/ClassLength
     include Enumerable
     # It so happens that over riding the .new method in a super class also
     # tampers with the default .new method for class that inherit from the
@@ -220,6 +220,31 @@ module Daru
     def reorder(new_order)
       from = to_a
       self.class.new(new_order.map { |i| from[i] })
+    end
+
+    # Sorts a `Index`, according to its values. Defaults to ascending order sorting.
+    #
+    # == Options:
+    #
+    # ascending::
+    #   This can be `true` or `false`. False, to get descending order.
+    #
+    # == Returns:
+    # sorted `Index` according to its values.
+    #
+    # == Usage
+    #
+    #   di = Daru::Index.new [100, 99, 101, 1, 2]
+    #   # Say you want to sort descending order
+    #   di.sort(ascending: false)
+    def sort opts={}
+      opts = {ascending: true}.merge(opts)
+      if opts[:ascending]
+        new_index, = @relation_hash.sort.transpose
+      else
+        new_index, = @relation_hash.sort.reverse.transpose
+      end
+      self.class.new(new_index)
     end
 
     private
--- a/lib/daru/index/index.rb
+++ b/lib/daru/index/index.rb
@@ -222,21 +222,20 @@ module Daru
       self.class.new(new_order.map { |i| from[i] })
     end
 
-    # Sorts a `Index`, according to its values. Defaults to ascending order sorting.
+    # Sorts a `Index`, according to its values. Defaults to ascending order
+    # sorting.
     #
-    # == Options:
+    # @param [Hash] opts the options for sort method.
+    # @option opts [Boolean] :ascending False, to get descending order.
     #
-    # ascending::
-    #   This can be `true` or `false`. False, to get descending order.
+    # @return [Index] sorted `Index` according to its values.
     #
-    # == Returns:
-    # sorted `Index` according to its values.
-    #
-    # == Usage
-    #
+    # @example
     #   di = Daru::Index.new [100, 99, 101, 1, 2]
-    #   # Say you want to sort descending order
-    #   di.sort(ascending: false)
+    #   # Say you want to sort in descending order
+    #   di.sort(ascending: false) #=> Daru::Index.new [101, 100, 99, 2, 1]
+    #   # Say you want to sort in ascending order
+    #   di.sort #=> Daru::Index.new [1, 2, 99, 100, 101]
     def sort opts={}
       opts = {ascending: true}.merge(opts)
       if opts[:ascending]
@@ -244,6 +243,7 @@ module Daru
       else
         new_index, = @relation_hash.sort.reverse.transpose
       end
+
       self.class.new(new_index)
     end
 
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -623,8 +623,7 @@ module Daru
     #   #=> Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
     def sort_by_index opts={}
       opts = {ascending: true}.merge(opts)
-      new_arrangement = resort_index(@index.each_with_index, opts)
-      _, new_order = new_arrangement.transpose
+      _, new_order = resort_index(@index.each_with_index, opts).transpose
 
       reorder new_order
     end
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -603,6 +603,31 @@ module Daru
       Daru::Vector.new(vector, index: index, name: @name, dtype: @dtype)
     end
 
+    # Sorts the vector according to it's`Index` values. Defaults to ascending
+    # order sorting.
+    #
+    # == Options
+    #
+    # ascending::
+    #   This can be `true` or `false`. False, if false, will sort `index` in
+    #   descending order. Defaults to true.
+    #
+    # == Returns:
+    #   new sorted `Vector` according to the index values.
+    #
+    # == Usage
+    #
+    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   # Say you want to sort index in descending order
+    #   di.sort_by_index(ascending: false)
+    def sort_by_index opts={}
+      opts = {ascending: true}.merge(opts)
+      new_arrangement = resort_index(@index.each_with_index, opts)
+      _, new_order = new_arrangement.transpose
+
+      reorder new_order
+    end
+
     DEFAULT_SORTER = lambda { |(lv, li), (rv, ri)|
       case
       when lv.nil? && rv.nil?
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -606,20 +606,21 @@ module Daru
     # Sorts the vector according to it's`Index` values. Defaults to ascending
     # order sorting.
     #
-    # == Options
+    # @param [Hash] opts the options for sort_by_index method.
+    # @option opts [Boolean] :ascending false, will sort `index` in
+    #  descending order.
     #
-    # ascending::
-    #   This can be `true` or `false`. False, if false, will sort `index` in
-    #   descending order. Defaults to true.
+    # @return [Vector] new sorted `Vector` according to the index values.
     #
-    # == Returns:
-    #   new sorted `Vector` according to the index values.
+    # @example
     #
-    # == Usage
-    #
-    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   dv = Daru::Vector.new [11, 13, 12], index: [23, 21, 22]
+    #   # Say you want to sort index in ascending order
+    #   dv.sort_by_index(ascending: true)
+    #   #=> Daru::Vector.new [13, 12, 11], index: [21, 22, 23]
     #   # Say you want to sort index in descending order
-    #   di.sort_by_index(ascending: false)
+    #   dv.sort_by_index(ascending: false)
+    #   #=> Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
     def sort_by_index opts={}
       opts = {ascending: true}.merge(opts)
       new_arrangement = resort_index(@index.each_with_index, opts)
--- a/spec/index/index_spec.rb
+++ b/spec/index/index_spec.rb
@@ -71,16 +71,21 @@ describe Daru::Index do
   end
 
   context '#sort' do
-    let(:string_index) { Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
-    let(:num_index) { Daru::Index.new [100, 99, 101, 1, 2] }
-    it 'when index values are string' do
-      expect(string_index.sort()).to eq Daru::Index.new ['amp', 'guitar', 'mic', 'speaker']
-      expect(string_index.sort(ascending: false)).to eq string_index
+    let(:asc) { index.sort }
+    let(:desc) { index.sort(ascending: false) }
+
+    context 'string index' do
+      let(:index) { Daru::Index.new ['mic', 'amp', 'guitar', 'speaker'] }
+      specify { expect(asc).to eq Daru::Index.new ['amp', 'guitar', 'mic',
+       'speaker'] }
+      specify { expect(desc).to eq Daru::Index.new ['speaker', 'mic', 'guitar',
+       'amp'] }
     end
 
-    it 'when index values are numbers' do
-      expect(num_index.sort()).to eq Daru::Index.new [1, 2, 99, 100, 101]
-      expect(num_index.sort(ascending: false)).to eq Daru::Index.new [101, 100, 99, 2, 1]
+    context 'number index' do
+      let(:index) { Daru::Index.new [100, 99, 101, 1, 2]  }
+      specify { expect(asc).to eq Daru::Index.new [1, 2, 99, 100, 101] }
+      specify { expect(desc).to eq Daru::Index.new [101, 100, 99, 2, 1] }
     end
   end
 
--- a/spec/index/index_spec.rb
+++ b/spec/index/index_spec.rb
@@ -70,6 +70,20 @@ describe Daru::Index do
     end
   end
 
+  context '#sort' do
+    let(:string_index) { Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
+    let(:num_index) { Daru::Index.new [100, 99, 101, 1, 2] }
+    it 'when index values are string' do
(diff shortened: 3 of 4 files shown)


Shekharrajak:
I will add testcases in next commit.

Shekharrajak:
I didn't understand why [this case](https://travis-ci.org/SciRuby/daru/jobs/207636464#L1099) is failing . Let me know if someone find the mistake in testcase.

spec/index/index_spec.rb
```diff
@@ -70,6 +70,40 @@
...
+      expect(idx1.sort(ascending: false)).to eq(idx1_desc)
+      expect(idx3.sort(ascending: false)).to eq(idx3_desc)
+    end
+
+    it 'To get the new order, set get_new_order to true' do
+      expect(idx3.sort(get_new_order: true)).to eq [idx3_asc, idx3_asc_new_order]
+      expect(idx3.sort(ascending: false, get_new_order: true)).to eq [idx3_desc, idx3_desc_new_order]
+    end
+
+    it "raises ArgumentError if all the values are not of the same class" do
```
Shekharrajak: this case is failing https://travis-ci.org/SciRuby/daru/jobs/207636464#L1099

I don't know where is the mistake.

lib/daru/index/index.rb
```diff
@@ -222,6 +222,44 @@ def reorder(new_order)
...
+    #
+    # == Usage
+    #
+    #   di = Daru::Index.new [100, 99, 101, 1, 2]
+    #   # Say you want to sort descending order
+    #   di.sort(ascending: false)
+    #   # Want to get the new order as well
+    #   di.sort(get_new_order: true)
+    def sort opts={}
+      raise(ArgumentError, 'All elements are not of the same type/class.') unless check_index_data_class(to_a)
```
v0dro: Why does this line exist? Who said an index has to be the same class for sorting to happen? A symbol and string can be compared. There no need for such a check. Leave it to the user.

Also, why are you passing `to_a` into this method and then calling index_data[0].class`inside the other method? Didn't you think of the overhead that this will have?

lib/daru/index/index.rb
```diff
@@ -222,6 +222,44 @@ def reorder(new_order)
       self.class.new(new_order.map { |i| from[i] })
     end
 
+    # Sorts a `Index` according to its values. Defaults to ascending order sorting.
+    # raises ArgumentError when all the values are not of the same class.
+    #
+    # == Options
+    #
+    # * +:ascending+ - if false, will sort in descending order. Defaults to true.
```
v0dro: Use YARD syntax for docs.

lib/daru/index/index.rb
```diff
@@ -222,6 +222,44 @@ def reorder(new_order)
...
+    #
+    # * +:get_new_order+ - if true, then method will return the new order (in list) of
```
v0dro: Why does this option exist? What use case does it serve?

lib/daru/vector.rb
```diff
@@ -603,6 +603,29 @@ def sort opts={}, &block
...
+    #
+    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   # Say you want to sort index in descending order
+    #   di.sort_by_index(ascending: false)
+    def sort_by_index opts={}
+      opts = {ascending: true, get_new_order: true}.merge(opts)
+      new_index, new_order = @index.sort(opts)
+      new_vector = new_order.map { |i| @data[i] }
+
+      Daru::Vector.new(new_vector, index: new_index, name: @name, dtype: @dtype)
```
v0dro: If you're not even returning the `new_order` why even give an option to the user for it?

spec/index/index_spec.rb
```diff
@@ -70,6 +70,40 @@
     end
   end
 
+  context '#sort' do
+    let(:idx1) { Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
+    let(:idx2) { Daru::Index.new [:one, 'one', 1, 2, :two] }
```
v0dro: This is over testing. Why do you need so many indexes? Also, ìdx1...4`is a bad naming convention. It does not tell anything about your data.

lib/daru/vector.rb
```diff
@@ -603,6 +603,29 @@ def sort opts={}, &block
...
+    #
+    # == Usage
+    #
+    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   # Say you want to sort index in descending order
+    #   di.sort_by_index(ascending: false)
+    def sort_by_index opts={}
+      opts = {ascending: true, get_new_order: true}.merge(opts)
+      new_index, new_order = @index.sort(opts)
+      new_vector = new_order.map { |i| @data[i] }
```
v0dro: Using map is a slow way of doing it. Simply making tuples of the index and data and sorting by the index element is faster. Use that.

lib/daru/index/index.rb
Shekharrajak: I will edit the PR. Actually I was following these [syntax](https://github.com/SciRuby/daru/blob/master/lib/daru/vector.rb#L161)

lib/daru/index/index.rb
Shekharrajak: I am using it [here](https://github.com/SciRuby/daru/pull/315/files/f6ce54c0c5a91b89526427ddfe8450f50c28949c#diff-a1adf8860e167030bf548e2529c7d508R249) . It is for getting the new arrangement order of index value. 

e.g if index is `[100, 99, 101, 1, 2]` then sorted will be `[1, 2, 99, 100, 101]` so the new arrangement of indexes is `[3, 4, 1, 0, 2]` I am calling it as `new_order`.

lib/daru/index/index.rb
Shekharrajak: @v0dro , thanks . I was little confused about it. Actually there is the error when string and integers are in `Index` argument. I will remove it.

lib/daru/vector.rb
Shekharrajak: I think we can do this as well : 
```
new_order.each do |i|
  new_vector.push(@data[i]) 
end
```

when `new_order`  = `[2,1,0]`, `@data` = `[11, 12, 13]`
then after this loop `new_vector` = `[13,12,11]`

Is it better ?

lib/daru/vector.rb
Shekharrajak: After setting  `get_new_order = true`  this line => `new_index, new_order = @index.sort(opts)` will return `new_index, new_order ` if it is set to `false` (by default) then only `new_index` ill be returned. 

here `new_order` (new arrangement of the index value) helps to rearrange the `vector` data.

lib/daru/index/index.rb
v0dro: Then why keep it in a public function if its only use case is private?

lib/daru/vector.rb
v0dro: Calling `[]`on `@data` directly will not allow you to search the indexes that the user has supplied. You will be using numeric indexes if you do such a thing, which is not what you should be doing.

lib/daru/vector.rb
v0dro: My question is: if the actions that take place by passing `:get_new_order` as `true` are purely internal to the library, what is the need to expose this functionality to the user?

v0dro: My question is: if the actions that take place by passing `:get_new_order` as `true` are purely internal to the library, what is the need to expose this functionality to the user?

lib/daru/index/index.rb
v0dro: Leave it to the user to get the index right.

spec/index/index_spec.rb
zverok: `expect { idx.sort }.to raise_error` (curly braces). The idea is you pass some code to `expect`, and what you expect it to do (raise error). And then it is checked (code is run, and error is caught). With syntax `expect(idx2.sort)` what is going on is:
1. sort is called
2. expect is called with result of `sort` call (or, rather it is not, because error is already raised).

BTW, are you aware you can test it locally, just calling `rspec` command?

lib/daru/index/index.rb
zverok: You should drop it. No other Ruby collection provides this fucntionality, and it is hard to justify. YAGNI is a pretty useful principle.

lib/daru/index/index.rb
```diff
@@ -270,6 +308,16 @@ def validate_positions *positions
...
     end
 
+    # check whether all the index data are of same type/class
+    def check_index_data_class index_data
+      check = index_data.all? { |d| d.is_a? index_data[0].class }
+      if check
+        true
+      else
+        false
+      end
```
zverok: Serioulsy???

spec/index/index_spec.rb
zverok: Yes, if you want to show how different indexes behave, you should do it with `context`.
Like this:

```ruby
context '#sort' do
  subject { index.sort(params) }
  context 'when String index' do
    let(:index) { Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
    it { is_expected.to eq Daru::Index.new ['amp', 'guitar', 'mic', 'speaker'] }

    context 'descending' do
      let(:params) { {ascending: false} }
      it { is_expected.to eq Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
    end
  end
...
```

spec/index/index_spec.rb
Shekharrajak: Thanks @zverok . Yes I had run the testcase locally using `bundle exec rspec`.

lib/daru/index/index.rb
Shekharrajak: I have removed this option.

lib/daru/index/index.rb
Shekharrajak: It is done now.

lib/daru/vector.rb
Shekharrajak: In [2nd attempt](https://github.com/SciRuby/daru/pull/315/commits/a717aa0698dba7c5286bcc73ab2d31fbf4437753) I have used another way of sorting.

lib/daru/vector.rb
Shekharrajak: Yeah it is mostly for internal use. I didn't find other good way that time. But now I am not using this. [2nd attempt](https://github.com/SciRuby/daru/pull/315/commits/a717aa0698dba7c5286bcc73ab2d31fbf4437753)

spec/index/index_spec.rb
Shekharrajak: Thanks @zverok @v0dro , I have changed these line in[ new commit](https://github.com/SciRuby/daru/pull/315/commits/a717aa0698dba7c5286bcc73ab2d31fbf4437753)

lib/daru/index/index.rb
Shekharrajak: It was really bad code. I have removed these.

Shekharrajak:
@zverok @v0dro , I think now this PR is good to go. Please review.

lib/daru/vector.rb
```diff
@@ -603,6 +603,31 @@ def sort opts={}, &block
...
+    #   new sorted `Vector` according to the index values.
+    #
+    # == Usage
+    #
+    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   # Say you want to sort index in descending order
+    #   di.sort_by_index(ascending: false)
+    def sort_by_index opts={}
+      opts = {ascending: true}.merge(opts)
+      new_arrangement = resort_index(@index.each_with_index, opts)
```
Shekharrajak: I am using resort_index similar to [sort](https://github.com/SciRuby/daru/pull/315/files#diff-8129d70abcb202c11af77e9269873527R598)

lib/daru/index/index.rb
```diff
@@ -222,6 +222,31 @@ def reorder(new_order)
...
+    #   di = Daru::Index.new [100, 99, 101, 1, 2]
+    #   # Say you want to sort descending order
+    #   di.sort(ascending: false)
+    def sort opts={}
+      opts = {ascending: true}.merge(opts)
+      if opts[:ascending]
+        new_index, = @relation_hash.sort.transpose
+      else
+        new_index, = @relation_hash.sort.reverse.transpose
+      end
```
Shekharrajak: @zverok , I get this error in rubocop

```
 Style/MultilineTernaryOperator: Avoid multi-line ternary operators, use if or unless instead.
```

so I am not changing it.

lib/daru/vector.rb
```diff
@@ -603,6 +603,31 @@ def sort opts={}, &block
...
+    #   descending order. Defaults to true.
+    #
+    # == Returns:
+    #   new sorted `Vector` according to the index values.
+    #
+    # == Usage
+    #
+    #   dv = Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    #   # Say you want to sort index in descending order
+    #   di.sort_by_index(ascending: false)
```
Shekharrajak: Thanks , I have edited the docs.

spec/index/index_spec.rb
```diff
@@ -70,6 +70,20 @@
     end
   end
 
+  context '#sort' do
+    let(:string_index) { Daru::Index.new ['speaker', 'mic', 'guitar', 'amp'] }
+    let(:num_index) { Daru::Index.new [100, 99, 101, 1, 2] }
+    it 'when index values are string' do
+      expect(string_index.sort()).to eq Daru::Index.new ['amp', 'guitar', 'mic', 'speaker']
+      expect(string_index.sort(ascending: false)).to eq string_index
```
Shekharrajak: Thanks, I have edited the testcase.

lib/daru/index/index.rb
zverok: Yep, multi-line ternary operator is unacceptable. 
But!
You can have it in one line (if line length is not too large), or, alternatively, have 
```ruby
new_index = if ..... 
```
To show that entire `if` is about calculating this variable.

spec/index/index_spec.rb
```diff
@@ -70,6 +70,20 @@
...
+    end
+
+    it 'when index values are numbers' do
+      expect(num_index.sort()).to eq Daru::Index.new [1, 2, 99, 100, 101]
+      expect(num_index.sort(ascending: false)).to eq Daru::Index.new [101, 100, 99, 2, 1]
+    end
+  end
```
Shekharrajak: Thanks @zverok, these lines are really better. I have edited the PR accordingly.

lib/daru/index/index.rb
Shekharrajak: Yeah, it is better for small length line. I will keep this in mind.

spec/vector_spec.rb
```diff
@@ -1966,6 +1966,24 @@ def expect_correct_tokens hash
     end
   end
 
+  context "#sort_by_index" do
+    let(:asc) { vector.sort_by_index }
+    let(:desc) { vector.sort_by_index(ascending: false) }
+
+    context 'numeric vector' do
+      let(:vector) { Daru::Vector.new [11, 13, 12], index: [23, 21, 22] }
+      specify { expect(asc.to_a).to eq [13, 12, 11] }
```
v0dro: Why use this syntax and not the `its {expected.... }` type of syntax that is present in the other examples?

zverok: Well, I believe it was my suggestion. Personally I prefer to use `it` and `specify` semantically (e.g. what reads better). For this case, "it expect asc to eq" reads strange, while "specify: expect asc to eq" is better. And using `its` is inconvenient, because we need to test two different subjects (asc and desc) in one context, it is not reasonable to split them into even more contexts.

v0dro: Ah. Alright then. Lets stick with this.

lib/daru/vector.rb
```diff
@@ -603,6 +603,32 @@ def sort opts={}, &block
...
+    #   # Say you want to sort index in ascending order
+    #   dv.sort_by_index(ascending: true)
+    #   #=> Daru::Vector.new [13, 12, 11], index: [21, 22, 23]
+    #   # Say you want to sort index in descending order
+    #   dv.sort_by_index(ascending: false)
+    #   #=> Daru::Vector.new [11, 12, 13], index: [23, 22, 21]
+    def sort_by_index opts={}
+      opts = {ascending: true}.merge(opts)
+      new_arrangement = resort_index(@index.each_with_index, opts)
+      _, new_order = new_arrangement.transpose
```
Shekharrajak: Thanks, @zverok for review. I have [changed](https://github.com/SciRuby/daru/pull/315/commits/5f81687239139c7e756865f4158e2b23dc2d8a7c) the PR accordingly. But I didn't understand this error : https://travis-ci.org/SciRuby/daru/jobs/212725530#L813

zverok: Looks like random weirdness  (e.g. rubygems server was down/unresponsive). Just wait a bit and restart travis.

Shekharrajak:
@zverok , I think now this PR is good to go.

Shekharrajak:
@v0dro , have a look once.

PR #321, opened 2017-03-15
Update specs to use modern style

dshvimer:
Updated the specs to reflect a cleaner style of rspec testing. For info
about spec style visit http://betterspecs.org/

### DIFF
--- a/spec/io/sql_data_source_spec.rb
+++ b/spec/io/sql_data_source_spec.rb
@@ -14,57 +14,58 @@ RSpec.describe Daru::IO::SqlDataSource do
     ActiveRecord::Base.connection
   end
 
+  let(:dat_file) do
+    'spec/fixtures/bank2.dat'
+  end
+
   let(:query) do
     'select * from accounts'
   end
 
+  let(:source) do
+    active_record_connection
+  end
+
   describe '.make_dataframe' do
+    subject(:df) { Daru::IO::SqlDataSource.make_dataframe(source, query) }
+
     context 'with DBI::DatabaseHandle' do
-      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(dbi_handle, query) }
+      let(:source) { dbi_handle }
       it { is_expected.to be_a(Daru::DataFrame) }
-      it { expect(df.nrows).to eq 2 }
-      it { expect(df.row[0][:id]).to eq 1 }
-      it { expect(df.row[0][:name]).to eq 'Homer' }
+      it { expect(df.row[0]).to have_attributes(id: 1) }
+      it { expect(df.row[0]).to have_attributes(age: 20) }
+      its(:nrows) { is_expected.to eq 2 }
     end
 
     context 'with ActiveRecord::Connection' do
-      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(active_record_connection, query) }
+      let(:source) { active_record_connection }
       it { is_expected.to be_a(Daru::DataFrame) }
-      it { expect(df.nrows).to eq 2 }
-      it { expect(df.row[0][:id]).to eq 1 }
-      it { expect(df.row[0][:name]).to eq 'Homer' }
+      it { expect(df.row[0]).to have_attributes(id: 1) }
+      it { expect(df.row[0]).to have_attributes(age: 20) }
+      its(:nrows) { is_expected.to eq 2 }
     end
 
     context 'with path to sqlite3 file' do
-      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(db_name, query) }
+      let(:source) { db_name }
       it { is_expected.to be_a(Daru::DataFrame) }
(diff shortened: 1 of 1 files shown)


dshvimer:
Currently having an issue with Rubocop:
```
Metrics/BlockLength: Block has too many lines.
```
Rubocop is complaining about the length of the ```Rspec.describe``` block. I can't imagine being able to reduce it's size. Just figured I'd owuld mention that.

Otherwise this file is ready for review. Let me know if the style is suitable and I will proceed to work on more specs.

spec/io/sql_data_source_spec.rb
```diff
@@ -20,47 +20,19 @@
...
-      end
-    end
-
-    context 'with an object not a database connection' do
-      it 'raises ArgumentError' do
-        expect {
-          Daru::IO::SqlDataSource.make_dataframe(Object.new, query)
-        }.to raise_error(ArgumentError)
-      end
+      subject(:df) { Daru::IO::SqlDataSource.make_dataframe(active_record_connection, query) }
```
zverok: This could be written once, at the very beginning, as `subject(:df) { described_class.make_dataframe(source, query) }`,
and then you just `let(:source)` to different values in different contexts.

spec/io/sql_data_source_spec.rb
```diff
@@ -20,47 +20,19 @@
...
+      it { is_expected.to be_a(Daru::DataFrame) }
+      it { expect(df.nrows).to eq 2 }
```
zverok: its(:nrows) { is_expected.to eq 2 }

spec/io/sql_data_source_spec.rb
```diff
@@ -20,47 +20,19 @@
...
+      it { expect(df.row[0][:id]).to eq 1 }
+      it { expect(df.row[0][:name]).to eq 'Homer' }
```
zverok: `it { expect(df.row[0]).to have_attributes ...`
Better communicates that we are testing the same object (first row).

spec/io/sql_data_source_spec.rb
```diff
@@ -71,13 +43,28 @@
...
+          Daru::IO::SqlDataSource.make_dataframe(active_record_connection, Object.new)
         }.to raise_error(ArgumentError)
       }
     end
 
+    context 'with an object not a database connection' do
+      it {
+        expect {
+          Daru::IO::SqlDataSource.make_dataframe(Object.new, query)
+        }.to raise_error(ArgumentError)
```
zverok: If you'd follow my suggestion about common subject, this context could be turned into
```ruby
let(:source) { Object.new }
it { expect { subject }.to raise_error(ArgumentError) }
```

And I'd suggest to check error message too.

zverok:
> Rubocop is complaining about the length of the Rspec.describe block. I can't imagine being able to reduce it's size. Just figured I'd would mention that.

And yes, block length limitation has no value for specs. It could be disabled in local `specs/.rubocop.yml`

dshvimer:
I made the requested changes. I had an issue testing the ```:name``` property.

```
df.row[0].name #returns 0
df.row[0][:name] #returns "Homer"
```

It seems that the ```have_attributes``` method tests the first variation. I opted to test the ```:age``` column instead.

spec/io/sql_data_source_spec.rb
```diff
@@ -14,70 +14,58 @@
     ActiveRecord::Base.connection
   end
 
+  let(:dat_file) do
+    'spec/fixtures/bank2.dat'
+  end
```
zverok: It is used only in one context -- therefore should be initialized there.

zverok: And same about all others (dbi_handle, active_record_connection). In fact, you can just do in an appropriate context:
```ruby
let(:source) { DBI.connect("DBI:SQLite3:#{db_name}") }
```
...and that's it.

spec/io/sql_data_source_spec.rb
```diff
@@ -14,70 +14,58 @@
...
-          expect {
-            Daru::IO::SqlDataSource.make_dataframe(dbi_handle, Object.new)
-          }.to raise_error(ArgumentError)
-        end
-      end
+    context 'with DBI::DatabaseHandle' do
+      let(:source) { dbi_handle }
+      it { is_expected.to be_a(Daru::DataFrame) }
+      it { expect(df.row[0]).to have_attributes(id: 1) }
+      it { expect(df.row[0]).to have_attributes(age: 20) }
```
zverok: Join those two: `it { expect(df.row[0]).to have_attributes(id: 1, age: 20) }`, it is logically just one test "we have right row here".

dshvimer:
@zverok Thank you for the feedback. I moved the declarations into the contexts. The ActiveRecord connection is the default source since it is used twice. So far we have gone from 83 lines to 56 .

Issue #308, opened 2017-02-10
Add tests for testing block_given? functionality of DataFrame.from_csv() method.


v0dro:
The `DataFrame.from_csv` method currently has a provision for accepting blocks and performing some manipulation on a row that has been read before loading the data into a dataframe.

However, there are no tests in `io_spec.rb` for testing this.

Tests should amply test error conditions too.

zverok:
@v0dro is there a real life reason for this functionality, in the first place?

v0dro:
I recently saw a PR that attempted to remove the line containing `block_given?` and the specs passed since there was no test of this sort.

zverok:
Yes, I understand why we need the test, if functionality exists.
But my question is WHY the functionality exists?

v0dro:
Say you have a CSV column that contains dates in the form `DD MONTH YEAR` (example `12 february 2016`) and you want to convert this to a `DateTime` when you read the file to a dataframe by using your own conversion logic. The easiest way to do it would be by passing a block to `from_csv` that can modify the data as it comes.

gnilrets:
I process a lot of CSV files and have gotten into the habbit of reading all fields as strings and doing conversions after it's built into a dataframe.

zverok:
Got it, thanks :+1:

v0dro:
@gnilrets for smaller dataframes and simpler usage scenarios I think passing a block is more readable and straightforward.

GusAndrianos:
I want to work on this. I'm looking for something I can do for GSoC and I think it's a good fit.

v0dro:
@GusAndrianos yes this would be a great and simple issue to start with. Have you had a look at the source code yet? You should hurry up with your proposal since the deadline for submitting the final proposal is 4th April.

GusAndrianos:
@v0dro This wasn't what I had in mind for GSoC so having to submit patches for every organization I am interested in kind of caught me off guard. I'll try to solve this quickly as this is the only thing missing from my proposal. :)

GusAndrianos:
@v0dro That's awesome, I haven't really found anything that fits me better than SciRuby.

GusAndrianos:
I am a bit confused. My test should check if a block of code can be passed and perform some action on a row already imported right?
How do I pass this row into the `from.csv` method? 

Can you give a usage example? I'm stuck on this for a while now.

parthm:
Can we close this (based on discussion on #413)? #428 has been filed removal of block support.

PR #326, opened 2017-03-27
Partial solution to Issue #308

GusAndrianos:
This is partially solving #308 
I wasn't able to test the functionality using the method `from_csv_hash_with_headers` because it wouldn't change the original values it was given, but only the local ones. So trying to increment the number I was testing by one, was impossible for the original value. 

I found that I needed to pass the `&block` argument to `from_csv` and it's children to pass blocks correctly as it did not work otherwise. 

I passed a very simple block, the file contains two lines and four fields, I am unsure if that's enough because it doesn't check many cases, it just checks if the functionality is there, and it's up to the user to go crazy with it.

This is my first contribution ever, I'd love to hear your thoughts.

### DIFF
--- a/spec/fixtures/block_given_example.csv
+++ b/spec/fixtures/block_given_example.csv
@@ -1,0 +1,2 @@
+This,Is,A,Test
+To,Test,A,Method


zverok:
I believe, that solving the issue "there are no test for `<featurename>`" includes some analysis for "how `<featurename>` should be used".

For all I can understand from discussion in https://github.com/SciRuby/daru/issues/308, expected usage of block passing to `from_csv` is "additional processing for each row".

So, "intuitive" example would be something like this:

```ruby
df = Daru::DataFrame.from_csv('path/to/some.csv') { |row| 
 [row[0].to_i, row[1] || 'Default', Date.parse(row[2]) .... 
}
```

...which _probably_ means you should not only write tests but adjust functionality itself.

GusAndrianos:
@zverok You're right that's how it should work.

v0dro:
Closed due to no further progress.

Issue #130, opened 2016-05-20
Initialize DataFrame without specifying order


lokeshh:
I was working with DataFrame initialization and think following would be convenient to have. What are your thougts?

``` ruby
# Initialization with array of vectors or arrays without specifying order.
>>> a = b = (1..3).dv
>>> Daru::DataFrame.new [a, b]
#<Daru::DataFrame:27152780 @name = 9e6a7851-96f6-43cb-93af-3ef59081e9a2 @size = 3>
                    0          1 
         0          1          1 
         1          2          2 
         2          3          3 

# Initialization with array of Enumerable like ranges without specifying order.
>>> Daru::DataFrame.new [1..3, 1..3]
#<Daru::DataFrame:26079900 @name = b0fb36c0-fd41-4a7e-9a02-36823cc9e1a5 @size = 3>
                    0          1 
         0          1          1 
         1          2          2 
         2          3          3 
```

v0dro:
The first approach is alright (one with Vectors).

athityakumar:
Fixed with PR #329 - I think this issue can be closed now. :smile:

PR #329, opened 2017-04-07
Adds default order for Daru::DataFrame (issue #130)

athityakumar:
Rspec test added, has been written in the older format to main consistency. This will be easier to change while re-writing all tests to modern specifications file-wise, rather than modernizing specs for individual functions. RDoc documentation has also been added.

Tackles issue #130

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -250,10 +250,10 @@ module Daru
     #   # =>
     #   # #<Daru::DataFrame: bat_man (4x2)>
     #   #             0          1
-    #   #  0          6          1
-    #   #  1          7          2
-    #   #  2          8          3
-    #   #  3          9          4
+    #   #  0          1          6
+    #   #  1          2          7
+    #   #  2          3          8
+    #   #  3          4          9
 
     def initialize source, opts={} # rubocop:disable Metrics/MethodLength
       vectors, index = opts[:order], opts[:index] # FIXME: just keyword arges after Ruby 2.1
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -929,7 +929,7 @@ module Daru
 
     # creates a new vector with the data of a given field which the block returns true
     def filter_vector vec, &block
-      Daru::Vector.new each_row.select(&block).map { |row| row[vec] }
+      Daru::Vector.new(each_row.select(&block).map { |row| row[vec] })
     end
 
     # Iterates over each row and retains it in a new DataFrame if the block returns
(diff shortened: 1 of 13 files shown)


lib/daru/dataframe.rb
```diff
@@ -2278,8 +2289,10 @@ def initialize_from_array source, vectors, index, opts
 
       case source.first
       when Array
+        vectors = (0..source.size-1).to_a if vectors.nil?
```
zverok: Why not `||=`?

athityakumar:
@zverok @v0dro - After closer inspection of the logs of the last successful build and current failing build of Travis CI, I observe that `bundle install` used to install `rubocop v0.47.1` (build passing) and is now installing `rubocop v0.48.0` (build failing). Have a look at the respective [releases](https://github.com/bbatsov/rubocop/releases) of rubocop gem.

We can specify `rubocop (~>= 0.47.1)` in the `Gemfile.lock` and remove the `Gemfile.lock` from `.gitignore`. Do you think this would fix the build error, or could it be something else?

zverok:
> We can specify rubocop (~>= 0.47.1) in the Gemfile.lock and remove the Gemfile.lock from .gitignore. Do you think this would fix the build error, or could it be something else?

Instead of that, we can fix **current** rubocop version offences on each PR. They are reasonable, and being a small project we can't expect dedicated volunteers to "update rubocop" in separate PR. So whoever met new offences first is that unfortunate person who fixes them :)

zverok:
(but you can exclude Rakefile and Gemfile from rubocop-checked files, of course)

athityakumar:
@zverok - Alright, I'm taking one for the team. :)

Rubocop offenses are fixed, and build is finally successful. It does feel good to see the `Build passing` and green tick again. :tada:

lib/daru/dataframe.rb
```diff
@@ -244,6 +244,17 @@ def from_vector_rows source, opts
     #   #  b          7          2
     #   #  c          8          3
     #   #  d          9          4
+    #
+    #   df = Daru::DataFrame.new([[1,2,3,4],[6,7,8,9]], name: :bat_man)
+    #
+    #   # =>
+    #   # #<Daru::DataFrame: bat_man (4x2)>
+    #   #             0          1
```
v0dro: Given your example, shouldn't `[6,7,8,9]` be the 1st vector?

athityakumar: Thanks for pointing this out - fixed! :smile:

v0dro: Didn't you use actual executed code for the example? Its better if you do that.

athityakumar: I hadn't used the actual executed code in the initial commit (I had copy-pasted the previous `rdoc` example - my bad). I've added the output I get in the `irb` console while initializing the DataFrame given in the `rdoc` example, in the subsequent commit.

Issue #175, opened 2016-06-26
Transform a vector at values passing a particular condition


lokeshh:
`#where` is used to list vectors that pass a particular condition. May be it would be equally good to have something in Daru that lets you do transformation to a vector for values that specify a particular condition.

It would work as follow:
```ruby
> dv = Daru::Vector.new ['3 days', '5 weeks', '2 weeks']
> dv2 = dv.apply_where(dv.match /weeks/) do |val|
>   convert_to_week val
> end
> dv2\#<Daru::Vector:17000420 @name = nil @metadata = {} @size = 3 >
            nil
      0  3 days
      1 35 days
      2 14 days

```
`dv.match` comes from [this](https://github.com/v0dro/daru/issues/174) feature request and I assume `convert_to_week` is a method to convert week to days.

v0dro:
Nifty.

Issue #174, opened 2016-06-26
Regular expression matching for vectors


lokeshh:
I feel like we should have a function to match regular expression like we have `eq`, `lt`, etc.

It would work as follow:
```ruby
> dv = Daru::Vector.new ['abc', 'aaa', 'xet', 'ccc']
> dv.where(dv.match /a/)
#<Daru::Vector:17216960 @name = nil @metadata = {} @size = 2 >
    nil
  0 abc
  1 aaa
```

gnilrets:
What about just a generic function-to-boolean array process.  Something like
````ruby
dv.where(dv.recode_bool { |v| v.match /a/ })
````

`dv.recode_bool` would take anything "truthy" and convert it to `true`, otherwise `false`. 

That being said.  I would also prefer not having to rewrite `dv` numerous times while filtering.  My dataframe names are usually pretty long and it's a bit annoying to have to reuse these long names all of the time.  What I'd like to see is instead of

````ruby
very_long_data_vector_thing.where(very_long_data_vector_thing.eq(5))
````

it would be nice to have
````ruby
very_long_data_vector_thing.where.eq(5)
````

v0dro:
`recode_bool` looks good to me. It's very generic and can be used in a variety of use cases. The `#match` function can be a wrapper over that just calls the regexp matcher on each element of the Vector internally using `#recode_bool`.

@gnilrets your implementation of `where` might be feasible for Vectors but for DataFrames it would be tough since you need to select the vector which you want to use for the comparison.

For vectors, maybe we can invoke a new `VectorBoolMatcher` (or named something like that) object in case the user has not specified any arguments, which means that `#where` is being used for chaining with `eq` and others.

PR #333, opened 2017-04-15
Adds Daru::Vector#match and Daru::Vector#apply_where (issue #175)

athityakumar:
Rspec test added, has been written in the older format to main consistency. This will be easier to change while re-writing all tests to modern specifications file-wise. RDoc documentation has also been added. `Daru::Vector#match` uses Regexp match via the `=~` operator.

Tackles issue #175

### DIFF
--- a/lib/daru/core/query.rb
+++ b/lib/daru/core/query.rb
@@ -70,11 +70,10 @@ module Daru
           resultant_dv
         end
 
-        def vector_apply_where dv, bool_array, &block
+        def vector_apply_where dv, bool_array
           _data, new_index = fetch_new_data_and_index dv, bool_array
-          all_index = dv.index.to_a
-          yield_data = dv.map(&block).to_a
-          all_data = all_index.map { |idx| new_index.include?(idx) ? yield_data[idx] : dv[idx] }
+          all_index        = dv.index
+          all_data         = all_index.map { |idx| new_index.include?(idx) ? yield(dv[idx]) : dv[idx] }
 
           resultant_dv = Daru::Vector.new all_data,
             index: dv.index.class.new(all_index),
--- a/lib/daru/core/query.rb
+++ b/lib/daru/core/query.rb
@@ -70,6 +70,23 @@ module Daru
           resultant_dv
         end
 
+        def vector_apply_where dv, bool_array, &block
+          _data, new_index = fetch_new_data_and_index dv, bool_array
+          all_index = dv.index.to_a
+          yield_data = dv.map(&block).to_a
+          all_data = all_index.map { |idx| new_index.include?(idx) ? yield_data[idx] : dv[idx] }
+
+          resultant_dv = Daru::Vector.new all_data,
+            index: dv.index.class.new(all_index),
+            dtype: dv.dtype,
+            type: dv.type,
+            name: dv.name
+
+          # Preserve categories order for category vector
+          resultant_dv.categories = dv.categories if dv.category?
+          resultant_dv
+        end
+
         private
 
         def fetch_new_data_and_index dv, bool_array
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -1200,14 +1200,16 @@ module Daru
     # +regexp+ matching with the given array.
     #
     # @param Regexp A regular matching expression. For example, +/weeks/+.
-    # @return [Array] Conatining either +nil+ or integer values, according to the match with the given +regexp+
-    # @example
-    #   dv = Daru::Vector.new ['3 days', '5 weeks', '2 weeks']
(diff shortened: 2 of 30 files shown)


athityakumar:
@v0dro @zverok - Following the discussions on the issues #175 and #174 , `#match` support has been provided only to `Daru::Vector`. Do review the PR. :smile:

athityakumar:
@v0dro @zverok @lokeshh - Following the discussions on the issues #175 and #174 , #match support has been provided only to Daru::Vector. Do review the PR. :smile:

athityakumar:
@zverok @v0dro : Just wanted to remind you about this PR - kindly review whenever you're free. :slightly_smiling_face:

zverok:
Reviewing it now.

lib/daru/vector.rb
```diff
@@ -1141,6 +1159,20 @@ def jackknife(estimators, k=1) # rubocop:disable Metrics/AbcSize,Metrics/MethodL
...
+    #
+    # @param Regexp A regular matching expression. For example, +/weeks/+.
+    # @return [Array] Conatining either +nil+ or integer values, according to the match with the given +regexp+
+    # @example
+    #   dv = Daru::Vector.new ['3 days', '5 weeks', '2 weeks']
+    #   dv.match /weeks/
+    #
+    #   # => [nil, 2, 2]
+    def match(regexp)
+      @data.map { |x| x =~ regexp }
```
zverok: Let's use [===](http://www.rubyletter.com/blog/2017/01/18/ruby-triple-equals-operator.html) operator. This way we can also do `vector.match(1..20)` or `vector.match(Numeric)`.

spec/core/query_spec.rb
```diff
@@ -334,4 +334,14 @@
...
   end
+  describe "apply_where" do
+    context Daru::Vector do
+      it "matches regexp with a block input" do
+        dv = Daru::Vector.new ['3 days', '5 weeks', '2 weeks']
+        dv = dv.apply_where(dv.match /weeks/) { |x| "#{x.split.first.to_i * 7} days" }
+        expected_dv = Daru::Vector.new ['3 days', '35 days', '14 days']
+        expect(dv).to eq expected_dv
+      end
+    end
```
zverok: a) modern tests please (with `subject` and one-statement checks, see betterspecs.org)
b) you need to check several conditions, like "empty boolean vector", "block returning something unexpected" and so on.

spec/vector_spec.rb
```diff
@@ -2009,6 +2009,13 @@ def expect_correct_tokens hash
     end
   end
 
+  context '#match' do
+      it 'returns matching array for a given regexp' do
+        dv = Daru::Vector.new ['3 days', '5 weeks', '2 weeks']
+        expect(dv.match /weeks/).to eq [nil, 2, 2]
+      end
```
zverok: Same as above: modern specs and several conditions.

zverok: BTW, if I understand correctly, it is better to have only `false` and `true` values returned (though, if you'll follow `===` advice, you will have this automatically).

lib/daru/core/query.rb
```diff
@@ -70,6 +70,23 @@ def vector_where dv, bool_array
           resultant_dv
         end
 
+        def vector_apply_where dv, bool_array, &block
+          _data, new_index = fetch_new_data_and_index dv, bool_array
+          all_index = dv.index.to_a
+          yield_data = dv.map(&block).to_a
+          all_data = all_index.map { |idx| new_index.include?(idx) ? yield_data[idx] : dv[idx] }
```
zverok: Are you sure it is a sane approach (apply block to all data, and then select appropriate)? 
Imagine there is 10k values vector, and "where" returns only 5 values of all 10k, and calculation in `block` is expensive (some serious math, or even reading some files). "Right" implementation should call block 5 times, not 10k times and then select only 5 values.

v0dro:
Whats the update with the PR?

v0dro:
Please finish it off ASAP.

spec/core/query_spec.rb
```diff
@@ -334,4 +334,14 @@
       end
     end
   end
+
+  describe "apply_where" do
+    context "matches regexp with block input" do
+      subject { dv.apply_where(dv.match /weeks/) { |x| "#{x.split.first.to_i * 7} days" } }
```
athityakumar: @zverok - I remember seeing DSL support being added to `which` in PR #396. Should a similar DSL call be supported here, like `apply_which`?

athityakumar:
@zverok - Please review when you're free. 😄

v0dro:
I think @zverok is busy with refactoring :P

Issue #317, opened 2017-03-05
Enchance specs style 


zverok:
Our specs had grown "organically", and many of them use outdated style of testing things (all bootstrap & several checks is made inside single example, tons of hard-to-read large objects created on the fly, instance variables are used instead of `let` and so on).

What should be done in this ticket is:
* Refactoring of specs in accordance with [BetterSpecs](http://betterspecs.org/) guidelines;
* Making specs more clean/DRY (only several different dataframes are created as a test subjects, covering most of the situations)
* Probably, splitting large specs files into smaller ones (e.g. for dataframe -- basic functionality, math, output, import/export...)
* Consider turning on Rubocop for specs files (with probably some cops turned off or tuned in `spec/.rubocop.yml`)
* Consider introducing [rubocop-rspec](https://github.com/backus/rubocop-rspec) for rspec-specific checks (initially turned off, then turned on file-by-file & cop-by-cop)

As the full solution for the ticket will touch _all_ `spec/**/*.rb` files, it is advised to make several PRs (one PR per one spec file, or one PR per one problem solution).

You can rely on my (@zverok) advices and reviews while working on this ticket.

dshvimer:
I am interested in tackling this. I will be working on it this week. Is one PR per spec file ok?

dshvimer:
I think that since this a fairly large task. More than one person can contribute at any given time. I will list the files that I am working on as I begin to work on them and others should do the same.

zverok:
Yes, please list your progress & PRs just in this issue.

dshvimer:
```
| File  | Status | PR |
| ------------- | ------------- | ------------- |
| spec/io/sql_data_source_spec.rb  | In Progress  |  |
```

v0dro:
@dshvimer any update on this? Are you going to work on `io_spec`?

PR #334, opened 2017-04-21
Adds remote URL for Daru::DataFrame#from_csv

athityakumar:
@v0dro @zverok - This is a continuation of @anshuman23's PR #300. As there hasn't been any activity for quite some time, I've made the changes suggested there and have opened this PR.

### DIFF
--- a/lib/daru/io/io.rb
+++ b/lib/daru/io/io.rb
@@ -76,7 +76,6 @@ module Daru
       # Functions for loading/writing CSV files
       def from_csv path, opts={}
         daru_options, opts = from_csv_prepare_opts opts
-
         # Preprocess headers for detecting and correcting repetition in
         # case the :headers option is not specified.
         hsh =
@@ -86,7 +85,6 @@ module Daru
             from_csv_hash(path, opts)
               .tap { |hash| daru_options[:order] = hash.keys }
           end
-
         Daru::DataFrame.new(hsh,daru_options)
       end
 
@@ -201,23 +199,32 @@ module Daru
 
       def from_csv_hash_with_headers(path, opts)
         opts[:header_converters] ||= :symbol
-
-        ::CSV
-          .read(path, 'rb',opts)
-          .tap { |c| yield c if block_given? }
-          .by_col.map { |col_name, values| [col_name, values] }.to_h
+        url = URI.parse(path)
+        if %w(http https).include?(url.scheme)
+          ::CSV
+            .parse(open(path), opts)
+            .by_col.map { |col_name, values| [col_name, values] }.to_h
+        else
+          ::CSV
+            .read(path, 'rb',opts)
+            .tap { |c| yield c if block_given? }
+            .by_col.map { |col_name, values| [col_name, values] }.to_h
+        end
       end
 
       def from_csv_hash(path, opts)
+        url = URI.parse(path)
         csv_as_arrays =
-          ::CSV
-          .open(path, 'rb', opts)
-          .tap { |c| yield c if block_given? }
-          .to_a
-
+          if %w(http https).include?(url.scheme)
+            ::CSV.parse(open(path), opts)
+          else
+            ::CSV
+              .open(path, 'rb', opts)
+              .tap { |c| yield c if block_given? }
+              .to_a
+          end
         headers       = ArrayHelper.recode_repeated(csv_as_arrays.shift)
         csv_as_arrays = csv_as_arrays.transpose
-
         headers.each_with_index.map { |h, i| [h, csv_as_arrays[i]] }.to_h
       end
     end
--- a/lib/daru/io/io.rb
+++ b/lib/daru/io/io.rb
@@ -199,30 +199,18 @@ module Daru
 
       def from_csv_hash_with_headers(path, opts)
         opts[:header_converters] ||= :symbol
-        url = URI.parse(path)
-        if %w(http https).include?(url.scheme)
-          ::CSV
-            .parse(open(path), opts)
-            .by_col.map { |col_name, values| [col_name, values] }.to_h
-        else
-          ::CSV
-            .read(path, 'rb',opts)
-            .tap { |c| yield c if block_given? }
-            .by_col.map { |col_name, values| [col_name, values] }.to_h
-        end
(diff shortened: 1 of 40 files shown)


spec/io/io_spec.rb
```diff
@@ -1,39 +1,70 @@
...
       it "follows the order of columns given in CSV" do
-        df = Daru::DataFrame.from_csv 'spec/fixtures/sales-funnel.csv'
-        expect(df.vectors.to_a).to eq(%W[Account Name Rep Manager Product Quantity Price Status])
+        file = 'sales-funnel'
+        paths = [local_csv_url(file), remote_csv_url(file)]
+        paths.each do |path|
+          df = Daru::DataFrame.from_csv path
+          expect(df.vectors.to_a).to eq(%W[Account Name Rep Manager Product Quantity Price Status])
+        end
       end
```
zverok: Could you please do a proper specs here, with contexts, lets and subjects? We need to see what _exactly_ is tested, and for this we try to follow http://betterspecs.org/ in the new specs.

lib/daru/io/io.rb
```diff
@@ -201,25 +199,29 @@ def from_csv_prepare_opts opts
...
           .tap { |c| yield c if block_given? }
           .to_a
-
         headers       = ArrayHelper.recode_repeated(csv_as_arrays.shift)
         csv_as_arrays = csv_as_arrays.transpose
-
         headers.each_with_index.map { |h, i| [h, csv_as_arrays[i]] }.to_h
       end
+
+      def new_open(url)
```
zverok: I don't feel this method's name is communicative.

lib/daru/io/io.rb
```diff
@@ -201,25 +199,29 @@ def from_csv_prepare_opts opts
...
+        open(url)
+      rescue URI::InvalidURIError
+        host = url.match(".+\:\/\/([^\/]+)")[1]
+        Net::HTTP.get host, url.partition(host)[2] || '/'
```
athityakumar: Have a look at this issue - [Open-URI can't open URLs with underscores](https://bugs.ruby-lang.org/issues/8241). Though the issue has been closed, the scheme has been updated only from Ruby 2.2 onwards - but our Travis CI build takes earlier versions too into consideration.

Rather, this workaround `new_open` can open URLs with underscores too. Though this might be a discrete case, I wouldn't want to assume that NO user will ever use excel files from URLs that have underscores.

I am thinking of renaming the `new_open` function to `open_with_underscore` - please feel free to share your choice for the function name.

spec/io/io_spec.rb
athityakumar: This is something I wanted to discuss regarding the specs for local file and remote URL testing.  Previously, there were 4 tests for the `#from_csv` module - all deal with the local CSV files. Now, should we have 4 more tests by reading from the remote URLs, or just 1 more test to check the equivalence of local CSV files with respective dummy remote CSV files?

I personally prefer the latter, but we can definitely discuss about this. Also, wouldn't file-wise spec enhancements be easier to do, so that the specs are uniform at any given point of time?

spec/io/io_spec.rb
```diff
@@ -1,39 +1,70 @@
...
     end
-
+    
+
```
athityakumar: Roger that :smile:

lib/daru/io/io.rb
zverok: @athityakumar I believe, this approach is redundand. 
Daru is pretty complicated piece of software, and each new complication should be taken with a grain of salt (also known as YAGNI). But we now have in the core of library additional method which targets:
a) rare cases (underscored domains);
b) ...for outdated Ruby version (Daru currently supports 2.0+, yet we'll drop 2.0 soon, and 2.1 in a few months, and in 2.2 the problem is fixed already);
c) ...without much explanations.

Considering all of this, I'd say just drop this trick completely. If we'll have an issue saying "this VERY IMPORTANT research data I need is on obscure_domain.some_large_univercity.com", we'll just answer "It is known Ruby bug, just download data manually or update your Ruby version", and that would be it.

zverok: Even popular Ruby HTTP clients (for ex., HTTParty) fail on 2.1 and underscored domains, and they consider it acceptable. Why shouldn't we?

spec/io/io_spec.rb
zverok: >  Now, should we have 4 more tests by reading from the remote URLs

One should be enough. We know that internally it uses the same logic, just takes the data from URL, so no need to test whether it parses headers properly and so on.

> Also, wouldn't file-wise spec enhancements be easier to do, so that the specs are uniform at any given point of time?

Yes, we even have a ticket for this: https://github.com/SciRuby/daru/issues/317 without much volunteers though. I am not sure that rewriting entire 300 lines of io_spec.rb should go to this PR, as there are several concerns will be mixing. If you have a lot of enthusiasm, you can do it in the PR, though, but if not, just make your new specs pretty.

lib/daru/io/io.rb
athityakumar: I do understand that there won't be use for this trick / workaround from Ruby 2.2 onwards. And yes, we could address any issue about domains with underscore, but then the user will have to have the workaround in their code would have to override our existing #from_csv function - which is what I was trying to avoid. I was trying to avoid any hassle that the user may face, but this works too. :smile:

zverok: Well, this mindset is praiseworthy, yet it can easily lead us into integrating wrong data encoding guessers, redirect handlers and generic `Daru.find(data_about: 'pretty kittens')` things. So we do YAGNI when we can, unless this aproach is proved to be wrong (for this particular problem: IF we'll know there is significant number of daru users on Ruby 2.1 AND significant number of their data on cool_underscored_domains, we could rethink our approach).

Currently, it looks for me as small, yet existing complication with very low value for users. (BTW, even if implementing this; for this kind of things, like "fix the bug for particular Ruby version", I'd prefer just core ext with `if RUBY_VERSION < '2.2'` guard).

zverok:
Thanks @athityakumar & @anshuman23 !

Issue #204, opened 2016-07-31
Better error message to address invalid column in dataframe


lokeshh:
``` ruby
> df = Daru::DataFrame.new({
>   a: [1, 2, 3],
>   b: [4, 5, 6]
> })
> df[:c]
IndexError: Specified index :c does not exist
```

Instead of saying "Specified index `:c` does not exist", I think it would be better to say "Specified vector `:c` does not exist".

PR #335, opened 2017-04-23
Better error messages when a vector is missing from dataframe

sivagollapalli:
#204

### DIFF
--- a/spec/index/index_spec.rb
+++ b/spec/index/index_spec.rb
@@ -362,5 +362,13 @@ describe Daru::Index do
         it { is_expected.to eq klass.new(answer) }
       end
     end
+
+    context "#by_single_key" do
+      let(:df) { Daru::DataFrame.new(a: [1, 2, 3], b: [4, 5, 6] ) }
+
+      it 'raise error when vector is missing from dataframe' do
+	expect { df[:c] }.to raise_error(IndexError, /Specified vector :c does not exist/)
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/index/index_spec.rb
```diff
@@ -362,5 +362,13 @@
...
       end
     end
+
+    context "#by_single_key" do
+      let(:df) { Daru::DataFrame.new(a: [1, 2, 3], b: [4, 5, 6] ) }
+
+      it 'raise error when vector is missing from dataframe' do
+	expect { df[:c] }.to raise_error(IndexError, /Specified vector :c does not exist/)
+      end
+    end
```
athityakumar: @zverok - Should this test be shifted to `spec/dataframe_spec.rb`?

v0dro:
@sivagollapalli you've made changes to index.rb and are displaying an error message only for Dataframes with missing vectors. However, an Index is used by both Vector and DataFrame and therefore returning an error message stating a vector is missing from `index.rb` is not justified.

If you want better error reporting, you should catch the IndexError in the DataFrame class and change the error message to reflect the fact that a vector is missing from a dataframe.

spec/index/index_spec.rb
v0dro: Yes this test should be in dataframe_spec.

sivagollapalli:
@v0dro @zverok Could you check the changes?

zverok:
Please fix Rubocop (https://travis-ci.org/SciRuby/daru/jobs/225581190, you can check it locally by running `rubocop` command) before merge.

Issue #242, opened 2016-08-31
Suppress warnings if Daru.error_stream is nil


v0dro:
We should have a module variable called `Daru.error_stream` whose default will be set to `STDERR`, and which can be changed to `nil`. If the user sets it to `nil`, all warnings that are currently printed with `STDERR.puts` should be suppressed.

sivagollapalli:
@v0dro Apart from https://github.com/SciRuby/daru/blob/master/lib/daru.rb#L87 nowhere we are using `STDERR.puts` to print warning messages. I think we can allow an error message if gem didn't installed. Hence I avoided.

v0dro:
@sivagollapalli the reason for this issue is that the errors show up in people's logs every time they require daru in some app. Therefore it makes sense to be able to set a switch.

Setting the switch to `nil` will supress warnings once and for all.

baarkerlounger:
I think this can be closed given the PR merged above?

PR #343, opened 2017-05-09
[#242] Added error stream to daru

sivagollapalli:
#242

### DIFF
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -38,11 +38,13 @@ module Daru
 
   @plotting_library = :nyaplot
 
+  @@error_stream = $stderr
+
   class << self
     # A variable which will set whether Vector metadata is updated immediately or lazily.
     # Call the #update method every time a values are set or removed in order to update
     # metadata like positions of missing values.
-    attr_accessor :lazy_update, :error_stream
+    attr_accessor :lazy_update
     attr_reader :plotting_library
 
     def create_has_library(library)
@@ -72,6 +74,14 @@ module Daru
         raise ArgumentError, "Unsupported library #{lib}"
       end
     end
+
+    def error_stream= stream 
+      class_variable_set(:@@error_stream, stream)
+    end
+
+    def error_stream
+      @@error_stream
+    end
   end
 
   create_has_library :gsl
@@ -84,8 +94,8 @@ begin
   gem 'spreadsheet', '~>1.1.1'
   require 'spreadsheet'
 rescue LoadError
-  STDERR.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
-  ' spreadsheet functions.'
+  Daru.error_stream.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
+  ' spreadsheet functions.' if Daru.error_stream
 end
 
 autoload :CSV, 'csv'
--- a/lib/daru.rb
+++ b/lib/daru.rb
@@ -38,13 +38,13 @@ module Daru
 
   @plotting_library = :nyaplot
 
-  @@error_stream = $stderr
+  @error_stream = $stderr
 
   class << self
     # A variable which will set whether Vector metadata is updated immediately or lazily.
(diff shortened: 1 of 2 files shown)


sivagollapalli:
@zverok I thought that I have to use on https://github.com/SciRuby/daru/blob/master/lib/daru.rb#L87 to silence that warning, but since it is an installation error I thought it would be better to show. And also on remaining code base I didn't see any `STDERR.puts` statements. So I didn't make any changes. If you want to silence the error message I would really happy to make that change.

v0dro:
Here's how you can proceed:
- Set the error_stream default to `STDERR`.
- Change all calls to `STDERR` in the code base to `@@error_stream`.
- Write tests accordingly.

sivagollapalli:
@v0dro Since I didn't know how to test spreadsheet dependency in spec so I avoided that. Could you give suggestions regarding that?

lib/daru.rb
```diff
@@ -72,6 +74,14 @@ def plotting_library= lib
...
       end
     end
+
+    def error_stream= stream 
+      class_variable_set(:@@error_stream, stream)
+    end
+
+    def error_stream
+      @@error_stream
+    end
```
zverok: Why did you decided to go with class variable? Instance variable of `Daru` module would be enough -- with those two methods becoming just `attr_accessor :error_stream`

lib/daru.rb
```diff
@@ -84,8 +94,8 @@ def plotting_library= lib
   gem 'spreadsheet', '~>1.1.1'
   require 'spreadsheet'
 rescue LoadError
-  STDERR.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
-  ' spreadsheet functions.'
+  Daru.error_stream.puts "\nInstall the spreadsheet gem version ~>1.1.1 for using"\
+  ' spreadsheet functions.' if Daru.error_stream
```
zverok: I'd abstract this into method like `Daru.error(message)`. Then it would be easy to test.

sivagollapalli:
@zverok Added changes. Could you please check?

spec/daru_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    Daru.error('priting error message')
+    expect($stderr).to have_received(:puts)
+  end
+
+  it 'should not print error message if it is set to nil' do
+    Daru.error_stream = nil
+    allow($stderr).to receive(:puts).and_return(true)
+    Daru.error('priting error message')
+    expect($stderr).not_to have_received(:puts)
+  end
```
zverok:
```ruby
describe '#error' do
  context 'by default' do
    it { expect { Daru.error('test') }.to output('test').to_stderr }
  end

  context 'when set to nil' do
    before { Daru.error_stream = nil }
    it { expect { Daru.error('test') }.not_to output('test').to_stderr }
  end

  context 'when set to instance of custom class' do
    let(:custom_stream) { double(puts: nil) }
    before { Daru.error_stream = custom_stream }

    it 'calls puts' do
      expect { Daru.error('test') }.not_to output('test').to_stderr
      expect(custom_stream).to have_received(:puts).with('test')
    end
  end
end
```
Or something like that.

spec/daru_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+describe 'Daru' do
+  it 'default error stream should be $stderr' do
+    expect(Daru.error_stream).to eq($stderr)
+  end
+
+  it 'should be able to set error stream to nil' do
+    Daru.error_stream = nil
+    expect(Daru.error_stream).to be_nil
+  end
```
zverok:
```ruby
it { expect { Daru.error_stream = nil }.to change(Daru, :error_stream).to(nil) }
```
Though, this test is completely unnecessary, as you are, in fact, testing Ruby's `attr_accessor` implementation (and the fact of existance of `error_stream=` is tested below anyways).

sivagollapalli:
@zverok Incorporated the changes. Could you please check?

PR #345, opened 2017-05-20
make Daru::DataFrame#covariance faster

genya0407:
Roughly it becomes 2 times faster.

I confirmed that result of `#coveriance` doesn't change.

Should I write some benchmarks or tests?


v0dro:
No that's OK. Merging.

Issue #341, opened 2017-05-01
#tail and #head not working for Daru::DataFrame


pedrorio:
Hello, I was trying to backtest some bitcoin investing strategies and the #tail and #head methods do not work unless I specify a Daru::Vector.

For instance, `p df[0].tail works` but `p df.tail` does not.

I was using daru from rubygems and changed the gem source to this github repo but the #summary method also broke. I am using a Daru::DateTimeIndex index.

Here is my error:
```
/usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/index/index.rb:209:in `key': undefined method `[]' for nil:NilClass (NoMethodError)
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/index/index.rb:151:in `map'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/index/index.rb:151:in `at'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/vector.rb:250:in `at'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/vector.rb:1183:in `reject_values'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/maths/statistics/vector.rb:817:in `midpoint_percentile'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/maths/statistics/vector.rb:352:in `percentile'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/maths/statistics/vector.rb:27:in `median'
	from /usr/local/opt/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/bundler/gems/daru-7b70cfadf4d2/lib/daru/vector.rb:955:in `numeric_summary'
[...]

PR #347, opened 2017-05-26
[#341] fix for dataframe head and tail

cpankaj:


zverok:
Thanks for your contribution!
Could you please add the test that shows what was failing without your patch?
(You can ignore Rubocop offences on Travis, we are fixing it currently)

db579:
I think this can be closed given the changes have been incorporated into another PR and merged now.

PR #348, opened 2017-05-27
Actualize for the new Rubocop

zverok:
Rubocop 0.49.0 added several new cops, which had led to several new offences.

This PR also fixes Rubocop version in `daru.gemspec`. Turns out that new versions are frequent, and small 1-2 lines PR of innocent contributors become severely broken suddenly. I believe now it is wiser to just update the version from time to time, and fix new offences simultaneously.

### DIFF
--- a/daru.gemspec
+++ b/daru.gemspec
@@ -69,7 +69,7 @@ EOF
   spec.add_development_dependency 'dbi'
   spec.add_development_dependency 'activerecord', '~> 4.0'
   spec.add_development_dependency 'sqlite3'
-  spec.add_development_dependency 'rubocop', '>= 0.40.0'
+  spec.add_development_dependency 'rubocop', '~> 0.49.0'
   spec.add_development_dependency 'ruby-prof'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'gruff'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,7 +16,7 @@ AllCops:
   TargetRubyVersion: 2.0
 
 # Preferred codebase style ---------------------------------------------
-Style/ExtraSpacing:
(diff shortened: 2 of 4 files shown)


daru.gemspec
```diff
@@ -69,7 +69,7 @@ EOF
   spec.add_development_dependency 'dbi'
   spec.add_development_dependency 'activerecord', '~> 4.0'
   spec.add_development_dependency 'sqlite3'
-  spec.add_development_dependency 'rubocop', '>= 0.40.0'
+  spec.add_development_dependency 'rubocop', '~> 0.49.0'
```
arbox: @zverok it's a bit of offtopic but I always struggled to understand why could it be useful to have such development dependencies in the Gemspec for projects which use Bundler. The only valid usecase for me was to download the gem and use it as "source package" for development. Do you have a strong opinion about Gemfile vs. Gemspec for development dependencies? Thank you!

zverok: Not really strong (in my own gems I use both, depending on current mood). But exposing development dependencies in `.gemspec` has advantage of "visibility", for example on [rubygems](https://rubygems.org/gems/daru). Also, there are third-party services like Gemnasium ([example](https://gemnasium.com/github.com/molybdenum-99/infoboxer)), which allow to control & notify on dependent gem updates—I believe they use `.gemspec` too.

`Gemfile` could be preferred due to shorter syntax, and better granulated groups.

So, probably the matter of taste and circumstances.

arbox: Ok, i see the point. Thank you :) Gemnasium uses both sources in my experience .

zverok: Never checked that, to be honest :)
And, in fact, in my own gems I am lately prefer `Gemfile`

PR #350, opened 2017-06-03
return Nyaplot plot instance instead of nil. 

Shekharrajak:
This will help me to use the Nyaplot plot object for further configuration like weight, height, etc in daru-view : https://github.com/Shekharrajak/daru-view/pull/3 .

IRuby notebook will still display plot.

### DIFF
--- a/lib/daru/plotting/nyaplot/category.rb
+++ b/lib/daru/plotting/nyaplot/category.rb
@@ -12,7 +12,7 @@ module Daru
             # Set yrange for good view
             set_yrange plot, opts[:method]
             yield plot, diagram if block_given?
-            plot.show
+            plot
           else
             raise ArgumentError, "#{type} type is not supported."
           end
--- a/spec/plotting/nyaplot/vector_spec.rb
+++ b/spec/plotting/nyaplot/vector_spec.rb
@@ -13,20 +13,20 @@ describe Daru::Vector, 'plotting' do
   it 'plots the vector' do
     expect(plot).to receive(:add).with(:box, [11, 22, 33]).ordered
 
-    vector.plot(type: :box)
+    expect(vector.plot(type: :box)).to eq plot
   end
 
   context 'scatter' do
     it 'is default type' do
       expect(plot).to receive(:add).with(:scatter, instance_of(Array), instance_of(Array)).ordered
 
-      vector.plot
+      expect(vector.plot).to eq plot
     end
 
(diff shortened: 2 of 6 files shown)


Shekharrajak:
Ping @zverok @v0dro @lokeshh

lib/daru/plotting/nyaplot/category.rb
```diff
@@ -12,7 +12,7 @@ def plot opts
             # Set yrange for good view
             set_yrange plot, opts[:method]
             yield plot, diagram if block_given?
-            plot.show
```
Shekharrajak: `def plot` returns nothing previously.

spec/plotting/nyaplot/vector_spec.rb
```diff
@@ -37,7 +34,6 @@
     context type.to_s do
       it 'does not set x axis' do
         expect(plot).to receive(:add).with(type, [11, 22, 33]).ordered
-        expect(plot).to receive(:show).ordered
 
         vector.plot(type: type)
```
Shekharrajak: Should I add line 

```
        expect(vector).to receive(:plot).with(type: type).and_return(plot)
        vector.plot(type: type)
```
? Is it correct way to do this ? 
Ping @zverok

Shekharrajak: Is there any other good way for `InstanceDouble` ?

zverok: I believe this should be enough:
```ruby
expect(vector.plot(type: type)).to eq plot
```
It exactly checks that `#plot` method would return an instance of plot constructed, or am I missing something?

Shekharrajak: Fine! Previously I was comparing the class using `be_a`, `eq` , instance of class. Also didn't remove the `vector.plot(type: type)` line, that's why I was getting some `InstanceDouble ` error. Now I have updated the testcases.
Thanks!

v0dro:
@Shekharrajak can you confirm that you've checked that this works in iRuby? I'm good with this change if yes.

Shekharrajak:
@v0dro ,Yes I have checked it. It is working fine.

v0dro:
Ok. Let @zverok approve and we can merge.

Shekharrajak:
Ping @v0dro

Issue #354, opened 2017-06-13
lag should support negative arguments


parthm:
At the moment, if we call ```vector.lag -1``` the vector returned has all nils. It would be better if lag supported positive as well as negative shifts.

```ruby
[16] pry(main)> v = Daru::Vector.new(1..5)
=> #<Daru::Vector(5)>
   0   1
   1   2
   2   3
   3   4
   4   5
[18] pry(main)> v.lag 1
=> #<Daru::Vector(5)>
   0 nil
   1   1
   2   2
   3   3
   4   4
[19] pry(main)> v.lag -1
=> #<Daru::Vector(5)>
   0 nil
   1 nil
   2 nil
   3 nil
   4 nil
[20] pry(main)> 
```

Behavior for [pandas shift function](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.shift.html#pandas.Series.shift) for reference:

```python
In [7]: s = pd.Series(range(5))

In [8]: s
Out[8]: 
0    0
1    1
2    2
3    3
4    4
dtype: int32

In [9]: s.shift(1)
Out[9]: 
0    NaN
1    0.0
2    1.0
3    2.0
4    3.0
dtype: float64

In [10]: s.shift(-1)
Out[10]: 
0    1.0
1    2.0
2    3.0
3    4.0
4    NaN
dtype: float64

```

parthm:
I suppose this can be marked as fixed as its closed in #355

zverok:
:+1:

PR #355, opened 2017-06-13
[#354] Vector::lag supports -ve shift. tests updated.

parthm:
This has the fix for #354 
Tests have been updated. Comments / suggestions welcome.


v0dro:
Looks good. Just resolve the rubocop offences and I'll merge.

parthm:
Rubocop offenses resolved. Pretty much the only way to resolve ABC size too high was to break up the function. 
Added additional test cases for lag 0.

lib/daru/vector.rb
```diff
@@ -762,25 +762,25 @@ def replace_nils! replacement
 
     # Lags the series by k periods.
     #
-    # The convention is to set the oldest observations (the first ones
-    # in the series) to nil so that the size of the lagged series is the
-    # same as the original.
+    # Lags the series by k periods where k can be a positive or negative
+    # integer.
+    # Returns a new series with nil values inserted into the series
+    # keeping the size of the new series the same as the original.
```
zverok: Let's update the docs a bit. Probably this way (minus my broken English):
```ruby
# Lags the series by `k` periods, "shifting" data and inserting `nil`s 
# from beginning or end of a vector, while preserving original vector's size.
#
# `k` could be positive or negative number. If `k` is positive, `nil`s are inserted 
# at the beginning of the vector, otherwise they are inserted at the end.
```

lib/daru/vector.rb
```diff
@@ -762,25 +762,25 @@ def replace_nils! replacement
...
       return dup if k.zero?
-
-      dat = @data.to_a.dup
-      (dat.size - 1).downto(k) { |i| dat[i] = dat[i - k] }
-      (0...k).each { |i| dat[i] = nil }
-
-      Daru::Vector.new(dat, index: @index, name: @name)
+      return Daru::Vector.new([nil]*@data.size, index: @index, name: @name) if k.abs >= @data.size
+      _lag k
     end
```
zverok: WDYT about this implementation? Rubocop is happy, and instead of strangely named method "just to fool rubocop" we got internal useful method (create vector with a new data, and same index/name as a current):
```ruby
    def lag k=1
      case k
      when 0 then dup
      when 1...size
        copy([nil] * k + data.to_a)
      when -size..-1
        copy(data.to_a[k.abs...size])
      else
        copy([])
      end
    end

    private 
    def copy(values)
      # Make sure values is right-justified to the size of the vector
      values.concat([nil] * (size - values.size)) if values.size < size
      Daru::Vector.new(values[0...size], index: @index, name: @name)
    end
```

spec/vector_spec.rb
```diff
@@ -1970,16 +1970,65 @@ def expect_correct_tokens hash
...
+      lag_100 = @xiu.lag -100
+      expect(lag_100).to all( be_nil)
+      expect(lag_100.size).to be @xiu.size
+      expect(lag_100).to be_a(Daru::Vector)
+
+      lag_for_size = @xiu.lag @xiu.size
+      expect(lag_for_size).to all(be_nil)
+      expect(lag_for_size.size).to be @xiu.size
+      expect(lag_for_size).to be_a(Daru::Vector)
     end
```
zverok: Can you please rewrite the test in a modern RSpec (see betterspec.org), with communcative context names and atomic tests?..

parthm: Thanks for the review @zverok . Much appreciated. I am new to rspec. In fact this is the first time I have used this. Could you give an example / link to what might be a good way to use this? I can check that and adjust as needed.

zverok: OK, look (would be long):
```ruby
#1. what are we describing? `lag` method behavior
describe '#lag' do
end

# 2. what do we want to test? Result of lagging...
# a) by zero
# b) default
# c) positive
# d) negative
# e) too large k abs
describe '#lag'
  context 'by default'
  context '0'
  context 'positive'
  context 'negative'
  context 'large'
end

# 3. what is subject of test? Result of lagging!
describe '#lag' do
  let(:source) { described_class.new(....) }

  context 'by default' do
    subject { source.lag }
  end

  subject { source.lag(amount) }
  
  context '0' do
    let(:amount) { 0 }
  end
  
  context 'positive' do
    let(:amount) { 2 }
  end
  
  context 'negative' do
    let(:amount) { -1 }
  end
  
  context 'large positive' do
    let(:amount) { 100 }
  end
  
  context 'large negative' do
    let(:amount) { -100 }
  end
end

# 4. what we expect? to receive properly lagged vector
...
  context 'by default' do
    subject { source.lag }

    it { is_expected.to eq Daru::Vector.new([nil, <values of initial vector>]) }
  end
...
```

...and that would be it!

parthm: Ah. Perfect. Thanks for the detailed explanation. It very helpful. Will update the tests.

zverok: Thanks! If you are interested, you can also take a look at http://www.betterspecs.org/ -- the examples there are a bit abstract, but explained in great details.

parthm:
@zverok Please let me know if the code / test are OK to merge now. Thanks for your earlier inputs. The code and test certainly look much cleaner now.

Issue #356, opened 2017-06-15
Vector#macd should return macdhist in addition to current macd and signal


parthm:
Current implementation of Vector#macd returns macd and macdsignal. It would be good if it returns macdhist / histogram in addition to these values to be in line with standard practice, e.g. [TA-Lib wrapper for python](https://mrjbq7.github.io/ta-lib/func_groups/momentum_indicators.html) has the following signature ```macd, macdsignal, macdhist = MACD(close, fastperiod=12, slowperiod=26, signalperiod=9)```

This is a fairly easy enhancement as ```histogram = series - signal``` and we are already returning series and signal. Calculation details can [found here](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:moving_average_convergence_divergence_macd#calculation).

parthm:
Thanks for merging the PR. We can probably close this now.

PR #357, opened 2017-06-15
[#356] proposed patch to support histogram in Daru::Vector#macd

parthm:
```Vector#macd``` updated to return ```[macdseries, macdsignal, macdhist]```. Added tests. Docs updated for the change and to be in line with yard format. Suggestions / comments welcome.

### CODE REVIEW

File: spec/maths/statistics/vector_spec.rb [hunk 1]
```diff
@@ -699,17 +699,26 @@
...
+      expect(signal[-10]).to be_within(1e-5).of(sig_10)
+      expect(signal[-20]).to be_within(1e-5).of(sig_20)
+
+      # check histogram
+      hist_1, hist_10, hist_20 = (macd_1-sig_1), (macd_10-sig_10), (macd_20-sig_20)
+      expect(hist[-1]).to be_within(1e-4).of(hist_1)
+      expect(hist[-10]).to be_within(1e-4).of(hist_10)
+      expect(hist[-20]).to be_within(1e-4).of(hist_20)
+
     end
```
zverok: Again, modern RSpec please!

parthm:
@zverok Please let me know if the tests look OK now or if something can be improved further.

spec/maths/statistics/vector_spec.rb
```diff
@@ -693,24 +693,46 @@
...
+      subject { source['price'].macd }
+
+      it do
+        30.times do |i|
+          idx = stability_offset + i
+          expect(macd[idx]).to be_within(delta).of(source['macd_12_26_9'][idx])
+          expect(macdsig[idx]).to be_within(delta).of(source['macdsig_12_26_9'][idx])
+          expect(macdhist[idx]).to be_within(delta).of(source['macdhist_12_26_9'][idx])
+        end
+      end
```
zverok: I'd say that testing something in cycle is a really bad tone (because if, for ex, if row 2 fails, it would not be "different in second element", but "expected X, received Y").
What I'd like to see here is something in line with:
```ruby
it { expect { macd }.to be_all_withing(delta).of(source['macd_12_26_9']) }
```
To do this, you'll define your own matcher, it is not hard.

Second thing is those 3 lines (716-718) are not DRY, neither between them, neither comparing with 729-731. I'd say something like this:
```ruby
context 'by default'
 it_should_behave_like 'correct macd', '12_26_9'
end

context 'custom values' do
 it_should_behave_like 'correct macd', '12_26_9', 6, 13, 4
end

# ...and in that shared examples something like:
%w[macd macdsig macdhist].each_with_index do |field, i|
  it { expect(subject[i]).to be_all_withing(delta).of(source["#{field}_#{code}"]) }
end

**Comment by parthm** [comment]
Thanks for the inputs @zverok . Custom matcher :be_all_within was a great idea. It can probably be used in other tests as well for comparing float arrays. Please let me know if the macd tests look good now.

File: spec/maths/statistics/vector_spec.rb [hunk 3]
```diff
@@ -693,24 +693,43 @@
...
-      expect(macd[-10]).to be_within(1e-4).of(-1.07e-2)
-      expect(macd[-20]).to be_within(1e-5).of(-5.65e-3)
+    if settings.empty?
+      subject { source['price'].macd }
+      settings_str = '12_26_9'
+    else
+      fast, slow, signal = settings
+      subject { source['price'].macd fast, slow, signal }
+      settings_str = settings.join('_')
+    end
```
zverok: I'd say:
```ruby
RSpec.shared_examples 'correct macd' do |*arguments|
  subject { source['price'].macd(*arguments) } # should work good for both situations
  let(:descriptor) { arugments.empty? ? '12_26_9' : arguments.join('_') }
end

File: spec/maths/statistics/vector_spec.rb [hunk 4]
```diff
@@ -693,24 +693,43 @@
...
+
+    %w[ macd macdsig macdhist ].each_with_index do |field, i|
+      macd_type = "#{field}_#{settings_str}"
```
zverok: No need in this local variable, it is ugly and used only once in code below, so could easily be inlined.

parthm:
Thanks for the inputs @zverok . Does this look better.

PR #364, opened 2017-07-09
Fix missing axis labels for categorized scatter with gruff

xprazak2:
The block was not being passed down to `plot_with_category` method, therefore modifications to plot ( such as adding axis labels) were not showing for scatter with category when using gruff.


xprazak2:
I added a test covering the change, sorry it took me so long..

PR #366, opened 2017-07-13
html table code of DataFrame and Vector are modified 

Shekharrajak:
It will be better if we use `thead` and `tbody` tag in the html table code of  the DataFrame and Vector . Also I have defined `to_html_thead` and `to_html_tbody` method to get the header and body part of the html table. These things will be helpful for `daru-view` table ( generate table using [DataTables](https://github.com/Shekharrajak/datatables.rb) gem ). 

[See how I am using this PR branch in datatables example to generate table code with pagination, search and other features. ](https://github.com/Shekharrajak/datatables.rb/blob/master/spec/dummy_rails/app/controllers/application_controller.rb#L24) . Actually DataTables gem will check the thead and tbody of the table.


v0dro:
@athityakumar use this in daru-io.

Shekharrajak:
Few things that can be useful in future (TODO): 

- `df.to_html(options)` : where `options` may contains table id, class or other styling to include in table tag. (this can be done using `content_tag`)

- In each row , first column must have `th` tag (first column is always `index`. Right now it is in `tr` tag. )  Just need to modify the tbody erb file.

Issue #331, opened 2017-04-13
Vector method 'at' can't work when index is DateTimeIndex


user919lx:
For examle, this code can't work
```
require  'daru'
dti = Daru::DateTimeIndex.new(6.times.map{|i| DateTime.new(2012,1,i+1)})
a=6.times.map{|i| i}

v=Daru::Vector.new(a,index:dti)

puts v.at(*[1,2])
```
system tells  `'key': undefined method '[]' for nil:NilClass (NoMethodError)`

I have found the reason. DateTimeIndex's constructor doesn't call super.

I add follow code then everything is OK.
```
class Daru::DateTimeIndex
  def initialize data, opts={freq: nil}

    super data

    Helper.possibly_convert_to_date_time data

    @offset =
      case opts[:freq]
      when :infer then Helper.infer_offset(data)
      when  nil   then nil
      else  Helper.offset_from_frequency(opts[:freq])
      end

    @frequency = @offset ? @offset.freq_string : nil
    @data      = data.each_with_index.to_a.sort_by(&:first)

    @periods = data.size
  end
end
```

baarkerlounger:
This is fixed and should be closed.

PR #368, opened 2017-07-14
Fix head/tail calls on DataFrames with DateTime indexes and Vector_at splat calls (#341 & #331)

db579:
Fixes issues https://github.com/SciRuby/daru/issues/341 and https://github.com/SciRuby/daru/issues/331 with reference to https://github.com/SciRuby/daru/pull/347 this is just adding the failing spec.

### DIFF
--- a/lib/daru/date_time/index.rb
+++ b/lib/daru/date_time/index.rb
@@ -226,7 +226,7 @@ module Daru
       to_a.each(&block)
     end
 
-    attr_reader :frequency, :offset, :periods
+    attr_reader :frequency, :offset, :periods, :keys
 
     # Create a DateTimeIndex with or without a frequency in data. The constructor
     # should be used for creating DateTimeIndex by directly passing in DateTime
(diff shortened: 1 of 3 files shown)


lib/daru/date_time/index.rb
```diff
@@ -226,7 +226,7 @@ def each(&block)
       to_a.each(&block)
     end
 
-    attr_reader :frequency, :offset, :periods
+    attr_reader :frequency, :offset, :periods, :keys
```
zverok: Why this change?.. For what I can tell, nothing fails without this additional reader.

db579: Without this change you can't call `vector.at(*[1,2])` (or as far as I can tell, any `.at`) on a vector with a DateTime index (bug #331). This is because DateTime index inherits from index which calls `keys (position)` which has `@keys[value]`.

Without this change the DateTime indexed vector has no @keys.

db579:
@zverok @v0dro do you usually close issues once PRs have been merged into master or once a new release of the gem has been made? (and how often are releases made?) Sorry if I've missed this information elsewhere.

zverok:
@v0dro I believe it is good question, and I don't know actual answer, please clarify!

v0dro:
Issues are closed after PRs have been merged. Currently we have a schedule for releases, however we've had a new release every 4-5 months in the past.

Issue #153, opened 2016-06-03
Vector initialization failed when dtype is nmatrix


lokeshh:
Vector containing `nil`s can't be initialized using `:nmatrix` even though nmatrix supports `nil`'s.

``` ruby
[12] pry(main)> x = Daru::Vector.new [2, nil], dtype: :nmatrix
TypeError: no implicit conversion from nil to integer
from /home/ubuntu/workspace/daru/lib/daru/accessors/nmatrix_wrapper.rb:28:in `initialize'
```

(Discovered this while working with https://github.com/v0dro/daru/pull/134; writing it down in case I forget it.)

v0dro:
You'll need to specify `:nm_dtype` as `:object` or alternatively catch the error and change the required data type to `:object` from the default `:int32`.

PR #369, opened 2017-07-14
Fix Vector initialization when dtype is nmatrix, vector has nils and no nm_type is given (#153)

db579:
Fixes https://github.com/SciRuby/daru/issues/153

Cyclomatic Complexity rubocop disabled for the `cast_vector_to` method as it seems necessary with that many possible case statement branches and the method is actually all that complex.

### DIFF
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -1481,7 +1481,7 @@ module Daru
 
     # Note: To maintain sanity, this _MUST_ be the _ONLY_ place in daru where the
     # @dtype variable is set and the underlying data type of vector changed.
-    def cast_vector_to dtype, source=nil, nm_dtype=nil
+    def cast_vector_to dtype, source=nil, nm_dtype=nil # rubocop:disable Style/CyclomaticComplexity
       source = @data.to_a if source.nil?
 
       new_vector =
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -1481,22 +1481,19 @@ module Daru
 
     # Note: To maintain sanity, this _MUST_ be the _ONLY_ place in daru where the
     # @dtype variable is set and the underlying data type of vector changed.
-    def cast_vector_to dtype, source=nil, nm_dtype=nil # rubocop:disable Style/CyclomaticComplexity
+    def cast_vector_to dtype, source=nil, nm_dtype=nil
       source = @data.to_a if source.nil?
 
       new_vector =
         case dtype
         when :array   then Daru::Accessors::ArrayWrapper.new(source, self)
-        when :nmatrix then
-          # To avoid arrays with nils throwing TypeError for :int32 default nm_dtype
-          nm_dtype = :object if nm_dtype.nil? && source.any?(&:nil?)
-          Daru::Accessors::NMatrixWrapper.new(source, self, nm_dtype)
+        when :nmatrix then Daru::Accessors::NMatrixWrapper.new(source, self, nm_dtype)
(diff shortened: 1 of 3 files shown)


lib/daru/vector.rb
```diff
@@ -1481,13 +1481,16 @@ def prepare_bootstrap(estimators)
...
+    def cast_vector_to dtype, source=nil, nm_dtype=nil # rubocop:disable Style/CyclomaticComplexity
       source = @data.to_a if source.nil?
 
       new_vector =
         case dtype
         when :array   then Daru::Accessors::ArrayWrapper.new(source, self)
-        when :nmatrix then Daru::Accessors::NMatrixWrapper.new(source, self, nm_dtype)
+        when :nmatrix then
+          # To avoid arrays with nils throwing TypeError for :int32 default nm_dtype
+          nm_dtype = :object if nm_dtype.nil? && source.any?(&:nil?)
```
zverok: Don't you think this check can be performed in `NMatrixWrapper#initialize`? This way current method would be much cleaner.

db579: The downside is if I do it there I'd either have to remove the `nm_dtype :int32` default in the method signature (which would mean doing the `any?(&:nil?)` check every time the method is called without an explicit `nm_dtype` (not good). Or I would have to check for `nm_dtype == :int32` which might produce unexpected results for someone that's deliberately passed that and might reasonably expect the method to fail.

zverok: No, you are wrong. If `nm_dtype` is `nil`, then default (`:int32`) option would NOT be used, it is used only if you don't pass `nm_dtype` at all. I've just tried on your branch to move this check to `NMatrixWrapper#initialize` (without any other changes), and everything works.

db579: @zverok yes that's what I mean. If you don't pass `nm_dtype` at all then it will use `:int32`  so we would check for nil values in the vector. But if you explicitly pass `:int32` as your `nm_dtype` and you have nils in the vector, you'd expect it to fail right? Because checking for nils in that case would be unnecessarily inefficient. That's the scenario I'm trying to avoid.

zverok: Still not getting it, sorry :( I am slightly slow today. Could you show it in code?..

db579: Sorry I'm probably not explaining it well.

I'm thinking the initialiser would have to look something like this:

```
def initialize vector, context, nm_dtype=:int32
          # To avoid arrays with nils throwing TypeError for :int32 default nm_dtype
          if nm_dtype == :int32 && vector.any?(&:nil?)
            @nm_dtype = :object
          else
            @nm_dtype = @data.dtype
          end
          @size = vector.size
          @data = NMatrix.new [@size*2], vector.to_a, dtype: nm_dtype
          @context = context
          # init with twice the storage for reducing the need to resize
        end
```

Which means you're doing the `vector.any?(&:nil?)` part even for the default case where no `nm_dtype` is passed, which seems wrong to me.

What do you think?

zverok: > `if nm_dtype == :int32`

Why? If `nm_dtype` haven't been passed to `cast_vector_to`, we'll have `nil` here, not `:int32`

db579: Maybe I'm missing something but I don't think that's true. As long as the method signature is `def initialize vector, context, nm_dtype=:int32` `:int32` will be the default and not `nil`?

zverok: Just try it.

db579: I guess you mean because we only ever use it from the `cast_vector_to` method? In any case I have changed it. Please review again :)

Issue #371, opened 2017-07-18
From CSV should support automatic boolean conversion


db579:
The Ruby CSV reader doesn't seem to have an automatic boolean converter when reading CSVs and so Daru also doesn't have this. Pandas does it by default and it seems like it would be a handy option to have rather than writing your own block each time.

https://github.com/SciRuby/daru/issues/337

Issue #337, opened 2017-04-26
CSV reading performance issue. 


sivagollapalli:
I just tried to read a csv file of size 42.6 MB which consists of 550241 records. When I tried with daru here is following statistics

```
# analysis.rb
require 'daru'
df = Daru::DataFrame.from_csv('./IndiaAffectedWaterQualityAreas.csv')
p df.first(2)

$ time ruby analysis.rb

real	0m48.878s
user	0m47.898s
sys	0m0.798s
```

if I do the same with python then it shows as follows

```
# analysis.py
import numpy as np # linear algebra
import pandas as pd
data = pd.read_csv('./IndiaAffectedWaterQualityAreas.csv', encoding = "ISO-8859-1")
print(data.tail(25))

$ time python analysis.py

real	0m1.649s
user	0m1.439s
sys	0m0.195s
```

Since Github doesn't allow a file to upload more than 10 MB so I couldn't upload. But you can get the data from https://www.kaggle.com/venkatramakrishnan/india-water-quality-data

zverok:
Unfortunately, that is for now "how it is". Nothing is broken in library itself that easily can be optimized. When profiling this case, most of the time spent into Ruby's CSV parser, in things like converting strings to numbers and so on. There are some plans/efforts to replace Ruby's default CSV library with something faster, but none of them are close to implementation.

v0dro:
I think we should shift to a C based CSV parser like paratext. See https://github.com/SciRuby/daru/issues/170

info-rchitect:
What about [smarter_csv](https://rubygems.org/gems/smarter_csv/versions/1.1.0)?

- able to process large CSV-files
- able to chunk the input from the CSV file to avoid loading the whole CSV file into memory
- return a Hash for each line of the CSV file, so we can quickly use the results for either creating MongoDB or ActiveRecord entries, or further processing with Resque
- able to pass a block to the process method, so data from the CSV file can be directly processed (e.g. Resque.enqueue )
- allows to have a bit more flexible input format, where comments are possible, and col_sep,row_sep can be set to any character sequence, including control characters.
- able to re-map CSV "column names" to Hash-keys of your choice (normalization)
- able to ignore "columns" in the input (delete columns)
- able to eliminate nil or empty fields from the result hashes (default)

info-rchitect:
Here is a ~10x speed-up improvement workaround over the default :from_csv method.  The CSV file in this case is ~65k rows and 40 columns.

[...]

PR #373, opened 2017-07-20
Add a boolean converter to the CSV IO (#371)

db579:
Resolves https://github.com/SciRuby/daru/issues/371


lib/daru/io/io.rb
```diff
@@ -275,6 +275,17 @@ def html_table_to_dataframe(table)
           order: table[:order],
           name: table[:name]
       end
+
+      CSV::Converters[:boolean_converter] = lambda do |f, _|
```
zverok: Well, despite the fact that docs explicitly say "This Hash is intentionally left unfrozen and users should feel free to add values to it that can be accessed by all CSV objects", I don't believe any "global" solution is good or even necessary. 

Imagine large project that uses daru AND some other code to work with CSV. What's the probability they also have defined `:boolean_converter` with slightly different behavior (for ex., converting `'yes'` and `'no'` too)?..

Let's just have a constant with lambda in it?

db579: @zverok went by the conversation at the bottom of this for why I added it in Daru https://github.com/SciRuby/daru/issues/337

db579: Obviously people can still define their own custom boolean converters but I think it's good to have some sort of general convenience method built in. (For example Pandas converts booleans by default and I was surprised there was no built-in method here as well).

zverok: Daru should convert booleans, yes. All I am agains is to modify built-in hash of `CSV::Converters`. I think it is a bad practice, generally.

db579: In my opinion the convenience of modifying the hash beats the better practice. Can change it though if that's the consensus.

db579: @v0dro @gnilrets thoughts?

gnilrets: The converters option of the CSV parser accepts either the name (symbol) of a converter in the `CSV::Converters` hash OR a lamba.  So, you could probably build a `Daru::CSV::Converters` hash for Daru-specific converters, and then have `DataFrame.from_csv` accept converter names that first look in `CSV::Converters` and then in `Daru::CSV::Converters`.  That way a user could override the Daru ones by either adding to `CSV::Converters` or supplying their own lambda.

db579:
@zverok Have tried to implement the suggestions. Please have another look.

db579:
@athityakumar may be of interest

Issue #351, opened 2017-06-05
Convenient way of setting new/existing column: df[:colx] = 1.0


parthm:
It may be worthwhile having ```df[:c] = 1.0``` as a convenient method of setting a new column and existing columns. Right now it errors out.

Daru: 0.1.5
Ruby: 2.4.0

```
[4] pry(main)> df = Daru::DataFrame.new({a: [1, 2, 3], b: [10, 20, 30]})
=> #<Daru::DataFrame(3x2)>
       a   b
   0   1  10
   1   2  20
   2   3  30
[5] pry(main)> df[:c] = 1.0
NoMethodError: undefined method `size' for 1.0:Float
from /home/user/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/daru-0.1.5/lib/daru/dataframe.rb:2188:in `prepare_vector_for_insert'
[6] pry(main)> df[:c] = [1.0] * df.nrows
=> [1.0, 1.0, 1.0]
[7] pry(main)> df
=> #<Daru::DataFrame(3x3)>
       a   b   c
   0   1  10 1.0
   1   2  20 1.0
   2   3  30 1.0
```

While we can do ```df[:c] = [1.0] * df.nrows``` it's probably more intuitive and clearer to do ```df[:c] = 1.0```, especially for new users from other libs like pandas. 

Behavior for pandas.

```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({'a': [1, 2, 3], 'b' : [10, 20, 30]})

In [3]: df
Out[3]: 
   a   b
0  1  10
1  2  20
2  3  30

In [4]: df['c'] = 1.0

In [5]: df
Out[5]: 
   a   b    c
0  1  10  1.0
1  2  20  1.0
2  3  30  1.0

In [6]: 
```

zverok:
I love the idea. Accepted :)

baarkerlounger:
Think this can be closed given the above merged commit.

PR #377, opened 2017-08-04
Allow inserting or updating DataFrame vectors with single values #351

baarkerlounger:
https://github.com/SciRuby/daru/issues/351

Enable:

```
df = Daru::DataFrame.new({a: [1, 2, 3], b: [10, 20, 30]})
=> #<Daru::DataFrame(3x2)>
       a   b
   0   1  10
   1   2  20
   2   3  30

df[:c] = 1.0
=> #<Daru::DataFrame(3x3)>
       a   b   c
   0   1  10  1.0 
   1   2  20  1.0
   2   3  30  1.0

df[:a] = 1.0
=> #<Daru::DataFrame(3x3)>
       a   b   c
   0   1.0  10  1.0 
   1   1.0  20  1.0
   2   1.0  30  1.0
```

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -2319,14 +2319,13 @@ module Daru
             v[idx] = vector.index.include?(idx) ? vector[idx] : nil
           end
         }
-      else
-        # FIXME: No spec checks this case... And SizeError is not a thing - zverok, 2016-05-08
+      elsif vector.is_a?(Array) || vector.is_a?(Range)
         if @size != vector.size
-          raise SizeError,
-            "Specified vector of length #{vector.size} cannot be inserted in DataFrame of size #{@size}"
+          raise "Specified vector of length #{vector.size} cannot be inserted in DataFrame of size #{@size}"
         end
-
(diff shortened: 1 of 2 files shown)


zverok:
Ugh. Sorry for being late to the party, I've started reviewing yesterday but got distracted :( This merged code has one important problem. What if value assigned _is_ Enumerable, but _is not_ `Array` or `Range`? (Say, it is `ActiveRecord::Relation`.) What user expects is it would be converted `to_a`, and then inserted as a vector. What actually will happen, is pretty obscure error:
* `when Array, Range` would be skipped;
* in `prepare_value_for_insert`, there would be called `Array(relation)`, which will do `to_a` (producing array of objects)... and then multiplied by DF size, producing vector much larger than expected.

@baarkerlounger can you please do a small follow-up PR which fixes this (I believe, by replacing `when Array, Rang` with `if respond_to?(:to_a)`) It is OK if you have no time for this, I can do it myself.

baarkerlounger:
@zverok implemented your suggestion here: https://github.com/SciRuby/daru/pull/379

PR #385, opened 2017-08-12
bugfix for Vector#proportions, make it return float value

rainchen:
bugfix for Vector#proportions, make it return float value

### DIFF
--- a/lib/daru/maths/statistics/vector.rb
+++ b/lib/daru/maths/statistics/vector.rb
@@ -205,7 +205,7 @@ module Daru
         def proportions
           len = size - count_values(*Daru::MISSING_VALUES)
           frequencies.to_h.each_with_object({}) do |(el, count), hash|
-            hash[el] = count / len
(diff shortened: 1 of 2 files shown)


v0dro:
Tests aren't passing because you're using èq` instead of `be_within`. Just make that change and I'll merge.

rainchen:
@v0dro fixed and tests have passed.

But I'm aware that using :gsl will return convert result into a float value even the input value is integer, e.g.: 

```
@dv = Daru::Vector.new [323, 11, 555, 666, 234, 21, 666, 343, 1, 2], dtype: dtype
```

rainchen:
@v0dro all tests have passed

v0dro:
Merging your PR but can you please open an issue pointing out the behavior displayed by GSL?

rainchen:
@v0dro I'm not using GSL, not sure it's issue or not, although that behavior is unexpected for me.

Issue #336, opened 2017-04-24
Drop support for Ruby 2.0


zverok:
I believe we can allow ourselves to it now :)

This not only means specifying 2.1 as a minimum version in the Gemfile, but also switching from option hashes to proper keyword arguments everywhere, which will greatly enchance documentability and discoverability of features.

On this change, version should be bumped at least to 0.2, I believe.

PS: As a side note, I believe 0.x becames to be a bit odd for library of this size. Maybe we need something like "Roadmap to 1.0" somewhere?..

v0dro:
Agreed. And yes, we should have a roadmap. Maybe we take inspiration from JRuby for this purpose? I've heard quite a few people talk fondly about their release schedules and roadmap.

zverok:
Let's probably schedule the roadmap creation for August/September? When GSoC will be finished and we'll know where we are.

v0dro:
Yes. Works.

PR #387, opened 2017-08-15
Fixes min, max, index_of_max and index_of_min for Daru::Vector

athityakumar:


lib/daru/maths/statistics/vector.rb
```diff
@@ -855,6 +835,14 @@ def raw_sample_without_replacement sample
 
           out.collect { |i| valid[i] }
         end
+
+        def sort_without_nil(&block)
+          dv = reject_values(nil)
```
zverok: TBH, it looks for me like `data = @data.data.to_a.compact` and then sorting it could be much more efficient.

lib/daru/maths/statistics/vector.rb
```diff
@@ -855,6 +835,14 @@ def raw_sample_without_replacement sample
...
+
+          return dv.sort.to_a unless block_given?
+          return dv.sort_by(&block) if block.parameters.count == 1 # Object block like { |x| x.size }
+          dv.sort(&block).to_a # Comparative block like { |a,b| a.size <=> b.size }
```
zverok: I don't like this kind of smartiness. Ruby's collections have `max` and `max_by` for two different notations, and I believe it should either be kept this way, or only acceptin 1-argument block will be completely enough.

lib/daru/maths/statistics/vector.rb
athityakumar: But reflecting the effect of `@data.compact.to_a` correspondingly into the `@index.to_a` is something which could make `index_of_min` and `index_of_max` more ineffecient / unidiomatic. Rather, `reject_values(nil)` seemed like a better alternative. Let me know if it should be reverted back to `@data.data.to_a.compact` approach.

zverok: OK, let's leave it this way.

lib/daru/maths/statistics/vector.rb
athityakumar: Acknowledged. Changed `Daru::Vector#max` and `Daru::Vector#max_by` and other similar methods, to resemble the usual enum `#feature` and `#feature_by` behaviour.

lokeshh:
@athityakumar Why are you sorting to find the min/max? Sorting takes `O(n log n)` while min/max can be done in `O(n)`.

athityakumar:
@lokeshh - `Array#max_by` takes argument of `n` for number of max_by elements to return, from [Ruby 2.2.0 onwards](https://ruby-doc.org/core-2.2.0/Enumerable.html#method-i-max_by) and is NOT present in [Ruby 2.1.0](https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-max_by). Hence, sorting is a workaround to have `Daru::Vector#max_by(n)` work even for Ruby 2.1.0 environments.

I should probably add a note regarding this, so that these methods don't get forgotten to be changed while deprecating support for Ruby 2.1.0.

v0dro:
Can you add a statement that will run a particular algorithm if RUBY_VERSION is greater than 2.1.0? That way there's no need of the note, we'll find out and remove it anyway.

parthm:
Should we be using [`Gem::Version`](http://ruby-doc.org/stdlib-2.0.0/libdoc/rubygems/rdoc/Gem/Version.html) for comparison? `Gem::Version.new('0.4.1') > Gem::Version.new('0.10.1')`.

athityakumar:
@v0dro - Acknowledged, I've added the check for `RUBY_VERSION < 2.2.0`. However, Travis CI is now failing as coverage in `statistics/vector.rb` is less than 95% for Ruby 2.2+ versions.

v0dro:
Well then how about adding the coverage and fixing the issue? :P

v0dro:
Very weird failure on ruby 2.0 and 2.1.

athityakumar:
@v0dro @zverok - Ruby 2.2+ now have a coverage of 94.53% ( < 95%, duh). Also, I'm seeing some pretty strange failure on Ruby 2.0. I tried (hard) reverting to a previous commit whose Travis CI build was successful, and it turns out to fail now. And now I think it's because of backport's removal by PR #394.

```
2.0.0-p648 :001 > [[1,2],[3,4]].to_h #! Typical failures
NoMethodError: undefined method `to_h' for [[1, 2], [3, 4]]:Array
2.0.0-p648 :002 > require 'backports' #! This has now been removed from daru
 => true
2.0.0-p648 :003 > [[1,2],[3,4]].to_h #! No failure
 => {1=>2, 3=>4}
```

Is it time to drop support for Ruby 2.0 as per issue #336?

baarkerlounger:
@athityakumar it's the removal of backports completely in @zverok's PR. I fixed it here https://github.com/SciRuby/daru/pull/398 by just requiring the specific backport needed.

zverok:
Yes, it was a misunderstanding during solving backports problem. 
I believe we'll solve it just by merging #398, without separate PR :)

zverok:
Oh, to hell with that, quickfixed the CI problem myself :)
@athityakumar please rebase from master, everything should be OK.

athityakumar:
@v0dro @zverok - Finally managed to get the coverage issue solved. 😄

lib/daru/maths/statistics/vector.rb
```diff
@@ -82,24 +82,38 @@ def factors
         #
         #   dv.max(2) { |a,b| a.size <=> b.size }
         #   #=> ["Jon Starkgaryen","Daenerys"]
+        def max(size=nil, &block)
+          data           = reject_values(*Daru::MISSING_VALUES).to_a
+          range          = size.nil? ? 0 : (0..size-1)
+          default_method = RUBY_VERSION >= '2.2'
+          default_method ? data.max(size, &block) : data.sort(&block).reverse[range]
+        end
```
zverok: I have a LOT of questions to this solution :) I am generally unpleasant reviewer, you know.

First. Different implementations for a different Ruby versions is better to implement this way:
```ruby
if RUBY_VERSION >= '2.2'
  def implementation_method
  end
else
  def implementation_method
  end
end
```
Explanation: In your current version, `if` is performed every time `max` is called; in proposed one, it is performed only once, on files loading. And you'll see two different implementations would be cleaner, as only one of them needs `range`

lib/daru/maths/statistics/vector.rb
```diff
@@ -82,24 +82,38 @@ def factors
...
-                 else
-                   data.sort
-                 end
-          size.nil? ? data.last : data[data.count-size..-1].reverse
+        def max_by(size=nil, &block)
+          data           = reject_values(*Daru::MISSING_VALUES).to_a
+          range          = size.nil? ? 0 : (0..size-1)
+          default_method = RUBY_VERSION >= '2.2'
+          default_method ? data.max_by(size, &block || ->(i) { i }) : data.sort_by(&block || ->(i) { i }).reverse[range]
         end
```
zverok: Second: `->(i) { i }` is `&:itself` → which means entire construct could be shortened to `&(block || :itself)` (requires backports for Ruby < 2.2).

But, third: whether it is a good thing to use "sort by itself" by default? It is a) pretty different from `Array#max_by` behavior and b) useful for nothing (e.g. how `vector.max_by` block-less is better/more readable than `vector.max`?). I believe we should just raise on absent block.

lib/daru/maths/statistics/vector.rb
```diff
@@ -82,24 +82,38 @@ def factors
...
 
-        # Returns the index of the maximum value present in the vector.
+        # Returns the index of the maximum value present in the vector, with an optional
+        # comparator block.
```
zverok: Fourth: here and above, "..., with the optional comparator block" is not clear at all. (It reads like "returns the index... with an optional comparator", like "what???") I believe it should be something like: "Returns the index of the maximum value present in the vector. If the block is given, it is uses to decide max element, otherwise default comaprison is used." or something like this.

lib/daru/maths/statistics/vector.rb
```diff
@@ -115,17 +129,36 @@ def max(size=nil, &block)
...
-        #   dv.max(2) { |i| i.size }
+        #   dv = Daru::Vector.new (["Tyrion", "Daenerys", "Jon Starkgaryen"]), index: Daru::Index.new([:t, :d, :j])
+        #   #=>
+        #   #   #<Daru::Vector(3)>
+        #   #       t   Tyrion
+        #   #       d   Daenerys
+        #   #       j   Jon Starkgaryen
+        #
+        #   dv.index_of_max_by
+        #   #=> :t
```
zverok: Same as above: why block-less `index_of_max_by` is a thing? It is not readable.

lib/daru/maths/statistics/vector.rb
v0dro: Good one.

v0dro:
@athityakumar ping!

athityakumar:
@v0dro @zverok - Added the changes suggested above, and got caught again in the vicious cycle of coverage < 95%. 😢

zverok:
OK, I am officially done with that "coverage" limitation, I'll remove it in master shortly. We need another way for enforcing good coverage.

PR #389, opened 2017-08-17
allow to call count method without parameter for category type vector

rainchen:


lib/daru/category.rb
```diff
@@ -131,11 +131,17 @@ def add_category(*new_categories)
...
+    def count category = nil
+      if category.nil? # count all
+        @cat_hash.values.map(&:size).sum
+      else
+        raise ArgumentError, "Invalid category #{category}" unless
+          categories.include?(category)
 
-      @cat_hash[category].size
+        @cat_hash[category].size
+      end
```
zverok: Three notes:

1. Let's rely on early return idiom:
```ruby
return ... if category.nil?
rais ... unless
@cat_hash[category].size
```
2. Let's not rely on `sum` method, it is not in stdlib before 2.4. It accidentally works (because of active_support required in specs, and ruby-ole required in lib), but it is pure accident.
3. Maybe some dedicated "undefined" default value would be more reliable? Can't we have categorical vector with `nil`s?

rainchen:
@zverok updated. 

About the nil value issue, I'm not quite sure how to deal with. 
I wish to keep using the same style with `count` method after the vector converted to category type, e.g.: 

```
vc.count  # => 5
vc2 = vc.to_category
vc2.count  # => 5, excepted
```

how about using a special `:nil` key for counting nil values? .e.g: `vc.count(:nil)`

zverok:
> About the nil value issue, I'm not quite sure how to deal with.

There is "undefined object" pattern for such cases:
```ruby
UNDEFINED = Object.new.freeze # anonymous object, equal only to itself
def count category=UNDEFINED
  return .... if category == UNDEFINED
```

With this idiom, we know for sure that `UNDEFINED` would not be one of the categories in the vector, so, everything works always as expected, even for `nil` category.

rainchen:
@zverok good idea

rainchen:
@zverok All checks have passed, plz review

Issue #393, opened 2017-08-24
Matrix.column_count returns nil after require 'daru'


tramuntanal:
The bug appears in two forms.
1. If the Matrix was created before requiring 'daru' `colum_size` doesn't work:
```
irb(main):002:0> require 'matrix'
=> true
irb(main):003:0> m= Matrix.zero(2)
=> Matrix[[0, 0], [0, 0]]
irb(main):004:0> m.column_count
=> 2
irb(main):005:0> m.column_size
=> 2
irb(main):006:0> require 'daru'
=> true
irb(main):007:0> m.column_count
=> 2
irb(main):008:0> m.column_size
=> nil
```

2. If the Matrix is created after requiring 'daru' `colum_count` doesn't work:
```
irb(main):001:0> require 'matrix'
=> true
irb(main):002:0> m= Matrix.zero(2)
=> Matrix[[0, 0], [0, 0]]
irb(main):003:0> m.column_count
=> 2
irb(main):004:0> m.column_size
=> 2
irb(main):005:0> require 'daru'
=> true
irb(main):006:0> m= Matrix.zero(2)
=> Matrix[[0, 0], [0, 0]]
irb(main):007:0> m.column_count
=> nil
irb(main):008:0> m.column_size
=> 2
```

I can't find any reference to `#column_count` or `#column_size` in the Daru code.

zverok:
Aknowledged. It is nasty bag, caused by [backports](https://github.com/marcandre/backports) gem requireing. Unfortunately, it is not that easy to fix: backports required not only in daru itself, but also in `packable` gem, required by `nmatrix`. I've did a [PR](https://github.com/marcandre/packable/pull/9) to packable, but I am not sure when it could be merged :( Only after that we can proceed with fixing the bug.

v0dro:
Issue has been solved by removing backports dependency.

zverok:
Ugh. We can't mark this issue as a resolved. I should've mark corresponding PR as WIP, sorry :(
The problem describe in this comment: 

> It is nasty bag, caused by backports gem requireing. Unfortunately, it is not that easy to fix: backports required not only in daru itself, but also in packable gem, required by nmatrix. I've did a PR to packable, but I am not sure when it could be merged :( Only after that we can proceed with fixing the bug.

I provided a PR to original `packable` repo, and it was merged, but only after that I've noticed that SciRuby have taken over packable.

So, the course of actions are:
[...]

PR #394, opened 2017-08-25
Remove "backports"

zverok:
Looks like `backports` is not necessary now (was for Ruby 1.9 compatibility) AND causes some really weird bugs like https://github.com/SciRuby/daru/issues/393


v0dro:
Oh wait. Can you submit another PR that removes backports from the gemspec too?

Issue #392, opened 2017-08-22
Empty DataFrame creation


parthm:
Is there a way of creating an empty DataFrame? Not sure if my understanding of DataFrame is limited here. If not, it may be good to create an empty DataFrame with `DataFrame.new`. The use case for this is when you want to create an empty DataFrame and add columns from multiple data sources. Something like below

```ruby
df = DataFrame.new
sources.each do |col, csv|
  df[col] = Daru::DataFrame.from_csv(csv)['source_col']
end
```

Currently `DataFrame.new` fails with the below error:
```
irb(main):003:0> df = Daru::DataFrame.new
ArgumentError: wrong number of arguments (given 0, expected 1..2)
        from /home/parth/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daru-0.1.5/lib/daru/dataframe.rb:242:in `initialize'
```

At the moment I am creating a DataFrame with a dummy column and then adding new columns. The dummy columns is deleted after this. Not sure if there is a better way.

Python pandas equivalent will be:
```python
>>> import pandas as pd
>>> df = pd.DataFrame()
>>> df
Empty DataFrame
Columns: []
Index: []
>>>

```

zverok:
TBH, this example:
```ruby
df = DataFrame.new
sources.each do |col, csv|
  df[col] = Daru::DataFrame.from_csv(csv)['source_col']
end
```
...doesn't looks like a pretties Ruby code possible to me; and I am not sure that creation of row-less and column-less DataFrame is really semantically meaningful... Though, creation of empty dataframes should be definitely simplified.

parthm:
@zverok , yeah, probably the code can be better. I was merely using that as a use case of why we might need an empty `DataFrame`.

Looking into the source code, I realized that empty DataFrame is already supported. The approach is:

```ruby
irb(main):003:0> df = Daru::DataFrame.new({})
=> limited output
 #<Daru::DataFrame(0x0)>

irb(main):004:0> df[:foo] = [1, 2, 3]
=> limited output
 [1, 2, 3]
irb(main):005:0> df
=> limited output
 #<Daru::DataFrame(3x1)>
     foo
   0   1
   1   2
   2   3
```

It would be a nice enhancement to have `Daru::DataFrame.new` be equivalent to `Daru::DataFrame.new({})`. Seems more discoverable. Thoughts?

zverok:
Aknowledged.

PR #395, opened 2017-08-26
Issue 392 empty dataframe

parthm:
This patch allows `Daru::DataFrame.new` to create an empty dataframe as noted in issue #392. Comments / suggestions welcome.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -277,6 +277,17 @@ module Daru
     # Default to *true*.
     #
     # == Usage
+    #
+    #   df = Daru::DataFrame.new
+    #   # =>
+    #   # <Daru::DataFrame(0x0)>
+    #   # Creates an empty DataFrame with no rows or columns.
+    #
+    #   df = Daru::DataFrame.new({}, order: [:a, :b])
+    #   #<Daru::DataFrame(0x2)>
+    #     a   b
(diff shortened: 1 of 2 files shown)


rainchen:
+1 for this feature, default value is neat solution. 

But you should rebase your branch and there should be only 1 commit for this PR.

parthm:
Good catch @rainchen . Rebased.

zverok:
> there should be only 1 commit for this PR.

It is not necessary, in fact. :) We use Squash&Merge github feature.

rainchen:
@zverok "Don't make think too much", actually this is a code review rule in my team. With clear commits can save reviewer's time. Reviewers can quickly to get the summary of a PR like "oh, there is 1 commit and 2 file changes and it's expected if I do this task by myself".

v0dro:
You have a point. However, in this project where the only means of communication is a Pull Request and the issue tracker, having one PR per feature becomes a more generalized and easier way of communication. If there is one feature per commit, one needs to sit and read a list of commits, instead of quickly looking at the PR name and description. Its just more 'in your face' if you know what I mean.

Though it does not hurt to have one commit per feature in separate PRs :)

zverok:
@rainchen Yep, @v0dro is right. 
Being one of the guys who does reviewing and merging, and considering we have people of various backgrounds trying to contribute, I explicitly state I am are OK with several commits per PR, not a big deal.

PR #396, opened 2017-08-26
add which_dsl extension: use `which{ }` method as query DSL for accessing `where()`

rainchen:
add a simple query DSL for accessing where(), inspired by gem "squeel"
e.g.:

```
df.which{ `FamilySize` == `FamilySize`.max }`
```

equals

```
df.where( df['FamilySize'].eq( df['FamilySize'].max ) )
```

e.g.:

```
df.which{ (`NameTitle` == 'Dr') & (`Sex` == 'female') }
```

equals

```
df.where( df['NameTitle'].eq('Dr') & df['Sex'].eq('female') )
```

btw, repeating `df` variable is NOT DRY at all to me.

### DIFF
--- a/lib/daru/extensions/which_dsl.rb
+++ b/lib/daru/extensions/which_dsl.rb
@@ -1,0 +1,55 @@
+# Support for a simple query DSL for accessing where(), inspired by gem "squeel"
+
+module Daru
+  class DataFrame
+    # a simple query DSL for accessing where(), inspired by gem "squeel"
+    # e.g.:
+    # df.which{ `FamilySize` == `FamilySize`.max }
+    # equals
+    # df.where( df['FamilySize'].eq( df['FamilySize'].max ) )
+    #
+    # e.g.:
+    # df.which{ (`NameTitle` == 'Dr') & (`Sex` == 'female') }
+    # equals
+    # df.where( df['NameTitle'].eq('Dr') & df['Sex'].eq('female') )
+    def which(&block)
+      WhichQuery.new(self, &block).exec
+    end
+  end
+
+  class WhichQuery
+    def initialize(df, &condition)
+      @df = df
+      @condition = condition
+    end
+
+    # executes a block of DSL code
+    def exec
+      query = instance_eval(&@condition)
+      @df.where(query)
+    end
+
+    def `(vector_name)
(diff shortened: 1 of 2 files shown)


zverok:
Pretty smart! TBH, I am not totally sure about backtick trick... Maybe method_missing/const_missing would be more predictable?.. (Sequel does this for its block DSL) It has its own drawbacks also, of course.

> btw, repeating df variable is NOT DRY at all to me.

Yep, it is ugly as hell. It is an ancestry of copying Python's libraries, which I personally hate.

rainchen:
some column names(from csv) may containing space or special characters like `'` or '-' etc， for these cases backtick is better then method_missing, and it's really recognizable compared to a normal method.

zverok:
@rainchen Yep, makes sense for me.

I'm willing to happily accept this, but it needs some tweaks for docs and specs:
* for docs: we should be sure to hide `WhichQuery` from YARD (and probably add a bit more explanations to `which`... But not sure it is necessary)
* for specs: we are (slowly) trying to move to moder RSpec, as per betterspecs.org (with subjects, lets, contexts and stuff), for new features/modules they should be written this way.

Are you willing to make those changes? If not, I'll merge your PR, as it is a valuable contribution, and then update the style of specs/docs myself.

rainchen:
@zverok

I implemented this feature as an "extension", requiring to use `require 'daru/extensions/which_dsl'` to enable this feature (took example from rserve extension)

I'm glad you can work on the docs and code origination parts. 

Here are some actually use cases that I'm using this `which` method:

```
# getting oldest passengers
passengers.which { `Age` == `Age`.max }
```

```
# getting youngest passengers
passengers.which { `Age` == `Age`.min }
```

```
# getting female Dr
passengers.which { (`NameTitle` == 'Dr') & (`Sex` == 'female') }
```

```
# getting passengers had the same name
same_name = passengers.which { `Name`.frequencies > 1 } ["Name"].to_a
# => ["Connolly, Miss. Kate", "Kelly, Mr. James"]
passengers.which { `Name` =~ same_name }
```

hope these helpful for you or others be interested.

zverok:
Thank you! :hugs:

rainchen:
Thank you all!

Issue #391, opened 2017-08-22
DataFrame#vector_sum should ignore nils


parthm:
`Daru::DataFrame#vector_sum` makes the sum `nil` if the row contains a `nil`. It may be worth while ignoring the `nil` entries for sum.

Note that the last row below shows the sum as `nil` instead of `5`.

```
irb(main):001:0> df
=> limited output
 #<Daru::DataFrame(11x4)>
      v0  v1  v2  v3
   0  -5  11  21  31
   1  -4  12  22  32
   2  -3  13  23  33
   3  -2  14  24  34
   4  -1  15  25  35
   5   0  16  26  36
   6   1  17  27  37
   7   2  18  28  38
   8   3  19  29  39
   9   4  20  30  40
  10   5 nil nil nil
irb(main):002:0> df[:vx] = df.vector_sum
=> limited output
irb(main):003:0> df
=> limited output
 #<Daru::DataFrame(11x5)>
      v0  v1  v2  v3  vx
   0  -5  11  21  31  58
   1  -4  12  22  32  62
   2  -3  13  23  33  66
   3  -2  14  24  34  70
   4  -1  15  25  35  74
   5   0  16  26  36  78
   6   1  17  27  37  82
   7   2  18  28  38  86
   8   3  19  29  39  90
   9   4  20  30  40  94
  10   5 nil nil nil nil
irb(main):004:0>
```

`Vector#sum` ignores nil and presents the sum correctly. `DataFrame#vector_sum`  ignoring `nil` would make the behavior more consistent.
`Vector#sum` below ignores the 10th entry which is `nil`.
```
irb(main):010:0> v
=> limited output
 #<Daru::Vector(11)>
      v1
   0  11
   1  12
   2  13
   3  14
   4  15
   5  16
   6  17
   7  18
   8  19
   9  20
  10 nil
irb(main):011:0> v.sum
=> limited output
 155

```

parthm:
Looking into the tests, there seems to be a [specific test](https://github.com/SciRuby/daru/blob/4879ff3bd2c863c7a874e7064b87363a6347473c/spec/dataframe_spec.rb#L3194) to ensure that the sum is `nil` if the row contains any nils. So, I assume this is intended behavior at least for some use cases.

[...]

PR #398, opened 2017-08-27
Allow vector_sum to ignore nils #391

baarkerlounger:
https://github.com/SciRuby/daru/issues/391

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -661,6 +661,10 @@ module Daru
       self
     end
 
+    def replace_nils replacement
+      replace_values nil, replacement
+    end
+
     # Iterate over each index of the DataFrame.
     def each_index &block
       return to_enum(:each_index) unless block_given?
@@ -1272,10 +1276,10 @@ module Daru
 
     # Returns a vector with sum of all vectors specified in the argument.
     # If vecs parameter is empty, sum all numeric vector.
-    def vector_sum vecs=nil
+    def vector_sum vecs=nil, skipnil=false
       vecs ||= numeric_vectors
       sum = Daru::Vector.new [0]*@size, index: @index, name: @name, dtype: @dtype
-
+      self.replace_nils(0) if skipnil
       vecs.inject(sum) { |memo, n| memo + self[n] }
     end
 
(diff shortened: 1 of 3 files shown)


lib/daru/dataframe.rb
```diff
@@ -1272,10 +1276,10 @@ def tail quantity=10
 
     # Returns a vector with sum of all vectors specified in the argument.
     # If vecs parameter is empty, sum all numeric vector.
-    def vector_sum vecs=nil
+    def vector_sum vecs=nil, skipnil=false
       vecs ||= numeric_vectors
       sum = Daru::Vector.new [0]*@size, index: @index, name: @name, dtype: @dtype
-
+      replace_nils(0) if skipnil
```
zverok: Unfortunately, there is a problem with this approach. `replace_nils(0)` (as well as `replace_values`) changes the source DF. The result would be the following:
```ruby
df = Daru::DataFrame.new({
        a: [1,    2,          3,   nil,        Float::NAN, nil, 1,   7],
        b: [:a,  :b,          nil, Float::NAN, nil,        3,   5,   8],
        c: ['a',  Float::NAN, 3,   4,          3,          5,   nil, 7]
    })
# => #<Daru::DataFrame(8x3)>
#       a   b   c
#   0   1   a   a
#   1   2   b NaN
#   2   3 nil   3
#   3 nil NaN   4
#   4 NaN nil   3
#   5 nil   3   5
#   6   1   5 nil
#   7   7   8   7 

df.vector_sum(nil, true)
# => #<Daru::Vector(8)>
#   0   1
#   1   2
#   2   3
#   3   0
#   4 NaN
#   5   0
#   6   1
#   7   7 

df
# => #<Daru::DataFrame(8x3)>
#       a   b   c
#   0   1   a   a
#   1   2   b NaN
#   2   3   0   3
#   3   0 NaN   4
#   4 NaN   0   3
#   5   0   3   5
#   6   1   5   0
#   7   7   8   7 
```

baarkerlounger: @zverok You're right, can fix by making it clone and sum or we can go with @parthm's approach. What do you think?

PR #400, opened 2017-08-28
Issue 391 vector sum skipnil

parthm:
Adds support for `skipnil` flag as discussed in issue #391 . There is some common code between `add_skipnil` and `v2v_binary` but I decided to keep these two separate to avoid performance penalty for all binary operation to support one scenario of `vector_sum` with `skipnil`. Suggestions welcome.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1272,11 +1272,16 @@ module Daru
 
     # Returns a vector with sum of all vectors specified in the argument.
     # If vecs parameter is empty, sum all numeric vector.
-    def vector_sum vecs=nil
+    def vector_sum vecs=nil, opts={skipnil: false}
       vecs ||= numeric_vectors
       sum = Daru::Vector.new [0]*@size, index: @index, name: @name, dtype: @dtype
-
-      vecs.inject(sum) { |memo, n| memo + self[n] }
+      vecs.inject(sum) do |memo, n|
+        if opts[:skipnil]
+          self[n].add_skipnils(memo)
+        else
+          memo + self[n]
+        end
+      end
     end
 
     # Calculate mean of the rows of the dataframe.
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1323,13 +1323,7 @@ module Daru
 
       vecs ||= numeric_vectors
       sum = Daru::Vector.new [0]*@size, index: @index, name: @name, dtype: @dtype
-      vecs.inject(sum) do |memo, n|
-        if skipnil
-          self[n].add_skipnil(memo)
-        else
-          memo + self[n]
-        end
-      end
+      vecs.inject(sum) { |memo, n| self[n].add(memo, skipnil: skipnil) }
     end
 
     # Calculate mean of the rows of the dataframe.
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1270,14 +1270,62 @@ module Daru
 
     alias :last :tail
 
-    # Returns a vector with sum of all vectors specified in the argument.
-    # If vecs parameter is empty, sum all numeric vector.
-    def vector_sum vecs=nil, opts={skipnil: false}
+    # Sum all numeric/specified vectors in the DataFrame.
+    #
+    # Returns a new vector that's a containing a sum of all numeric
+    # or specified vectors of the DataFrame. By default, if the vector
+    # contains a nil, the sum is nil.
+    # With :skipnil argument set to true, nil values are assumed to be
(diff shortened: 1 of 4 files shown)


parthm:
Ha ha. Didn't realize @baarkerlounger had created a PR for the same request. Perhaps @zverok and @v0dro can merge one. I am OK if the other PR #398 is merged.

In case this one is to be merged, please let me know, I would like to add documentation for `add_skipnil`. In the process of fixing this I realized that vector addition `v0 + v1` suffers from this issue. So having a public method `add_skipnil` can give users an option to add two vectors. `v0.add_skipnil v1`.

parthm:
Weird failure on Ruby 2.0 travis.

```ruby
     Failure/Error: @cat_hash = categories.map { |cat, group| [cat, group.map(&:last)] }.to_h
     NoMethodError:
       undefined method `to_h' for [[:a, [0, 2]], [1, [1, 3]], [:c, [4]]]:Array
```

Anyone knows what that is? Looks like PR #387 is also seeing the same error.

baarkerlounger:
@parthm it's the removal of backports completely in @zverok's PR. I fixed it here #398 by just requiring the specific backport needed.

lib/daru/dataframe.rb
```diff
@@ -1270,13 +1270,66 @@ def tail quantity=10
...
+    #       2 nil
+    #    df.vector_sum skipnil: true
+    #    => #<Daru::Vector(3)>
+    #           c
+    #       0   4
+    #       1   4
+    #       2   4
+    #
+    def vector_sum(*args)
+      defaults = {vecs: nil, skipnil: false}
```
baarkerlounger: What's the advantage of this over just having named args with defaults since there's only 2?

zverok: I believe the advantage is you can do `vector_sum(skip_nil: true)` instead of `vector_sum(nil, true)` of your PR, which is a bit inelegant. 

In addition, when we'll drop Ruby 2.0 support (which is immediately after Google Summer of Code end), here would be proper keyword args. So, I believe it is a good solution.

parthm: @baarkerlounger The intent was to provide an easy way skip the first argument while providing the second one. At the cost of some arg processing complexity, arguably the user interaction gets simplified a bit. The following combinations work.

```ruby
df.vector_sum
df.vector_sum skipnil: true
df.vector_sum [:a, :b], skipnil: true
df.vector_sum [:a, :b]
```

baarkerlounger:
Since this looks to be a better solution I'll close my PR.

lib/daru/maths/arithmetic/vector.rb
```diff
@@ -42,6 +42,30 @@ def round precision=0
...
+          index = (@index.to_a | other.index.to_a)
+          elements = index.map do |idx|
+            this = self.index.include?(idx) ? self[idx] : nil
+            that = other.index.include?(idx) ? other[idx] : nil
+            this = 0 if this.nil?
+            that = 0 if that.nil?
+            this && that ? this + that : nil
+          end
+          Daru::Vector.new(elements, name: @name, index: index)
+        end
```
zverok: I believe it is too verbose way. WDYT about this:
```ruby
# existing method, just update it a bit:
def v2v_binary operation, other, empty = nil
  index = (@index.to_a | other.index.to_a).sort

  elements = index.map do |idx|
    this = self.index.include?(idx) ? self[idx] : nil
    that = other.index.include?(idx) ? other[idx] : nil

    this && that ? this.send(operation, that) : empty
  end

  Daru::Vector.new(elements, name: @name, index: index)
end

# more poverful synonym for +
def add other, empty = nil
  binary_op :+, other, empty
end

# finally, in vector_sum:
...
vecs.inject(sum) { |memo, n| memo.add(self[n], skipnil ? 0 : nil) }
...

Comment (parthm) [reply lib/daru/maths/arithmetic/vector.rb]: @zverok an idea of clean `add` looks good. The suggested implementation will need some changes. In the above, `[1, 2, nil] add [1, 2,  3]` will produce `[1, 2, 0]`. To make this verbose I am thinking an approach similar to what you suggest but having a condition added to `v2v_binary`:

```ruby

            if (this.nil? || that.nil?) && skipnil
              this = 0 if this.nil?
              that = 0 if that.nil?
            end
```

Note the additional `(this.nil? || that.nil?) `. Reviewing the code I realized, that `nil add nil` for the vector should produce `nil`. Without this condition, it will produce `0`.

Will push a change along the lines suggested by you and the condition added.

parthm: I was mistaken about `nil add nil`. 0 makes sense as its consistent with `DataFrame#vector_sum` which zeroes out the value.

PR #403, opened 2017-08-29
Add Installation to README

koishimasato:


zverok:
Ugh. I am late for the party again :(
I'd like to say that it is, unfortunately, not very helpful installation instructions.
Better one for `daru` should consist of, at least:
* mentioning of usage with Bundler (OK, it is obvious, but `gem install daru` also obvious);
* mentioning of optional requirements and their installation (this is pretty important and could be tricky), and activation; and when you need them.

v0dro:
@zverok I think your concerns can be a separate issue?

zverok:
Seems so. Creating.

Issue #198, opened 2016-07-28
Proposal: Function for filling in missing data


sunderwood:
I would like to be able to fill or pad the nil values in vectors or data frames. I work primarily with time series data and being able to roll values forward to fill the "missing" spaces is very useful. I have used the python Pandas library before and [Pandas fillna function](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.fillna.html) allows me to fill nil values with a scalar, roll values forward, or roll values backwards.

v0dro:
Yes. That would be useful.

rainchen:
how is this going?
Is there an elegant way of updating missing data like pandas fillna function do?

currently I have to do sth like this:

```
        # fill missing Fare
        df.where(df['Fare'].eq(nil)).each_row_with_index do |row, idx|
          row['Fare'] = 123
          df.set_row_at([idx], row)
        end
```
comparing to `full['Embarked'].fillna('S',inplace=True)`

I'm expecting there is an one-liner solution for daru.

v0dro:
@lokeshh?

lokeshh:
See http://www.rubydoc.info/gems/daru/Daru/DataFrame#replace_values-instance_method

v0dro:
@sunderwood if your issue is resolved can you close this ticket?

sunderwood:
The function that @lokeshh referenced does not fully satisfy my request. I would like to be able to "roll" values forwards or backwards. For example, using the replace_values function example I would expect a `fillna(:forward)` to result in:

```
df = Daru::DataFrame.new({
  a: [1,    2,          3,   nil,        Float::NAN, nil, 1,   7],
  b: [:a,  :b,          nil, Float::NAN, nil,        3,   5,   8],
  c: ['a',  Float::NAN, 3,   4,          3,          5,   nil, 7]
}, index: 11..18)
df.fillna(:forward)
# => #<Daru::DataFrame(8x3)>
#          a   b   c
#   11   1   a   a
#   12   2   b   a
#   13   3   b   3
#   14   3   b   4
#   15   3   b   3
#   16   3   3   5
#   17   1   5   5
#   18   7   8   7
```
and a `fillna(:backward)` to result in: 

```
df = Daru::DataFrame.new({
  a: [1,    2,          3,   nil,        Float::NAN, nil, 1,   7],
  b: [:a,  :b,          nil, Float::NAN, nil,        3,   5,   8],
[...]

PR #408, opened 2017-09-02
#198 - Add rolling_fillna method to Daru::Dataframe and Daru::Vector

baarkerlounger:
https://github.com/SciRuby/daru/issues/198

Initial attempt at implementing a rolling fillna function. Straight forward replace with scalar value is not implemented since we already have `replace_values` function to do that. This just allows you to fill all Nil or NAN values with the last valid preceding or next valid following value.

### DIFF
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -784,14 +784,9 @@ module Daru
     def rolling_fillna!(direction=:forward)
       missing_idxs = indexes(*Daru::MISSING_VALUES)
       valid_idxs = Array(0...dv.size) - missing_idxs
-      if (direction == :forward)
-        valid_idxs = valid_idxs.reverse
-        comparator = :<
-      else
-        comparator = :>
-      end
+      valid_idxs, comp = (direction == :forward) ? [valid_idxs.reverse, :<] : [valid_idxs, :>]
       missing_idxs.each do |idx|
-        repl_idx = valid_idxs.find { |i| i.public_send(comparator, idx) }
+        repl_idx = valid_idxs.find { |i| i.public_send(comp, idx) }
         repl = repl_idx ? self[repl_idx] : 0
         self[idx] = repl
       end
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -794,7 +794,7 @@ module Daru
 
     # Non-destructive version of rolling_fillna!
     def rolling_fillna(direction=:forward)
-      dup.rolling_fillna!
+      dup.rolling_fillna!(direction)
     end
 
     # Lags the series by `k` periods.
--- a/lib/daru/vector.rb
+++ b/lib/daru/vector.rb
@@ -760,6 +760,44 @@ module Daru
       self
     end
 
+    # Rolling fillna
+    # replace all Float::NAN and NIL values with the preceeding or following value
+    #
+    # @param [Symbol] (:forward, :backward) whether replacement value is preceeding or following
+    #
+    # @example
+    #
+    # dv = Daru::Vector.new([1, 2, 1, 4, nil, Float::NAN, 3, nil, Float::NAN])
+    #
+    # 2.3.3 :068 > dv.rolling_fillna(:forward)
+    # => #<Daru::Vector(9)>
+    # 0   1
+    # 1   2
+    # 2   1
+    # 3   4
+    # 4   4
+    # 5   4
+    # 6   3
+    # 7   3
+    # 8   3
+    #
+    def rolling_fillna!(direction=:forward)
+      missing_idxs = indexes(*Daru::MISSING_VALUES)
+      missing_idxs.each do |idx|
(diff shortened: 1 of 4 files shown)


lib/daru/vector.rb
```diff
@@ -760,6 +760,43 @@ def replace_nils! replacement
...
+    #
+    def rolling_fillna!(direction=:forward)
+      missing_idxs = indexes(*Daru::MISSING_VALUES)
+      valid_idxs = Array(0...dv.size) - missing_idxs
+      valid_idxs, comp = direction == :forward ? [valid_idxs.reverse, :<] : [valid_idxs, :>]
+      missing_idxs.each do |idx|
+        repl_idx = valid_idxs.find { |i| i.public_send(comp, idx) }
+        repl = repl_idx ? self[repl_idx] : 0
+        self[idx] = repl
+      end
```
zverok: Don't you think it could be done simpler? I am too lazy to write working code, but idea is (pseudocode):
```ruby
enumerator = direction == :forward ? each : reverse_each
prev_valid = nil
enumerator.each do |idx, val| 
  if valid?(val)
    prev_valid = val
  else
    self[idx] = prev_valid
  end
end
```

spec/vector_spec.rb
```diff
@@ -1769,6 +1769,24 @@
...
+
+    context 'rolling_fillna! forwards' do
+      before { subject.rolling_fillna!(:forward) }
+      its(:to_a) { is_expected.to eq [1, 2, 1, 4, 4, 4, 3, 3, 3] }
+    end
+
+    context 'rolling_fillna! backwards' do
+      before { subject.rolling_fillna!(direction: :backward) }
+      its(:to_a) { is_expected.to eq [1, 2, 1, 4, 3, 3, 3, 0, 0] }
+    end
```
zverok: Don't forget about edge cases (vector starting with non-value, vector entirely consisting of non-values and so on)

lib/daru/vector.rb
baarkerlounger: Hmm I might be missing the idea but I think my method basically follows this pattern?

zverok: I've meant this (works against your specs):
```ruby
    def rolling_fillna!(direction=:forward)
      enumerator = direction == :forward ? index : index.reverse_each
      last_valid_value = 0
      enumerator.each do |idx|
        if valid_value?(self[idx])
          last_valid_value = self[idx]
        else
          self[idx] = last_valid_value
        end
      end
    end

    def valid_value?(v)
      v.respond_to?(:nan?) && v.nan? || v.nil? ? false : true
    end
```
What's the difference? Let's look at initial version:
1. List of missing indexes is created: enumeration through the vector (performance) + array of indexes (memory)
2. List of valid indexes is created: array of them (memory, and probably performance substracting say 10k items arrays)
3. Missing indexes is enumerated, and for each of them we search through valid indexes array (performance)

Proposed version:
1. Only one enumeration of entire vector
2. No additional huge arrays
;)

And, last but not least, try this test:
```ruby
    context 'with non-default index' do
      subject do
        Daru::Vector.new(
          [Float::NAN, 2, 1, 4, nil, Float::NAN, 3, nil, Float::NAN],
          index: %w[a b c d e f g h i]
        )
      end

      before { subject.rolling_fillna!(direction: :backward) }

      it { is_expected.to eq Daru::Vector.new([2, 2, 1, 4, 3, 3, 3, 0, 0], index: %w[a b c d e f g h i]) }
    end
```
...and prepare to be surprised.

baarkerlounger: Thanks for the detailed explanation - much clearer now, have changed.

lib/daru/vector.rb
```diff
@@ -1437,6 +1475,11 @@ def nan_positions
         end
     end
 
+    # Helper method for rolling_fillna
```
zverok: I believe the comment, if comment it, the comment could be more helpful, like "Helper method returning validity of arbitrary value" or something like that. Other methods probably could find it useful too ;)

zverok:
Thanks! :tada:

baarkerlounger:
@zverok thanks for merging. Think you accidentally merged rather than squash and merge though. Just a heads up.

zverok:
Ooops. Strange, I believe squash-and-merge were our default button. I'll squash them on master.

zverok:
Fixed, thanks for noticing!

lokeshh:
How about we expose`#rolling_fillna` through more generic `replace_values` like this `replace_values([nil], :forward_rolling)`?

lokeshh:
We will have a simplified API as decided here https://github.com/SciRuby/daru/issues/137#issuecomment-236146562

Further one doesn't first need to convert missing value representation other than nil to nil first. One can directly apply replace_values on say `_` for example without converting it to nil first using replace_values (['_'], :forward_rolling)

zverok:
ok.

Issue #402, opened 2017-08-28
Nyaplot uses index as default axis for vector but not for dataframe


baarkerlounger:
```
df = Daru::DataFrame.new({y1: [11, 15, 8], y2: [9, 10, 3]}, 
     index: ["a", "b", "c"])
```
```
df.plot type: :bar, y: :y1 do |plot, diagram|
  plot.y_label "Frequency"
  plot.x_label "Category"
  plot.yrange [0,20]
end
```
Throws `NoMethodError: undefined method `to_sym' for nil:NilClass`

I would expect it to behave like `Daru::Vector` and take the index as default axis.

```
dv = df[:y1]
```

```
dv.plot type: :bar, y: :y1 do |plot, diagram|
  plot.y_label "Frequency"
  plot.x_label "Category"
  plot.yrange [0,20]
end
```

![image](https://user-images.githubusercontent.com/5101747/29797112-40565c86-8c4d-11e7-9b9d-6bd8f21b9d57.png)

matugm:
Full stacktrace for reference:

```
/home/jesus/.gem/ruby/2.4.1/gems/nyaplot-0.1.6/lib/nyaplot/data.rb:105:in `column': undefined method `to_sym' for nil:NilClass (NoMethodError)
	from /home/jesus/.gem/ruby/2.4.1/gems/nyaplot-0.1.6/lib/nyaplot/diagram.rb:69:in `process_data'
	from /home/jesus/.gem/ruby/2.4.1/gems/nyaplot-0.1.6/lib/nyaplot/diagram.rb:16:in `initialize'
	from /home/jesus/.gem/ruby/2.4.1/gems/nyaplot-0.1.6/lib/nyaplot/plot.rb:75:in `new'
	from /home/jesus/.gem/ruby/2.4.1/gems/nyaplot-0.1.6/lib/nyaplot/plot.rb:75:in `add_with_df'
	from /home/jesus/.gem/ruby/2.4.1/gems/daru-0.1.6/lib/daru/plotting/nyaplot/dataframe.rb:174:in `block in add_multiple_diagrams'
	from /home/jesus/.gem/ruby/2.4.1/gems/daru-0.1.6/lib/daru/plotting/nyaplot/dataframe.rb:173:in `map'
	from /home/jesus/.gem/ruby/2.4.1/gems/daru-0.1.6/lib/daru/plotting/nyaplot/dataframe.rb:173:in `add_multiple_diagrams'
[...]

PR #414, opened 2017-09-19
Set index as default x axis for DataFrame

matugm:
Hi,
this is my attempt at fixing https://github.com/SciRuby/daru/issues/402.

It breaks one test, but otherwise it seems to work.

If you have a better idea on how to do this let me know :)

### DIFF
--- a/lib/daru/plotting/nyaplot/dataframe.rb
+++ b/lib/daru/plotting/nyaplot/dataframe.rb
@@ -22,8 +22,8 @@ module Daru
         #   df.plot type: :bar, x: :a, y: :b
         def plot opts={}, &block
           unless opts[:x]
-            opts[:x] = :_index
-            self.[]=(:_index, @index.to_a)
+            opts[:x]      = :_index
+            self[:_index] = @index.to_a
           end
 
           if opts[:categorized]
--- a/lib/daru/plotting/nyaplot/dataframe.rb
+++ b/lib/daru/plotting/nyaplot/dataframe.rb
@@ -21,7 +21,7 @@ module Daru
         #   df = Daru::DataFrame.new({a:['A', 'B', 'C', 'D', 'E'], b:[10,20,30,40,50]})
         #   df.plot type: :bar, x: :a, y: :b
         def plot opts={}, &block
-          index_as_default_x_axis(opts) if x_axis_undefined?(opts)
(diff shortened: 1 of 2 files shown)


lib/daru/plotting/nyaplot/dataframe.rb
```diff
@@ -21,6 +21,11 @@ module NyaplotLibrary
         #   df = Daru::DataFrame.new({a:['A', 'B', 'C', 'D', 'E'], b:[10,20,30,40,50]})
         #   df.plot type: :bar, x: :a, y: :b
         def plot opts={}, &block
+          unless opts[:x]
+            opts[:x] = :_index
+            self.[]=(:_index, @index.to_a)
```
zverok: Why not `self[:_index] = @index.to_a`?

v0dro:
@matugm ping!

matugm:
Will work on this today.

matugm:
Ok I made the requested changes. Only thing left is 2 Rubocop warnings.

lib/daru/plotting/nyaplot/dataframe.rb
```diff
@@ -30,6 +32,15 @@ def plot opts={}, &block
 
         private
 
+        def x_axis_undefined?(opts)
+          !(opts[:x] || opts.keys.any? { |k| k.to_s.match(/x\d+/) })
```
athityakumar: Just a minor nit-picking from my part. 😄 

This format ...
```ruby
x_axis_defined = cond_1 || cond_2
index_as_default_x_axis(opts) unless x_axis_defined
```
... seems a bit more readable, when compared to the current format ...
```ruby
x_axis_undefined = !(cond_1 || cond_2)
index_as_default_x_axis(opts) if x_axis_undefined
```

matugm: Ok, done.

lib/daru/plotting/nyaplot/dataframe.rb
matugm: No specific reason. I made the change :)

zverok:
@matugm But where's any specs for the new behavior?..

matugm:
Spec added & also tested manually with IRuby.

zverok:
Thanks!

Issue #399, opened 2017-08-28
Integrate yard-junk into CI.


v0dro:
The yard-junk gem identifies docs that are using outdated syntax. Use of this gem will tell us where it is outdated.

Reference:
https://github.com/SciRuby/daru/issues/318

athityakumar:
Thanks for working on this, @rohitner. I think this issue can be closed now.

PR #417, opened 2017-09-29
integrated yard-junk into Cl

rohitner:


rohitner:
issue #399

.travis.yml
```diff
@@ -15,6 +15,7 @@ matrix:
 script:
   - bundle exec rspec
   - bundle exec rubocop
+  - yard-junk
```
athityakumar: I think it'd be better to change `yard-junk` -> `bundle exec yard-junk`

daru.gemspec
```diff
@@ -54,6 +54,8 @@ EOF
 
   # it is required by NMatrix, yet we want to specify clearly which minimal version is OK
   spec.add_runtime_dependency 'packable', '~> 1.3.9'
+  # for integrating yard-junk with Cl
+  spec.add_runtime_dependency 'yard-junk', '~>0.0.7'
```
athityakumar: `yard-junk` is a `development_dependency` rather than a `runtime_dependency`, along with `rubocop` and `rspec`

rohitner:
@zverok please suggest changes to ignore yard script for ruby 2.0

zverok:
> please suggest changes to ignore yard script for ruby 2.0

Let's try to approach the problem from another side? To **include** yard-junk script only on `ruby-2.4` (because it will produce the same errors on any version of Ruby)
And that's also true for Rubocop, BTW

zverok:
Ah, and you should look at how it fails on Travis. It fails not on script running stage, but on gem installation. It could be easily avoided with `if RUBY_VESION >= '2.1'` in gemspec

rohitner:
The problem is that config in the travis build wasn't changed as expected for ruby 2.0, if that is solved, the build can be modified as required.

zverok:
> The problem is that config in the travis build wasn't changed as expected for ruby 2.0, if that is solved, the build can be modified as required.

It was. But before using the config, travis does `bundle install`. Which tries to install `yard-junk`, and fails (because gem specifies minimal Ruby version)

rohitner:
I have changed the build matrix as required.

lib/daru/maths/statistics/dataframe.rb
```diff
@@ -43,33 +43,33 @@ def max opts={}
         #   Standardize each Vector
         # @!method acf
         #   Calculate Autocorrelation coefficient
-        #   @param [Integer] max_lags (nil) Number of initial lags
+        #   @yieldparam [Integer] max_lags (nil) Number of initial lags
```
zverok: That's a wrong way. Right is:
```ruby
# @!method acf(max_lags)
#   Calculate Autocorrelation coefficient
#   @param [Integer] max_lags (nil) Number of initial lags
```

lib/daru/maths/statistics/vector.rb
```diff
@@ -549,28 +549,28 @@ def rolling function, n=10
 
         # @!method rolling_mean
         #   Calculate rolling average
-        #   @param [Integer] n (10) Loopback length
+        #   @yieldparam [Integer] n (10) Loopback length
```
zverok: Same as above.

lib/daru/index/index.rb
```diff
@@ -258,7 +258,7 @@ def self._load data
 
     # Provide an Index for sub vector produced
     #
-    # @param input_indexes [Array] the input by user to index the vector
+    # @option args [Array] the input by user to index the vector
```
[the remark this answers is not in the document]
rohitner: since symbol * isn't allowed as a parameter, I thought of * as set of options, will it do if i replace args with * ?

.travis.yml
```diff
@@ -9,13 +9,42 @@ rvm:
...
+      - bundle exec rspec
+      - bundle exec rubocop
+      - bundle exec yard-junk
+  - rvm: '2.4.0'
+    script:
+      - bundle add yard-junk
+      - bundle install
+      - bundle exec rspec
+      - bundle exec rubocop
+      - bundle exec yard-junk
```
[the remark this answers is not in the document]
rohitner: This is the only problem left, the exclude in matrix doesn't work as script: tag is not supported for build matrix. It is treated as another job and leads to double number of jobs.

rohitner:
@zverok is it okay to allow 2.0 build to fail for now to keep the number of builds intact?

.travis.yml
```diff
@@ -9,12 +9,17 @@ rvm:
   - '2.4.0'
 
 matrix:
+  allow_failures:
+    - rvm: '2.0'
```
athityakumar: I wouldn't prefer to allow failures on ruby-2.0 just because of yard-junk. Especially, there are quite a few features missing in ruby-2.0 (like `array.sort_by(n)`) which shouldn't overlooked by passing Travis CI builds. 

P.S - We are in the process of dropping support for ruby-2.0 as per issue #336.But until we support drop for ruby-2.0, let's keep it the correct form?

zverok:
Screw the Ruby 2.0 :)

Issue #324, opened 2017-03-20
#group_by with boolean gives `ArgumentError: comparison of Array with Array failed`


dansbits:
I was recently trying to get a count of true and false values in a column of my data frame and received `ArgumentError: comparison of Array with Array failed` when I tried to run group_by using only that column. I thought this might be something with my data set but the below example reproduces the error.

```
df = Daru::DataFrame.new({ id: (1..10).to_a, is_training: (1..10).to_a.map { rand(2) == 0 } })
df.group_by([:is_training]).count
```

zverok:
I believe it is due to the fact that currently `group_by` for some reason sorts groups, and booleans can't be compared with `<=>`. 
This redesign ticket is somehow related: https://github.com/SciRuby/daru/issues/152

PR #420, opened 2017-10-07
Fix tuple sorter

baarkerlounger:
```
keys = [[" Black and White", nil, 31],[" Black and White", "Rob Marshall", 205]]`
keys.sort(&TUPLE_SORTER)
ArgumentError: comparison of Array with Array failed
```


baarkerlounger:
@zverok made some changes, let me know what you think - will try to stop posting WIP PRs too!

zverok:
>  will try to stop posting WIP PRs too!

It is not necessary, in fact :) Just include `[WIP]` in the title of those PRs, and write a comment like "Now it is finished!" when it is :)

baarkerlounger:
@zverok think this is ready for another look

lib/daru/core/group_by.rb
```diff
@@ -13,7 +13,15 @@ def each_group
...
+          a_comp_len = a_comp.length
+          b_comp_len = b_comp.length
+          if a_comp_len == b_comp_len
+            a_comp <=> b_comp || 0
+          else
+            0
+          end
         else
           a ? 1 : -1
         end
```
zverok: WDYT about this version?
```ruby
TUPLE_SORTER = lambda do |left, right|
  return -1 if !right
  return 1 if !left

  left = left.compact
  right = right.compact
  return left <=> right if left.lenght == right.length
  left.length <=> right.length
end
```
It looks a bit more readable.

spec/dataframe_spec.rb
```diff
@@ -1846,6 +1846,13 @@
     end
   end
 
+  context 'Tuple Sorter' do
+    it 'sorts data frames with nils' do
+      df = Daru::DataFrame.from_csv 'spec/fixtures/duplicates.csv'
+      expect { df.group_by(df.vectors.map(&:to_s)) }.to_not raise_error(ArgumentError)
+    end
+  end
```
zverok: It should be where all other `group_by` specs are, as a separate `context`, named like `when there are nil in dataframe` or something like this.

And I don't think that separate fixture is necessary for testing this case, on-the-fly dataframe creation would be more visible.

lib/daru/core/group_by.rb
baarkerlounger: Done with necessary corrections

baarkerlounger:
@zverok made both suggested changes

spec/core/group_by_spec.rb
```diff
@@ -465,4 +465,21 @@
...
+          'string1' => ["Color", "Color", "Color", "Color", nil, "Color", "Color", " Black and White"],
+          'string2' => ["Test", "test2", nil, "test3", nil, "test", "test3", "test5"],
+          'num' => [1, nil, 3, 4, 5, 6, 7, nil]
+        }
+      )
+    }
+
+    it 'groups by without errors' do
+      expect { df.group_by(df.vectors.map(&:to_s)) }.to_not raise_error(ArgumentError)
+    end
```
zverok: Let's add a bit more specs :)
1. For results (not just "it doesn't fail")
2. For different cases for sorting, it can be buggy (for ex., I wrote that from the top of the head, and not sure if it should be `1` and `-1` in the first two lines or vice versa).

baarkerlounger: @zverok I didn't bother with testing the actual sorting too much because I think that's beyond the scope of this change - there are plenty of other specs that test the actual sorting part for all the other scenarios that break when you change the current behaviour).

baarkerlounger:
https://github.com/SciRuby/daru/issues/324

Issue #419, opened 2017-10-07
Delete duplicates method


baarkerlounger:
As an enhancement it would be useful to have a delete duplicates method like http://pandas.pydata.org/pandas-docs/version/0.17.1/generated/pandas.DataFrame.drop_duplicates.html that lets you optionally specify a column to consider.

PR #421, opened 2017-10-07
Unique

baarkerlounger:
https://github.com/SciRuby/daru/issues/419

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -706,7 +706,7 @@ module Daru
     end
 
     def unique(vecs=[])
-      vectors = vecs.empty? self.vectors.map(&:to_s) : vectors
+      vectors = vecs.empty? ? self.vectors.map(&:to_s) : vecs
       grouped = self.group_by(vectors)
       indexes = grouped.groups.values.map { |v| v[0] }
       self.row[*indexes]
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -706,7 +706,7 @@ module Daru
 
     # Return unique rows by vector specified or all vectors
     #
-    # @param vector_names [String][Symbol] that should be considered
+    # @param other [String][Symbol] vector names(s) that should be considered
     #
     # @example
     #
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -706,7 +706,7 @@ module Daru
 
     # Return unique rows by vector specified or all vectors
     #
-    # @param [String][Symbol][Array] Vector name or array of vector names that should be considered
+    # @param [String][Symbol] Vector name(s) that should be considered
     #
     # @example
     #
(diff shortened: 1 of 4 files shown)


lib/daru/dataframe.rb
```diff
@@ -705,6 +705,14 @@ def rolling_fillna(direction=:forward)
       dup.rolling_fillna!(direction)
     end
 
+    # Return unique rows by vector specified or all vectors
+    def unique(vecs=[])
+      vectors = vecs.empty? ? self.vectors.map(&:to_s) : vecs
+      grouped = self.group_by(vectors)
+      indexes = grouped.groups.values.map { |v| v[0] }
+      self.row[*indexes]
```
zverok: Unnecessary `self.`s

lib/daru/dataframe.rb
```diff
@@ -705,6 +705,14 @@ def rolling_fillna(direction=:forward)
       dup.rolling_fillna!(direction)
     end
 
+    # Return unique rows by vector specified or all vectors
```
zverok: Docs should be more detailed, with some examples and description how uniqueness is considered. And param and return value descriptions.

spec/dataframe_spec.rb
```diff
@@ -1820,6 +1820,27 @@
...
+      before { subject.unique }
+      its(:'a.to_a') { is_expected.to eq [1, 2, 3, 4, 5] }
+      its(:'b.to_a') { is_expected.to eq ['a', 'b', 'c', 'd', 'c'] }
+    end
+
+    context 'given a vector' do
+      before { subject.unique(:b) }
+      its(:'a.to_a') { is_expected.to eq [1, 2, 3, 4] }
+      its(:'b.to_a') { is_expected.to eq ['a', 'b', 'c', 'd'] }
+    end
```
zverok: This spec contradicts method definition, which implies array of vector names should be passed.

lib/daru/dataframe.rb
```diff
@@ -705,6 +705,14 @@ def rolling_fillna(direction=:forward)
...
+    def unique(vecs=[])
```
zverok: a) method should be named `uniq` (see Ruby's Array API: https://docs.ruby-lang.org/en/2.4.0/Array.html#method-i-uniq)
b) It probably should also support blocks, like `Array#uniq`

baarkerlounger:
Requires https://github.com/SciRuby/daru/pull/420 to be merged first as the tuple sorting fix is needed but shouldn't form part of this PR.

spec/dataframe_spec.rb
```diff
@@ -1820,6 +1820,33 @@
...
+    context 'given a vector' do
+      it do
+        result = df.uniq("color")
+        expect(result.shape.first).to eq 2
+      end
+    end
+
+    context 'given an array of vectors' do
+      it do
+        result = df.uniq(["color", "director_name"])
```
zverok: Why not `uniq(:color, :director_name)` (e.g. splatting arguments)? It looks more reasonable for me.

baarkerlounger: Mostly just because the `group_by` takes an array but both work for me. Have changed.

baarkerlounger:
@zverok anything left on this?

PR #430, opened 2017-10-10
[WIP] Removes IO modules from Daru

athityakumar:
Through this PR, I'd like to safely port away all IO modules of Daru, keeping the Marshalling methods intact. I'll request a review in a couple of days.


spec/maths/statistics/vector_spec.rb
```diff
@@ -710,7 +710,7 @@
   end
 
   RSpec.shared_examples 'correct macd' do |*args|
-    let(:source) { Daru::DataFrame.from_csv('spec/fixtures/macd_data.csv') }
+    let(:source) { Daru::DataFrame.read_csv('spec/fixtures/macd_data.csv') }
```
athityakumar: I somehow want to keep out the IO methods from the tests of daru. For example, a simple parsing error with daru-io could be misunderstood as a logical error with the method (of course, unless the error log is seen). Only 2 more fixture files are remaining - `macd_data.csv` and `duplicates.csv`. Can a truncated version of both these files be fed as manual data in the specs and then maybe remove both the files?

Gemfile
```diff
@@ -1,2 +1,3 @@
 source 'https://rubygems.org'
 gemspec
+gem 'daru-io', :git => 'https://github.com/athityakumar/daru-io.git'
```
athityakumar: Will `daru-io` be a `runtime_dependency` or `development_dependency` of daru?

athityakumar:
@zverok - Ready for a review 😄

spec/maths/statistics/vector_spec.rb
zverok: Well, for me an ability to use CSV fixtures in general Daru tests seems important, especially for future (more complicated stats methods and stuff). So let's leave this as is.

And I believe that daru-io errors would be clearly seen in specs, if the problem is daru-io

daru.gemspec
```diff
@@ -33,6 +33,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.1.0'
 
   spec.add_runtime_dependency 'backports'
+  spec.add_runtime_dependency 'daru-io'
```
athityakumar: @zverok - daru-io and daru are both runtime_dependencies of each other, and that's creating an infinite loop in travis builds while bundle installing. :trollface:

zverok: Yeah, already thought of that. I believe that daru should be _development_ dependency of `daru-io` (meaning since "daru 1.0" there is no sense in installing `daru-io` by itself, as it is installed with `daru`)

athityakumar: Hmm, should v0.1.1 be released for daru-io with this dependency change?

zverok: Probably so. At the same time when `daru-1.0.0-pre` would be released, probably?..

athityakumar: Acknowledged. Let me just add `daru-io` to the Gemfile rather than then gemspec, so that it doesn't interfere with travis CI builds. Before `daru-1.0.0-pre` release, this can be changed after `daru-io-0.1.1` releases. Can you add this into the milestone?

athityakumar:
@zverok - Opened #434 for noting down the dependency issue. Can you add it into the milestone to keep track?

zverok:
Done.

v0dro:
This needs a serious relook and separating IO from the core gem should be seriously considered before the next major release.

Issue #438, opened 2017-11-03
Error when using Group by with single row Data Frame


IdeoConcepts:
Hi,

With Daru v0.2.0, it seems that if the DataFrame has only one row, the group_by method will fail with "ArgumentError: Can't create DataFrame from xx" error.

Example:
````
> df = Daru::DataFrame.new({a: [1], b: [2]})
> df.group_by([:a])
ArgumentError: Can't create DataFrame from [2]
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/dataframe.rb:176:in `rows'
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/core/group_by.rb:351:in `resultant_context'
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/core/group_by.rb:302:in `init_groups_df'
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/core/group_by.rb:37:in `initialize'
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/dataframe.rb:1461:in `new'
        from /usr/local/rvm/gems/ruby-2.0.0-p598/gems/daru-0.2.0/lib/daru/dataframe.rb:1461:in `group_by'`
   
````    
Whereas a multi rows DataFrame will be fine.
````
> df = Daru::DataFrame.new({a: [1,3], b: [2,4]})
> df.group_by([:a])
 => #<Daru::DataFrame(2x1)>
           b
   1   0   2
   3   1   4
````
Many thanks

baarkerlounger:
A group_by on a single row dataframe should just return that same dataframe right?

baarkerlounger:
Can this be closed given the above has been merged?

PR #440, opened 2017-11-03
#438 - Fix group_by for DataFrame with single row

baarkerlounger:
https://github.com/SciRuby/daru/issues/438

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1449,7 +1449,7 @@ module Daru
     #   # ["foo", "three", 8]=>[7],
     #   # ["foo", "two", 3]=>[2, 4]}
     def group_by *vectors
-      return self unless nrows > 1
+      return self if nrows == 1
       vectors.flatten!
       missing_vectors = vectors - @vectors.to_a
       unless missing_vectors.empty?
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -1449,12 +1449,17 @@ module Daru
     #   # ["foo", "three", 8]=>[7],
     #   # ["foo", "two", 3]=>[2, 4]}
     def group_by *vectors
+      return self unless nrows > 1
       vectors.flatten!
-      # FIXME: wouldn't it better to do vectors - @vectors here and
-      # raise one error with all non-existent vector names?.. - zverok, 2016-05-18
-      vectors.each { |v|
-        raise(ArgumentError, "Vector #{v} does not exist") unless has_vector?(v)
-      }
+      missing_vectors = vectors - @vectors.to_a
+      unless missing_vectors.empty?
+        errorString = if missing_vectors.size > 1
+          "Vectors #{missing_vectors.join(", ")} do not exist"
+        else
+          "Vector #{missing_vectors.first} does not exist"
+        end
+        raise(ArgumentError, errorString)
+      end
 
       vectors = [@vectors.first] if vectors.empty?
 
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -172,9 +172,11 @@ module Daru
           DataFrame.new(source.transpose, opts)
         elsif ArrayHelper.array_of?(source, Vector)
           from_vector_rows(source, opts)
-        else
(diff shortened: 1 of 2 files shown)


spec/dataframe_spec.rb
```diff
@@ -3266,6 +3266,15 @@
     end
   end
 
+  context "group_by" do
+    context "on a single row DataFrame" do
+      let(:df){ Daru::DataFrame.new({a: [1], b: [2]}) }
+      it "returns itself" do
```
zverok: I am not sure it is the most reasonable behavior. Here is what happens if we have large DF:
```ruby
df = Daru::DataFrame.new(city: %w[Kyiv Kharkiv Odesa Kyiv Kharkiv Odesa], year: [2015, 2015, 2015, 2016, 2016, 2016], value: [1, 2, 3, 4, 5, 6])
# => #<Daru::DataFrame(6x3)>
#            city    year   value
#       0    Kyiv    2015       1
#       1 Kharkiv    2015       2
#       2   Odesa    2015       3
#       3    Kyiv    2016       4
#       4 Kharkiv    2016       5
#       5   Odesa    2016       6 
df.group_by(:city)
# => #<Daru::DataFrame(6x2)>
#                    year   value
# Kharkiv       1    2015       2
#               4    2016       5
#    Kyiv       0    2015       1
#               3    2016       4
#   Odesa       2    2015       3
#               5    2016       6 
```
Here is what (I believe) should happen with 1-row DF:
```ruby
df = Daru::DataFrame.new(city: %w[Kyiv], year: [2015], value: [1])
# => #<Daru::DataFrame(1x3)>
#        city  year value
#     0  Kyiv  2015     1 
df.group_by(:city)
# => #<Daru::DataFrame(1x2)>
#                    year   value
#    Kyiv       0    2015       1
```
(E.g. the same index/vectors change as with other groupings).

lib/daru/dataframe.rb
```diff
@@ -1449,12 +1449,13 @@ def vector_mean max_missing=0
...
+      return self if nrows == 1
       vectors.flatten!
-      # FIXME: wouldn't it better to do vectors - @vectors here and
-      # raise one error with all non-existent vector names?.. - zverok, 2016-05-18
-      vectors.each { |v|
-        raise(ArgumentError, "Vector #{v} does not exist") unless has_vector?(v)
-      }
+      missing = vectors - @vectors.to_a
+      unless missing.empty?
+        e = missing.size > 1 ? "Vectors #{missing.join(', ')} do not exist" : "Vector #{missing.first} does not exist"
```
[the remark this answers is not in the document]
baarkerlounger: Updated

spec/dataframe_spec.rb
baarkerlounger: Fair point, made it match the behaviour of other cases.

lib/daru/dataframe.rb
```diff
@@ -173,7 +173,8 @@ def rows source, opts={}
         elsif ArrayHelper.array_of?(source, Vector)
           from_vector_rows(source, opts)
         else
-          raise ArgumentError, "Can't create DataFrame from #{source}"
+          row = source.map { |v| Array(v) }
```
zverok: What would be consequences? What if `source` doesn't respond to `map`? 

And what if it _does_ respond, where are the tests for the change? Now not only `group_by` behavior affected.

baarkerlounger: Have added a check for that case.

In terms of specs - I've added one for the breaking case and all others are still passing, do you have a particular spec case in mind?

zverok: Not that. Rather, I believe that local bug should have local fix, e.g. either in `GroupBy#resultant_context` (where DF is tried to be created from "wrong" data), or in `DataFrame#access_row_tuples_by_indexs` (which probably should always return array-of-arrays, shouldn't it?)

baarkerlounger: Ah I get it, agreed that's a better way. Have updated.

baarkerlounger:
@zverok happy to merge?

baarkerlounger:
@zverok updated

zverok:
Thanks!

PR #441, opened 2017-12-13
fix Documentation link

takkanm:
I think that http://www.rubydoc.info/gems/daru/ is better than https://rubygems.org/gems/daru


zverok:
Thanks!

Issue #390, opened 2017-08-19
Vector#set doesn't work for Category vectors


zverok:
```ruby
v = Daru::Vector.new([1,2,3])
vc = v.to_category
vc.send(:set, 5, 6)
# NoMethodError: undefined method `<<' for nil:NilClass
```
I believe the reason is category vector, while imitating being the same vector, cleans up `@data` variable. Probably it is not the only broken method for category vectors.

Showcased by https://github.com/SciRuby/daru/pull/388

@lokeshh WDYT?

lokeshh:
@zverok Yes, `:set` method is missing for category vector. We have to fix this.

rohitner:
![image](https://user-images.githubusercontent.com/21151546/33787715-97576236-dc94-11e7-9b56-915d806dd4d5.png)
![image](https://user-images.githubusercontent.com/21151546/33787768-d1546e66-dc94-11e7-92a2-6dc6fdcd23ac.png)

I added these lines of code but it didn't work. Should I send it as a PR?

rohitner:
ping! @zverok @lokeshh @athityakumar

lokeshh:
@rohitner When you do `vc.send(:set, 5, 6)` you setting a new category `6` in index `5`. Because `6` is a new category you need to first call `#add_category` and then modify `@cat_hash` and `@array` accordingly.

lokeshh:
@rohitner It would be better if you could create a PR and we discuss it there.

PR #446, opened 2018-01-09
add method set to category.rb

rohitner:
created a PR for discussion on issue #390

### DIFF
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -922,6 +922,13 @@ module Daru
       @cat_hash[category] << pos
     end
 
+    def set indexes, val
+      @index = @index.add(indexes)
+      @cat_hash.store(val,[indexes])
+      @array << indexes
+      modify_category_at indexes, val
+    end
+
     def order_with new
       if new.to_set != categories.to_set
         raise ArgumentError, 'The contents of new and old order must be the same.'
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -924,6 +924,7 @@ module Daru
 
     def set index, val
       add_category(val)
+      return self.[]= index, val if @index.include? index
       @index = @index.add(index)
       @array << index
       @cat_hash.store(val,[index])
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -927,11 +927,17 @@ module Daru
         raise ArgumentError, "Invalid category #{category}, "\
           'to add a new category use #add_category'
       end
-      return self.[]= pos, category if @index.include? pos
-      @array << categories.index(category)
-      @index = @index.add(pos)
-      @cat_hash.store(category,[pos])
-      self.[]= pos, category
+      pos = pos.is_a?(Array) ? pos : [pos]
+      pos.each do |x|
+        if @index.include? x
+          self.[]= x, category
+          next
+        end
+        @array << categories.index(category)
+        @index = @index.add(x)
+        @cat_hash.store(category,[x])
+        self.[]= x, category
+      end
     end
 
     def order_with new
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -918,15 +918,16 @@ module Daru
       end
       old_category = category_from_position pos
       @array[pos] = int_from_cat category
-      @cat_hash[old_category].delete pos
+      @cat_hash[old_category].delete pos unless old_category.nil?
       @cat_hash[category] << pos
     end
 
-    def set indexes, val
-      @index = @index.add(indexes)
-      @cat_hash.store(val,[indexes])
-      @array << indexes
-      modify_category_at indexes, val
+    def set index, val
+      add_category(val)
+      @index = @index.add(index)
+      @array << index
+      @cat_hash.store(val,[index])
+      self.[]= index, val
     end
 
     def order_with new
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -922,20 +922,21 @@ module Daru
       @cat_hash[category] << pos
     end
 
-    def set indexes, category
+    def set *indexes, category
       unless categories.include? category
         raise ArgumentError, "Invalid category #{category}, "\
           'to add a new category use #add_category'
       end
-      indexes = indexes.is_a?(Array) ? indexes : [indexes]
-      indexes.each do |pos|
-        if pos.is_a? Range
-          pos.each do |index|
-            set_util index, category
+
+      indexes.flatten.each do |index|
+        case index
+        when Range
+          index.each do |pos|
+            set_util pos, category
           end
-          next
+        else
+          set_util index, category
         end
-        set_util pos, category
       end
     end
 
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -922,21 +922,31 @@ module Daru
       @cat_hash[category] << pos
     end
 
-    def set pos, category
+    def set indexes, category
       unless categories.include? category
         raise ArgumentError, "Invalid category #{category}, "\
           'to add a new category use #add_category'
       end
-      pos = pos.is_a?(Array) ? pos : [pos]
-      pos.each do |x|
-        if @index.include? x
-          self.[]= x, category
+      indexes = indexes.is_a?(Array) ? indexes : [indexes]
+      indexes.each do |pos|
+        if pos.is_a? Range
+          pos.each do |index|
+            set_util index, category
+          end
           next
         end
-        @array << int_from_cat(category)
-        @index = @index.add(x)
-        @cat_hash[category] << x
+        set_util pos, category
       end
+    end
+
+    def set_util pos, category
+      if @index.include? pos
+        self.[]= pos, category
+        return
+      end
+      @array << int_from_cat(category)
+      @index = @index.add(pos)
+      @cat_hash[category] << pos
     end
 
     def order_with new
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -935,7 +935,6 @@ module Daru
         end
         @array << categories.index(category)
         @index = @index.add(x)
-        @cat_hash.store(category,[x])
         self.[]= x, category
       end
     end
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -935,7 +935,7 @@ module Daru
         end
         @array << categories.index(category)
         @index = @index.add(x)
-        self.[]= x, category
+        @cat_hash[category] << x
       end
     end
 
--- a/lib/daru/category.rb
+++ b/lib/daru/category.rb
@@ -918,7 +918,7 @@ module Daru
       end
       old_category = category_from_position pos
       @array[pos] = int_from_cat category
-      @cat_hash[old_category].delete pos unless old_category.nil?
+      @cat_hash[old_category].delete pos
       @cat_hash[category] << pos
     end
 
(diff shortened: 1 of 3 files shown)


v0dro:
@lokeshh ping!

lib/daru/category.rb
```diff
@@ -922,6 +922,13 @@ def modify_category_at pos, category
       @cat_hash[category] << pos
     end
 
+    def set indexes, val
+      @index = @index.add(indexes)
+      @cat_hash.store(val,[indexes])
+      @array << indexes
+      modify_category_at indexes, val
+    end
+
```
lokeshh: `@array` is supposed to store categories (in integer form) not indexes. That's what you are are getting error here.

rohitner: I doubt that because ` puts @array` gives `0,1,2,5`.

rohitner:
@lokeshh does it LGTY ?

lib/daru/category.rb
lokeshh: 0, 1, 2 refers to categories not indexes in @array

rohitner:
Adding a method `array` as follows
```
def array
      @array
end
```
gives the following output
```
irb(main):002:0> v = Daru::Vector.new([1,2,3])
=> #<Daru::Vector(3)>
   0   1
   1   2
   2   3
irb(main):003:0> vc = v.to_category
=> #<Daru::Vector(3):category>
   0   1
   1   2
   2   3
irb(main):004:0> vc.categories
=> [1, 2, 3]
irb(main):005:0> vc.array
=> [0, 1, 2]
irb(main):006:0> vc.index
=> #<Daru::Index(3): {0, 1, 2}>
```

rohitner:
Can you please check :sweat_smile: 
```
irb(main):026:0> v = Daru::Vector.new([1,2,3])
=> #<Daru::Vector(3)>
   0   1
   1   2
   2   3
irb(main):027:0> vc = v.to_category
=> #<Daru::Vector(3):category>
   0   1
   1   2
   2   3
irb(main):028:0> vc.send(:set, 5, 6)
=> #<Daru::Vector(4):category>
   0   1
   1   2
   2   3
   5   6
irb(main):029:0> vc.to_ints
=> [0, 1, 2, 3]
irb(main):030:0> vc[1]=6
=> 6
irb(main):031:0> vc.to_ints
=> [0, 3, 2, 3]
```

lib/daru/category.rb
```diff
@@ -918,10 +918,20 @@ def modify_category_at pos, category
...
       @cat_hash[category] << pos
     end
 
+    def set index, val
+      categories = @cat_hash.keys
+      add_category(val)
+      return self.[]= index, val if @index.include? index
+      @index = @index.add(index)
+      @array << @array.max + 1 unless categories.size == @cat_hash.keys.size
+      @cat_hash.store(val,[index])
```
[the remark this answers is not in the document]
lokeshh: Here's a simple way this can be implemented:

1. Raise exception is category is new else continue.
2. Make changes to `@cat_hash` to add new entry.
3. Make changes to `@array` for new index.
That's it!

rohitner:
ping! @lokeshh

rohitner:
@lokeshh I doubt whether `set` works for multiple indices. Please check.

rohitner:
ping:exclamation: @lokeshh

lokeshh:
You need to pass an array like `v.set [5, 6], 2`. And sorry for replying late.

rohitner:
Can you please check, if I have made the required changes?

lib/daru/category.rb
```diff
@@ -918,10 +918,28 @@ def modify_category_at pos, category
...
+      pos = pos.is_a?(Array) ? pos : [pos]
+      pos.each do |x|
+        if @index.include? x
+          self.[]= x, category
+          next
+        end
+        @array << categories.index(category)
+        @index = @index.add(x)
+        @cat_hash.store(category,[x])
+        self.[]= x, category
```
lokeshh: I think this line `@cat_hash.store(category,[x])` will delete the previous value of `@cat_hash[category]` and hence forgetting the previous indexes where the category occurs.

```ruby
[1] pry(main)> a = {cat: [1, 2, 3]}
=> {:cat=>[1, 2, 3]}
[2] pry(main)> a.store(:cat, [4])
=> [4]
[3] pry(main)> a
=> {:cat=>[4]}

```

lokeshh: Why are you calling `self.[]= x, category` at line 939?

spec/category_spec.rb
```diff
@@ -828,6 +828,11 @@
         expect(@dv[:yoda]).to eq(666)
       end
 
+      it "assigns at a new index" do
+        @dv.send(:set,:xtbr,3)
+        expect(@dv[:xtbr]).to eq(3)
+      end
+
       it "assigns correctly for a mixed index Vector" do
```
lokeshh: You should add a spec that tests `set` correctly adds a new category. Here you are testing the existing category only.

lib/daru/category.rb
rohitner: I have removed these lines in next commit because the new index added will be hashed in continuation of the previous index by the`[ ]=` method. So storing an index say 100 in cat_hash after the natural index 2 will be problematic because `[ ]=` will store the index 100 as 3. Removing this line doesn't affect the working.

rohitner: This line is added to make sure all other variables are updated which would be the case if you set the index usually.

spec/category_spec.rb
rohitner: According to your previous suggestion of raising an error in case of an invalid category, `set` is not supposed to add only new indices.

lib/daru/category.rb
lokeshh: That seems like an over kill. There are just two variables which needs updating `@array` and `@cat_hash`. You have updated the `@array` and for `@cat_hash`, you can just do `@cat_hash[category] << pos`.

spec/category_spec.rb
lokeshh: You can do the following:
```ruby
@dv.add_category :abc
@dv.send(:set, :xtbr, :abc)
```

rohitner: done :+1: can you please check

v0dro:
@lokeshh is this good for merge?

spec/category_spec.rb
```diff
@@ -828,6 +828,12 @@
         expect(@dv[:yoda]).to eq(666)
       end
 
+      it "assigns a new category at a new index " do
+        @dv.add_category :abc
+        @dv.send(:set, :xtbr, :abc)
+        expect(@dv[:xtbr]).to eq(:abc)
```
lokeshh: It would be good if you also test multiple assignment like `dv.set [:a, :v], :abc`

lokeshh:
@v0dro No not yet but almost done.

rohitner:
@lokeshh are there any other changes/additions needed?

lib/daru/category.rb
```diff
@@ -918,10 +918,27 @@ def modify_category_at pos, category
       end
       old_category = category_from_position pos
       @array[pos] = int_from_cat category
-      @cat_hash[old_category].delete pos
+      @cat_hash[old_category].delete pos unless old_category.nil?
       @cat_hash[category] << pos
     end
```
lokeshh: I guess this is no longer needed.

lib/daru/category.rb
```diff
@@ -918,10 +918,27 @@ def modify_category_at pos, category
...
+          'to add a new category use #add_category'
+      end
+      pos = pos.is_a?(Array) ? pos : [pos]
+      pos.each do |x|
+        if @index.include? x
+          self.[]= x, category
+          next
+        end
+        @array << categories.index(category)
+        @index = @index.add(x)
```
lokeshh: I think it will be better to use cat_to_int function here

rohitner: I think you mean `int_to_cat`. Please review the changes.

lib/daru/category.rb
```diff
@@ -922,6 +922,23 @@ def modify_category_at pos, category
       @cat_hash[category] << pos
     end
 
+    def set pos, category
+      unless categories.include? category
+        raise ArgumentError, "Invalid category #{category}, "\
+          'to add a new category use #add_category'
+      end
+      pos = pos.is_a?(Array) ? pos : [pos]
+      pos.each do |x|
```
lokeshh: Please make sure that this would also work for iterators like dv.set 1..10, 7

Also it would be good if you name your variables depending upon what purpose they serve. For example x can be replaced by pos and pos by positions

lokeshh: I think you can look at vector.rb on how to do that.

rohitner: please review

spec/category_spec.rb
```diff
@@ -828,6 +828,22 @@
...
+      end
+
+      it "assigns a new category at a multiple indices " do
+        @dv.add_category :rbn
+          @dv.send(:set, [:padme, :bdn, :cy..:db], :rbn)
+        expect(@dv[:bdn]).to eq(:rbn)
```
lokeshh: The iterators aren't supposed to be supported inside an array. What I meant was that they can be used like this `dv.set 1..4, :abc`. This is equivalent to `dv.set [1, 2, 3, 4], :abc`

v0dro:
@rohitner such a huge discussion for a simple PR is not a good thing at all. You should focus more, understand all the issues at once and submit a PR that is in spirit of the rest of the gem. Most of @lokeshh 's comments so far are mostly about things that are already there in daru, but which you were unaware of. If you explore the codebase more you can find them on your own, and you should.

Please note that all of us are volunteers and we cannot continuously point out nitty gritties.

spec/category_spec.rb
rohitner: Okay, I have done changes which can handle arrays and iterators at the same time, can we add this functionality to the method in vector.rb as well? Also, you have been calling `set` as a public method, does that change need to be done too?

lokeshh: I think method in vector.rb already has this functionality. No there's no need to change the method to public.

rohitner: Please check this [notebook](http://nbviewer.jupyter.org/gist/rohitner/1d1223d29338715d59f0b91532595544). There is also a bug which can be seen in the last query. I hope there aren't any other changes needed before this could be merged.

lokeshh: You need not support array and range at the same time. So it's not a bug. The user can either use an array, single element, range or iterator but not a mixture of these at the same time.

rohitner: I have dropped support for a mixture of arguments, please check.

spec/category_spec.rb
```diff
@@ -828,6 +828,23 @@
...
+        @dv.add_category :abc
+        @dv.send(:set, :xtbr, :abc)
+        expect(@dv[:xtbr]).to eq(:abc)
+      end
+
+      it "assigns a new category at a multiple indices " do
+        @dv.add_category :rbn
+          @dv.send(:set, [:padme, :bdn], :rbn)
+          @dv.send(:set, :cy..:db, :rbn)
+        expect(@dv[:bdn]).to eq(:rbn)
```
lokeshh: It's good idea to test ranges and arrays in different tests. In general always test two different things in different tests.

rohitner: Sure, please check.

rohitner: ping! @lokeshh

PR #447, opened 2018-01-18
rolling_fillna! bugfixes on Daru::Vector and Daru::DataFrame

mhammiche:
According to documentation `rolling_fillna!` should return the modified object
Actually the Vector is correctly updated but it return the index! 

```ruby
dv = Daru::Vector.new([1, 2, 1, 4, nil, Float::NAN, 3, nil, Float::NAN])
dv.rolling_fillna!(:forward)
# => #<Daru::Index(9): {0, 1, 2, 3, 4, 5, 6, 7, 8}>
dv
=> #<Daru::Vector(9)>
   0   1
   1   2
   2   1
   3   4
   4   4
   5   4
   6   3
   7   3
   8   3
```

`rolling_fillna` is not working either as it clone the object before calling `rolling_fillna!`

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -698,6 +698,7 @@ module Daru
     #
     def rolling_fillna!(direction=:forward)
       @data.each { |vec| vec.rolling_fillna!(direction) }
+      self
     end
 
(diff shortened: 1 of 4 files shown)


zverok:
Thanks!

PR #449, opened 2018-02-05
Add link to blogpost

Shkrt:
I recently wrote an introductory blogpost about daru. The article is entry level and basically an example of manipulating small, simple dataset.


v0dro:
Thanks :)

PR #461, opened 2018-06-14
add module clean and bi

rohitner:
- [ ] Fix statsample dependency issue
- [x] Add frequently used methods from other sources 
- [ ] Make the structure modular
- [ ] Add tests


lib/daru/clean/clean.rb
```diff
@@ -0,0 +1,76 @@
+module Daru
+  # Module encapsulating advanced data cleaning methods on DataFrame.
+  module Clean
+    # replace elements with those from given set of elements
+    # based on similarity
```
zverok: 1. What "similarity" is in this context?
2. I don't believe that method name properly reflects what it does (the name suggests something is _matched_ -- probably returning "match result", while in example/docs values are _replaced_).

lib/daru/clean/clean.rb
```diff
@@ -0,0 +1,76 @@
...
+    #   #   0    MALE
+    #   #   1  FEMALE
+    #   #   2    MALE
+    #   #   3  FEMALE
+    #   #   4  FEMALE
+    def fuzzy_match vector, dict
+      # TODO
+    end
+
+    # fill missing places using linear regression
```
zverok: 1. Reason for this exact name?
2. I believe that there could be other means of filling missing places (like, IDK, mean of neighbour data; constant value; rolling fill that is already implemented; and so on). Don't you think this task in fact requires something like `fill(&algorythm)` + library of filling algorithms?

lib/daru/clean/clean.rb
```diff
@@ -0,0 +1,76 @@
...
+    #   #   1   3
+    #   #   2   5
+    #   #   3   7
+    #   #   4   9
+    #   #   5   11
+    def impute vector
+      # TODO
+    end
+
+    # Split multi valued cells in more than one column into rows
```
zverok: The example looks useful, yet underthought:
* how do we specify which columns to split?
* how do we specify what is split substring? (I believe `,` is not the only possibility)
* split_**cell** (singular) looks like not the best name possible.

lib/daru/clean/clean.rb
```diff
@@ -0,0 +1,76 @@
...
+    #   #   0  2002  t-shirt1     M     23
+    #   #   1  2003  t-shirt1     L     23
+    #   #   2  3001  t-shirt2     S     24
+    #   #   3  3002  t-shirt2     M     24
+    #   #   4  3003  t-shirt2     L     24
+    def split_cell
+      # TODO
+    end
+
+    # check dependency of one column to another
```
zverok: What's the definition of "dependency"?.. I believe it is something about "correlation", no?.. And how it is related to data cleaning?

spec/clean/clean.rb
```diff
@@ -0,0 +1,62 @@
...
+  before(:each) do
+    @df = Daru::DataFrame.new(
+      {
+        a: [1,2,3,4,5],
+        b: ['mal','femal','M','F','Fem'],
+        c: [10,nil,30,40,nil],
+        d: ['a,c','a','b','b','b,a'],
+        e: [:A,:N,:B,:C,:N]
+      })
+  end
```
zverok: Please use "new" style of specs (see daru-io for good examples) -- with `subject`, `let` and so on.

lib/daru/clean/clean.rb
rohitner: 1. By similarity, it means the Levenshtein distance between the strings as mentioned [here](https://github.com/seamusabshere/fuzzy_match#top-3-reasons-you-should-use-fuzzymatch).
2. I will rename it to fuzzy_replace to make it more intuitive. We can perhaps add a non-destructive version.

lib/daru/clean/clean.rb
rohitner: 1. The name was inspired by its use in the R package [imputeR](https://cran.r-project.org/web/packages/imputeR/imputeR.pdf)
2. Yes, we can add an interpolation module to statsample which will take care of the proposed generalized approach of filling algorithms. WDYT?

lib/daru/clean/clean.rb
rohitner: I feel letting the user specify the details would make them use the method less likely (which is the case in the corresponding method in OpenRefine). Although modularity does no harm, I was thinking of applying split to all the noticeable columns with such error. I will update the doc with an option for specifying the delimiter and a better method name.

lib/daru/clean/clean.rb
rohitner: The method serves as a way of calculating correlation but with binary output for use on-the-go. It will be helpful for cleaning accidentally duplicated columns.

v0dro:
Closing since there's no issue which demands modules like `BI` and `Clean`, lack of tests and oldness of PR.

Issue #462, opened 2018-07-03
Alternate method for `access_row_tuples_by_indexs`


Prakriti-nith:
Recently, `access_row_tuples_by_indexs` method has been removed from daru. Please let us know the alternative.
Refer: [#454](https://github.com/SciRuby/daru/pull/454#pullrequestreview-133706760), [Sciruby/daru-view#103 (issue)](https://github.com/SciRuby/daru-view/issues/103)

Shekharrajak:
@Prakriti-nith , please create a PR in daru repo and putting the same method in same place (revert the changes for this method) with testcases. 

Thanks!

PR #463, opened 2018-07-04
Added access_row_tuples_by_indexs method

Prakriti-nith:
#462 
This PR aims to revert back the changes for `access_row_tuples_by_indexs` method.

### DIFF
--- a/spec/dataframe_spec.rb
+++ b/spec/dataframe_spec.rb
@@ -4040,6 +4040,9 @@ describe Daru::DataFrame do
   context '#access_row_tuples_by_indexs' do
     let(:df) {
       Daru::DataFrame.new({col: [:a, :b, :c, :d, :e], num: [52,12,07,17,01]}) }
+    let(:df_idx) {
+      Daru::DataFrame.new({a: [52, 12, 07], b: [1, 2, 3]}, index: [:one, :two, :three])
+    }
     context 'when no index is given' do
       it 'returns empty Array' do
         expect(df.access_row_tuples_by_indexs()).to eq([])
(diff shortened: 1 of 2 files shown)


spec/dataframe_spec.rb
```diff
@@ -4037,6 +4037,22 @@ def create_test(*args, &_proc)
...
+    context 'when no index is given' do
+      it 'returns empty Array' do
+        expect(df.access_row_tuples_by_indexs()).to eq([])
+      end
+    end
+    context 'when index(s) are given' do
+      it 'returns Array of row tuples' do
+        expect(df.access_row_tuples_by_indexs(1)).to eq([[:b, 12]])
+        expect(df.access_row_tuples_by_indexs(0,3)).to eq([[:a, 52], [:d, 17]])
+      end
```
Shekharrajak: Please also add testcases for dataframe having custom indexes like `index: [:one, :two, :three, :four, :five])`

Prakriti-nith:
I tried it with multi indexes and found two different bugs which needs to be managed separately. This method will work with multi indexes only when a single index is provided
1. I think there is a bug in [#pos](https://github.com/SciRuby/daru/blob/master/lib/daru/index/multi_index.rb#L160) method of Daru::MultiIndex when multiple indexes are provided. `access_row_tuples_by_index` will work in this case when the issue with `pos` method will be resolved. 
daru1
daru2

2. #pos method is working fine in the below case but #access_row_tuples_by_indexs gives error as it [uses `index`](https://github.com/SciRuby/daru/pull/463/files#diff-cfd587da29af6a7a2d1f98c614ef123fR2286) for creating new rows but the provided `index` is not actually the valid index. It is used to get the positions.
daru3

Shekharrajak:
Ping @zverok , @v0dro Please look into the PR as soon as possible. 

Thanks!

Shekharrajak:
Ping @zverok , @v0dro Please look into the PR as soon as possible. 

Thanks!

paisible-wanderer:
Note: in dataframe.rb, it is possible to use instead:
```
def access_row_tuples_by_indexs *indexes
  get_sub_dataframe(indexes, by_position: false).map_rows(&:to_a)
end
```
It is about 10 times slower that the current reverted code but, among the two bugs found by @Prakriti-nith , it is not affected by the second one (ie I have: ```
df.access_row_tuples_by_indexs [:a, :one]
=> [[11, 1], [12, 2]]
```)

Shekharrajak:
Thanks @paisible-wanderer for informing us. @Prakriti-nith , please use similar thing for dataframe having multi index and test it for different cases.

PR #470, opened 2018-11-29
Fix reindex vector on argument error

Yuki-Inoue:
Currently, when one gives non-`Daru::Index` object to `Daru::DataFrame#reindex_vectors` method, exception of `NoMethodError` is raised instead of `ArgumentError`, the intended exception.

Reason is that when `raise`ing the exception, an non-existing variable of `new_index` is referenced.  I think this was caused by simply copying the code from `reindex` method.

This PR fixes this and when non index is given, it properly throws ArgumentError.


Yuki-Inoue:
ping @v0dro

PR #476, opened 2018-12-26
Adds option to display all data on inspect

paisible-wanderer:
Hello,

By default, using inspect on dataframe/vector/other will not display the whole data.

I have added an option that I think can be useful.

```
vect = Daru::Vector.new((1..16))

# before
vect
# => #<Daru::Vector(16)>
#    0   1
#    1   2
#    2   3
#    3   4
#    4   5
#    5   6
#    6   7
#    7   8
#    8   9
#    9  10
#   10  11
#   11  12
#   12  13
#   13  14
#   14  15
#  ... ...


# after
puts vect.inspect(all: true);nil
# #<Daru::Vector(16)>
#    0   1
#    1   2
#    2   3
#    3   4
#    4   5
#    5   6
#    6   7
#    7   8
#    8   9
#    9  10
#   10  11
#   11  12
#   12  13
#   13  14
#   14  15
#   15  16
# => nil
```

Let me know if there is an better way, or any improvement I can make to this PR.

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -2229,7 +2229,11 @@ module Daru
     end
 
     # Pretty print in a nice table format for the command line (irb/pry/iruby)
-    def inspect spacing=10, threshold=15
+    #
+    # @option [Boolean] all: when true, will remove thresholding (and will override any value set for threshold)
+    def inspect spacing=10, threshold=15, all: false
+      threshold = :all if all
+
       name_part = @name ? ": #{@name} " : ''
 
       "#<#{self.class}#{name_part}(#{nrows}x#{ncols})>\n" +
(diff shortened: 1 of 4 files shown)


lib/daru/dataframe.rb
```diff
@@ -2229,7 +2229,11 @@ def transpose
     end
 
     # Pretty print in a nice table format for the command line (irb/pry/iruby)
-    def inspect spacing=10, threshold=15
+    #
+    # @option [Boolean] all: when true, will remove thresholding (and will override any value set for threshold)
+    def inspect spacing=10, threshold=15, all: false
```
Shekharrajak: I think we should have a method that can set the threshold for Daru::DataFrame and Vector to display maximum number of rows and columns. 

So I believe we should have  a method like 

```
Daru::DataFrame.inspect_max_rows = nil # to load all the rows

```
Default value will be set to 15.

What do you think ? @paisible-wanderer

Shekharrajak: In that case, you can mention the number of rows in the threshold option, itself.

PR #477, opened 2018-12-28
Index#dup should copy reference to name too

Yuki-Inoue:
Currently, if `idx.instance_of?(Daru::Index)`, then following statement does not hold:

```ruby
idx.dup.name == idx.name
# => returns false!!
```

This is because name is not copied on `Daru::Index#dup`, which I think is not the expected behavior at least to the library user.

So, this PR fixes it.

### DIFF
--- a/lib/daru/index/index.rb
+++ b/lib/daru/index/index.rb
@@ -239,7 +239,7 @@ module Daru
     end
 
     def dup
-      Daru::Index.new @keys
+      Daru::Index.new @keys, name: @name
     end
 
     def add *indexes
--- a/spec/index/index_spec.rb
+++ b/spec/index/index_spec.rb
@@ -401,4 +401,17 @@ describe Daru::Index do
          )
     }
   end
+
+  context "#dup" do
+    let(:idx) do
+      Daru::Index.new(['speaker', 'mic', 'guitar', 'amp'],
+                      name: 'instruments')
+    end
+    subject { idx.dup }
+
+    it { is_expected.to eq idx }
+    it 'have same names' do
+      expect(subject.name).to eq idx.name
+    end
+  end
 end


Shekharrajak:
I think `==` should check name of the index as well : 

```
[28] pry(main)> dii = di.dup
=> #<Daru::Index(4): {speaker, mic, guitar, amp}>
[29] pry(main)> dii == di
=> true

```
This should return false.

Shekharrajak:
This PR looks good to fix [the issue mentioned above](https://github.com/SciRuby/daru/pull/477#issue-241291828). But it came up another question :  [Mentioned in above comment](https://github.com/SciRuby/daru/pull/477#issuecomment-456343248).

Shekharrajak:
I will wait for 24 hours for any other comments. If fine then I will go ahead and merge.

Yuki-Inoue:
@Shekharrajak Hi, thanks for the feedback.

The problem fixed by this PR and the concern brought by your comment is, IMHO, a separate concern, so I'd be glad if you go ahead and merge this.

The concern you brought up, although I, myself haven't encountered such a case where I need to compare the Daru objects based on the content of its data, I can agree that such method will be useful.

About the naming of such method, I disagree with naming the method as `#equal`.  The reason is that in Ruby, following methods are used for equality comparison, and implementing a new method named `#equal` is somewhat confusing.

```
equals?(other)   <-> python `is` correspondent
eql?(other)         -> used for objects with `hash` method, for hash equality check
==(other)            <-> python `==` correspondent
===(other)         -> used for `case` matching
```

If such a method is needed, I'd recommend using a name which does not conflict with the ruby's equality methods.  For example, how about `data_eq?(other)` or `content_eq?(other)`?

Shekharrajak:
@Yuki-Inoue , yes we need to discuss about my concern in different issue ticket ([Ruby doc about ==, eq ,equal](http://ruby-doc.org/core-2.6/Object.html#method-i-eql-3F) ) and it is different from this PR. 

`.dup` should duplicate the all class attributes. So it should duplicate the `name` as well. It looks good Vector , DataFrame and MultiIndex already.

This PR is in.

PR #489, opened 2019-02-18
fix table style

kojix2:
# Improve the appearance of the table. 

I am [running IRuby & Daru on Google Colab](https://dev.to/kojix2/ruby-kernel-in-google-colaboratory-32ni). Google Colab officially supports Python. But Ruby works.

### The `dataframe` style is not adapted to the table.
```html
<table></table>
```
image

### After adding the dataframe class to the table tag, Daru::DataFrame looks better.
This is as same as Pandas. 
```html
<table border="1" class="dataframe"></table>
```
image

I hope the Google Colab team officially support the IRuby kernel someday. Therefore, I think that such minor change in appearance is more important than expected.

### DIFF
--- a/lib/daru/iruby/templates/dataframe.html.erb
+++ b/lib/daru/iruby/templates/dataframe.html.erb
@@ -1,5 +1,5 @@
 <b> Daru::DataFrame<%= name ? ": #{name} " : ''%>(<%=nrows%>x<%=ncols%>) </b>
-<table>
+<table border="1" class="dataframe">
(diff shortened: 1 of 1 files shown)


kojix2:
I tried it.
The notebook example seems to be OK.

https://nbviewer.jupyter.org/gist/kojix2/986aef56aedcbfb59be3283a886457cf

image

Shekharrajak:
Thanks @kojix2 , It will be merged after the travis error is sovled.

PR #492, opened 2019-02-28
fix typos

kojix2:
Arguement -> Argument

image


kojix2:
I searched `daru` directory with `grep -rl`
```bash
grep -rl "arguement"
```
I found two other typos. (lower case)
I think it is all with this.

Issue #502, opened 2019-03-27
How to set the maximum number of rows to display


kojix2:
Daru's data frame displays up to 30 rows.
I think the default threshold of 30 is reasonable in most cases.

However, when processing messy data, I often want to see 100 lines or more with my own eyes.

The threshold is defined as default values ​​for the arguments here. 

https://github.com/SciRuby/daru/blob/d17887e279c27e2b2ec7a252627c04b4b03c0a7a/lib/daru/dataframe.rb#L2094

So, you can change the threshold with the following patch.

```ruby
class Daru::DataFrame
    alias_method :__to_html__, :to_html
    def to_html
        __to_html__()
    end
end
```

This is not a cool way.
I want a better way to configure Daru.

Shekharrajak:
Yes! It makes sense. There was a similar discussion happened https://github.com/SciRuby/daru/pull/476 , please have a look.

kojix2:
My first idea is like
```ruby
Daru.configure do |c|
  c.max_rows = 100
  c.max_columns = nil
end
```

But I think this is good too.
```ruby
Daru::DataFrame.inspect_max_rows = 10
Daru::DataFrame.inspect_max_colms = nil
Daru::Vector.inspect_max_rows = nil 
```

Shekharrajak:
Thanks for letting me this. Yes! We can have configuration as well, where all these default things can be set.

kojix2:
I realized that setting options alone was not enough for me.
I want to display a particular data frame / vector only once. For such use, a method like `show_all` may be more appropriate than config.

`df.show_all`

Shekharrajak:
Thanks for suggestion @kojix2 .

I think we should have global option to set the maximum number of rows to display as well as `all_rows` and `all_colms` property in dataframe level.

WDYT?

PR #506, opened 2019-04-18
Add support for options to Daru module

kojix2:
This is an experimental implementation that adds options and settings to Daru.
#476 #502

Yes. I have to write test code, but I do not know how to do it.
Please review.


Shekharrajak:
Thanks! Can you please paste some examples as well, in the description box?

Issue #512, opened 2019-04-28
Map method. Should the Daru::Vector return an Array?


kojix2:
Daru's vector returns an array using the map method.

```ruby
v = Daru::Vector[1,2,3,4]
v.map {|i| i}
# => [1, 2, 3, 4]
```

If you use module Enumerable, the map method returns an array.
However, in Ruby there is no rule that the map method must return an array.
For example, 

```ruby
require 'numo/narray'
a = Numo::SFloat.new(10).seq
a.map{|i| i}
#=> Numo::SFloat#shape=[10]
#[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

Daru has prepared `recode` as a method to return a Vector. 

```ruby
v = Daru::Vector[1,2,3,4]
v.recode {|i| i}
#=> #<Daru::Vector(4)>
#   0   1
#   1   2
#   2   3
#   3   4
```

This practice is not so common.
I feel that Daru's iterator has some confusion. 
If there is room to change the API from now, please consider.

him666:
looking into this

kojix2:
Thank you for your work. @him666 

But what I suggested was to change the behavior of the map method.

Currently, the map method returns an Array, but I suggested changing it to return Daru::Vector.

This is not a technical issue but a user consensus issue.

`map_to_vector` is similar to `recode`.  
So I think we don't need a new method.

ncs1:
I too, think that map on Daru::Vector should return Daru::Vector and not Array.

PR #513, opened 2019-05-07
Feature/map to vector

him666:
added map_to_vector wrapper methods in order to solve the next issue
https://github.com/SciRuby/daru/issues/512

tried to avoid breaking changes since the Gem relies a lot on the normal map method that returns an array.

### DIFF
--- a/spec/vector_spec.rb
+++ b/spec/vector_spec.rb
@@ -1339,6 +1339,7 @@ describe Daru::Vector do
           a = @common_all_dtypes.map_to_vector { |v| v }
           expect(a).to eq(Daru::Vector.new([5, 5, 5, 5, 5, 6, 6, 7, 8, 9, 10, 1, 2, 3, 4, 11, -99, -99]))
         end
+
         it "maps to vector with change" do
           a = @common_all_dtypes.map_to_vector { |v| v + 1}
           expect(a).to eq(Daru::Vector.new([6, 6, 6, 6, 6, 7, 7, 8, 9, 10, 11, 2, 3, 4, 5, 12, -98, -98]))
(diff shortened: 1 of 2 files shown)


Shekharrajak:
Thanks @him666 for the PR. Can you please write few examples from old method and new method, so that later we can track the changes ?

Thanks again!

spec/vector_spec.rb
```diff
@@ -1334,6 +1334,30 @@
...
+        it "maps to vector with change" do
+          a = @common_all_dtypes.map_to_vector { |v| v + 1}
+          expect(a).to eq(Daru::Vector.new([6, 6, 6, 6, 6, 7, 7, 8, 9, 10, 11, 2, 3, 4, 5, 12, -98, -98]))
+        end 
+      end
+
+      context "#map_to_vector!" do
+        it "maps to vector destructive" do
+          @common_all_dtypes.map_to_vector! { |v| v }
+          expect(@common_all_dtypes).to eq(Daru::Vector.new([5, 5, 5, 5, 5, 6, 6, 7, 8, 9, 10, 1, 2, 3, 4, 11, -99, -99]))
```
Shekharrajak: The line length should not go > 79 characters.

spec/vector_spec.rb
```diff
@@ -1334,6 +1334,30 @@
...
+        end
+
+        it "maps to vector destructive" do
+          @common_all_dtypes.map_to_vector! { |v| v + 1}
+          expect(@common_all_dtypes).to eq(Daru::Vector.new([6, 6, 6, 6, 6, 7, 7, 8, 9, 10, 11, 2, 3, 4, 5, 12, -98, -98]))
```
Shekharrajak: The line length should not go > 79 characters.

kojix2:
Thank you for your PR.

But to be honest, I don't want to add the `map_to_vector` method. 
Because `recode` method already exists.

The problem is that the `recode` method is hard to find.
The naming of `map_to_vector` may be better than `recode`.

However, my opinion is that the behavior of `map` should be that of `recode` or `map_to_vector`.
And the `to_a` method is sufficient to convert a `Daru::Vector` to an `Array`.

But I'm not sure if this idea is really good.
I will read Daru's code this month..

Shekharrajak:
@kojix2 , yes! I was thinking the same. Also it is written in [recode docs](https://github.com/SciRuby/daru/blob/master/lib/daru/vector.rb#L674). @him666 , please have a look whether `map` is overridden or not

v0dro:
Closing since we need a changed `map` method.

Issue #515, opened 2019-07-05
Why Daru has these inconsistent situations?


weqopy:
I'd like to use this data to show the situation made me confused:
```
[25] pry(main)> dates=["2018-03-30", "2018-04-02", "2018-04-27", "2018-05-31", "2018-06-29", "2018-07-31", "2018-08-31", "2018-09-28", "2018-10-31", "2018-11-30"]
=> ["2018-03-30",
 "2018-04-02",
 "2018-04-27",
 "2018-05-31",
 "2018-06-29",
 "2018-07-31",
 "2018-08-31",
 "2018-09-28",
 "2018-10-31",
 "2018-11-30"]
[26] pry(main)> val=[1.00000001, 0.9999, 0.9908, 1.0885, 1.0586, 1.0374, 0.9456, 0.9638, 0.8397, 0.8788]
=> [1.00000001, 0.9999, 0.9908, 1.0885, 1.0586, 1.0374, 0.9456, 0.9638, 0.8397, 0.8788]
[27] pry(main)> id=Daru::DateTimeIndex.new(dates)
=> #<Daru::DateTimeIndex(10) 2018-03-30T00:00:00+00:00...2018-11-30T00:00:00+00:00>
[28] pry(main)> df = Daru::DataFrame.new({val: val}, index: id)
=> #<Daru::DataFrame(10x1)>
                   val
 2018-03-30 1.00000001
 2018-04-02     0.9999
 2018-04-27     0.9908
 2018-05-31     1.0885
 2018-06-29     1.0586
 2018-07-31     1.0374
 2018-08-31     0.9456
 2018-09-28     0.9638
 2018-10-31     0.8397
 2018-11-30     0.8788
```
- first & last
```
[29] pry(main)> df.val.first
=> 1.00000001
[30] pry(main)> df.val.last 
NoMethodError: undefined method `last' for #<Daru::Vector:0x00007f43dbc591f0>
from /usr/local/lib/ruby/gems/2.4.0/gems/daru-0.2.1/lib/daru/vector.rb:1420:in `method_missing'
# which I supposed it returns 0.8788
```

- The return type
```
[31] pry(main)> df.val['2018-03-30','2018-04-30']
=> #<Daru::Vector(3)>
                                       val
 2018-03-30T00:00:00+           1.00000001
 2018-04-02T00:00:00+               0.9999
 2018-04-27T00:00:00+               0.9908
[32] pry(main)> df.val['2018-04']
[...]

PR #519, opened 2019-08-28
Add last for vector

kojix2:
#515


lib/daru/vector.rb
```diff
@@ -453,6 +453,11 @@ def tail q=10
       self[start..(size-1)]
     end
 
+    def last q=1
+      # The Enumerable mixin dose not provide the last method.
```
onelharrison: comment typo `does`

PR #521, opened 2020-01-02
Fixed keyword parameters warning with Ruby 2.7

ankane:
Fixes the follow warning with `Daru::DataFrame.from_csv`:

```text
.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/daru-0.2.2/lib/daru/io/io.rb:234: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
.rbenv/versions/2.7.0/lib/ruby/2.7.0/csv.rb:679: warning: The called method `parse' is defined here
```

### DIFF
--- a/lib/daru/io/io.rb
+++ b/lib/daru/io/io.rb
@@ -231,7 +231,7 @@ module Daru
       def from_csv_hash(path, opts)
         csv_as_arrays =
           ::CSV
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @v0dro!

Issue #507, opened 2019-04-19
Insert column/row into DataFrame at specified location.


kojix2:
Hello.
Daru::DataFrame doesn't seem to have `insert` function.

`set_at` or `set_row_at` will overwrite. 
`[]=` does not specify the insertion location.

Shekharrajak:
> `[]=` `add_vector` does not specify the insertion location.

Can you please elaborate a bit this line?

Shekharrajak:
This is good point. 

we can add another column easily but then we have to order it. You can also do `df[:player] = df.index` but yeah we should have a method to insert a column in specific place. may be we can have `#insert` method having information about index of column, vector name and vector value.

kojix2:
That's exactly what I wanted to say.

Shekharrajak:
Thanks for letting us know this usecase.

cyrillefr:
Hello @Shekharrajak , I would like to work on this one.

Shekharrajak:
@cyrillefr , please go ahead. You can discuss your approach here once so that we can verify that you are going in right way then you can open a PR.

cyrillefr:
Hello @Shekharrajak , sorry for the little delay.
My idea is : add some new data at the end, then reorder.
The method would look like this:
`def insert_vector n, name, source`
&emsp;`vector = Daru::Vector.new(source, index: @index, name: @name)`
&emsp;`@data << vector`
&emsp;`@vectors = @vectors.add name`
&emsp;`ordr = @vectors.dup.to_a`
&emsp;`elmnt = ordr.pop`
&emsp;`ordr = ordr.insert n, elmnt`
&emsp;`self.order=ordr`
`end`

Shekharrajak:
Thanks @cyrillefr , please go ahead open a PR with few corner test cases. Love to see the changes in new release :)

PR #525, opened 2020-03-10
Insert column data at specified location

cyrillefr:
- new method insert_vector in lib/daru/dataframe.rb (based on add_vector)
 - added specs

Aims to resolve issue #507

### DIFF
--- a/lib/daru/dataframe.rb
+++ b/lib/daru/dataframe.rb
@@ -545,6 +545,17 @@ module Daru
       self[n] = vector
     end
 
+    def insert_vector n, name, source
+      raise ArgumentError unless source.is_a? Array
+      vector = Daru::Vector.new(source, index: @index, name: @name)
+      @data << vector
+      @vectors = @vectors.add name
+      ordr = @vectors.dup.to_a
+      elmnt = ordr.pop
+      ordr = ordr.insert n, elmnt
+      self.order=ordr
+    end
+
     # Access a row or set/create a row. Refer #[] and #[]= docs for details.
     #
     # == Usage
--- a/spec/dataframe_spec.rb
+++ b/spec/dataframe_spec.rb
@@ -738,6 +738,47 @@ describe Daru::DataFrame do
     }
   end
 
+  context "#insert_vector" do
+    subject(:data_frame) {
+      Daru::DataFrame.new({b: [11,12,13,14,15], a: [1,2,3,4,5],
+        c: [11,22,33,44,55]}, order: [:a, :b, :c],
+        index: [:one, :two, :three, :four, :five])
+    }
+
+    it "insert a new vector at the desired slot" do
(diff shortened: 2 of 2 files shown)


spec/dataframe_spec.rb
```diff
@@ -738,6 +738,47 @@
...
+        d: [710, 720, 730, 740, 750],
+        b: [11, 12, 13, 14, 15],
+        c: [11,22,33,44,55]}, order: [:a, :d, :b, :c],
+        index: [:one, :two, :three, :four, :five]
+      )
+      data_frame.insert_vector 1, :d, [710, 720, 730, 740, 750]
+      expect(subject).to eq df
+    end
+
+    it "raises error for data array being too big" do
```
Shekharrajak: Isn't it Argument Error ?

spec/dataframe_spec.rb
```diff
@@ -738,6 +738,47 @@
...
+      expect {
+        source = (1..8).to_a
+        data_frame.insert_vector 1, :d, source
+      }.to raise_error(IndexError)
+    end
+
+    it "raises error for invalid index value" do
```
Shekharrajak: Isn't it IndexError ?

spec/dataframe_spec.rb
```diff
@@ -738,6 +738,47 @@
...
+      expect {
+        source = (1..5).to_a
+        data_frame.insert_vector 4, :d, source
+      }.to raise_error(ArgumentError)
+    end
+
+    it "raises error for invalid source type" do
```
Shekharrajak: It would be better to provide some message/hint along with error, so that user can fix it error .

lib/daru/dataframe.rb
```diff
@@ -545,6 +545,17 @@ def add_vector n, vector
       self[n] = vector
     end
 
+    def insert_vector n, name, source
```
Shekharrajak: - Why it is always necessary to provide name of the index ? 

- Please make the argument names clearly like: use `nthColumn` instead of `n`

lib/daru/dataframe.rb
```diff
@@ -545,6 +545,17 @@ def add_vector n, vector
...
+      raise ArgumentError unless source.is_a? Array
+      vector = Daru::Vector.new(source, index: @index, name: @name)
```
Shekharrajak: Can we have choice to pass vector directly to be inserted (think about other ways as well)?

lib/daru/dataframe.rb
```diff
@@ -545,6 +545,17 @@ def add_vector n, vector
...
+      @data << vector
+      @vectors = @vectors.add name
```
Shekharrajak: Can't we do `@vectors = @vectors.insert(n, name)` simply?

PR #528, opened 2020-03-31
Second, Minute, Hour and Day offsets now express equality

### LINKED ISSUE #526: Offsets should be comparable [feature request]
This is a feature request / bug report.

It would be awesome if the offsets where comparable.  However they are not...

```ruby
one_minute_in_seconds = Daru::Offsets::Second.new(60)
one_minute = Daru::Offsets::Minute.new()

raise 'not equal' unless one_minute_in_seconds == one_minute
```

Why would you want to do this?

Given I have a time series
Which I expect to have a specific frequency 
And I create the Daru::Vector from a hash (timestamp: value)
Then I want to be able to compare the derived frequency with the expected frequency
So that I can confirm I am working with validated data

jpaulgs:
Fixes #526 

I haven't implemented equality in all offsets because Week, Month and Year offsets do not implement the `multiplier` method.

### DIFF
--- a/lib/daru/date_time/offsets.rb
+++ b/lib/daru/date_time/offsets.rb
@@ -111,6 +111,14 @@ module Daru
       def - date_time
         date_time - @n*multiplier
       end
+
+      def ==(other_obj)
+        other_obj.is_a?(Tick) && period == other_obj.period
+      end
+
+      def period
+        @n * multiplier
+      end
     end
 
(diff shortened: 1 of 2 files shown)


jpaulgs:
Hmm it looks like 2.1, 2.2 and 2.3 are failing because pry-byebug isn't available...

image

PR #536, opened 2020-07-28
Fixed keyword parameters warning with Ruby 2.7

jonspalmer:
Cleanup a few more Ruby 2.7 issues similar to #521


dvalerojimenez:
I can confirm this as still being a bug in master branch, would be nice to have the patch as `Daru::DataFrame.write_csv` is broken.

It could also solve the issue with `Daru::DataFrame.from_csv`: https://github.com/SciRuby/daru/issues/543

